PR #2, opened 2015-09-05
Concretize the API

Lukasa:
This is a spike, a bit of investigative programming, to get a feel for how the API should fit together. This branch contains the full set of changes required to get an example Twisted server up and running.


Lukasa:
Ok, I think I've done enough work here to justify merging back into the mainline.

Issue #14, opened 2015-09-17
Priority implementation.


Lukasa:
Right now we don't handle priority at all. Of course, exactly what's done with priority information will be up to the full server implementation, but for now we at least need to be able to signal the priority information up the stack.ls

Goals:

- Have events for priority changes.
- Track priority information.
- Have an API for constructing and returning the correct priority tree.
- Correctly handle priority-carrying frames, both PRIORITY frames themselves and other frames that can carry priority information.

Lukasa:
Updating this quickly, we now have an [out-of-tree priority implementation](https://github.com/python-hyper/priority) that solves points two and three in the above list, and pull request #32 dealt with the severest problems of four.

All we need now are events that emit priority information when those are sent. I think the way to do this is to emit the events only when priority information is *explicitly* signaled: that causes a minimal change to the events API, and should ensure that code that knows a little too much about the way events are signaled (*cough* our test suite) doesn't start to break because new events start popping up.

PR #17, opened 2015-09-19
Add asyncio example

mhils:
This PR adds a (Python 3.5 only) asyncio example mainly because I wanted to play around with Python 3.5, hyper-h2 and asyncio :smile: 

hyper-h2 works impressively well, great work. This example runs out-of-the-box in Chrome. :relaxed: 
(Firefox requires #14 to be implemented)

(The certificate in this PR is a self-signed certificate for localhost)


examples/asyncio/asyncio-server.py
```diff
@@ -0,0 +1,76 @@
...
+        self.transport = None
+
+    def connection_made(self, transport: asyncio.Transport):
+        self.transport = transport
+        self.conn.initiate_connection()
+        self.transport.write(self.conn.data_to_send())
+
+    def data_received(self, data: bytes):
+        events = self.conn.receive_data(data)
+        if self.conn.data_to_send:
```
mhils: I copied this from the twisted example, but the check here doesn't seem to make much sense?

Lukasa: Yeah, you're right, that's probably not a sensible check. ;) It's left behind from before hyper-h2 had a proper interface for sending data.

I've substantially rewritten the Twisted example today because I'm at PyCon UK and I want to have something interesting to show off. Feel free to diverge from the Twisted example if you want!

Lukasa:
\o/ Thanks for doing this, it's fantastic!

Let me fix the tests up because they're making me sad, and then I'll go ahead and merge.

examples/asyncio/asyncio-server.py
mhils: Great stuff. I removed the check, have fun at PyCon! :snake: :smile:

Lukasa:
Mind rebasing once more? I had to pin a test dependency version.

mhils:
Done!

Lukasa:
Alright, I'm going to merge this!

I'm particularly impressed that you've written an asyncio server that doesn't actually contain the word 'coroutine' or 'yield from'! Very nice. =D

PR #24, opened 2015-10-05
Partial dependency update.

Lukasa:
Updates all our test dependencies except for py.test, which has a bug in 2.8.1.


Lukasa:
Apparently py.test 2.8.2 is out today, so I may hold off on merging this until it releases.

PR #25, opened 2015-10-05
Enforce settings values.

### LINKED ISSUE #5: Fully-functional settings logic
We have to keep track of two different kinds of settings: the settings that belong to the remote peer and the ones that belong to us. We must not violate the limits imposed by the remote peer, and we must enforce that the remote peer does not violate our limits.

To do:

- Add an API for making changes to local settings and emitting the SETTINGS frame necessary for those changes.
- Add an underlying data structure or interface that lets us tell the difference between the settings that the user requested and the ones that have been ACK'd by the remote peer (put another way, tell the difference between the settings we want and the ones the peer is currently meant to be obeying)
- Add a SettingsChangeAcknowledged event to signal to the user when a settings change is ACKd.
- Add enforcement for all settings.

Lukasa:
This pull request handles the remaining part of #5: settings enforcement. We need to enforce the following settings:

- [x] SETTINGS_HEADER_TABLE_SIZE
- [x] SETTINGS_ENABLE_PUSH
- [x] SETTINGS_MAX_CONCURRENT_STREAMS
- [x] SETTINGS_INITIAL_WINDOW_SIZE
- [x] SETTINGS_MAX_FRAME_SIZE
- [x] SETTINGS_MAX_HEADER_LIST_SIZE

This resolves #5.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -568,8 +568,14 @@ class H2Connection(object):
         """
         Handle a frame received on the connection.
         """
-        # I don't love using __class__ here, maybe reconsider it.
         try:
+            if frame.body_len > self.max_inbound_frame_size:
+                raise ProtocolError(
+                    "Received overlong frame: length %d, max %d" %
+                    (frame.body_len, self.max_inbound_frame_size)
+                )
+
+            # I don't love using __class__ here, maybe reconsider it.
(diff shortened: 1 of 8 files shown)


Lukasa:
I believe this accounts for settings policing.

PR #27, opened 2015-10-08
Police remote flow control windows.

### LINKED ISSUE #16: Complete remote flow control handling.
We should enforce the flow control windows remotely. We need to keep track of what we believe the window is, and abort connections if the remote party misunderstands it.

Lukasa:
Resolves #16.


Lukasa:
I don't think this is quite done: I need to fix up some names on some methods to add balance between local and remote flow control.

PR #35, opened 2015-10-17
Fix file docstring in eventlet example

alexwlchan:


Lukasa:
Thanks @alexwlchan! :sparkles: :cake: :sparkles:.

PR #47, opened 2015-10-21
Improve performance of checking if stream is open

Lukasa:
This turns out to be our major bottleneck during the h2load test, so making this faster is better. This relatively simple change improves the performance of the CPython version of this code by 3x.

The next step is to see whether we can reduce the number of these checks we have to do by expiring old streams.


Lukasa:
The performance improvement for this on PyPy is a little slimmer: call it 20%.

PR #52, opened 2015-10-27
Curio example

### LINKED ISSUE #51: Add a Curio example
Why not? @dabeaz has written a 3.5-only async library called [curio](https://github.com/dabeaz/curio). He claims it's "barely in a prerelease state", but why not give it a chance to show us what it can do by having an example for it. More examples can only be better!

Lukasa:
Resolves #51.

So, this turns out to work really well! I wrote my standard example (JSON header response) in just a couple of minutes with no drama, so I extended it to be a bit more detailed: this is now the equivalent of the twisted example server.

It acts essentially as a HTTP/2-only equivalent of SimpleHTTPServer. `python curio-server.py <path to directory>` serves that directory: then use your favourite modern browser to hit the page and marvel at the HTTP/2 wonder!

/cc @dabeaz


dabeaz:
Awesome!

PR #77, opened 2016-01-08
Reject frames with invalid padding.

### LINKED ISSUE #73: Reject DATA frames with invalid padding length.
Found when testing proposed Twisted HTTP/2 server.

```
  6.1. DATA
    ✓ Sends a DATA frame with 0x0 stream identifier
    ✓ Sends a DATA frame on the stream that is not opend
    × Sends a DATA frame with invalid pad length
      - The endpoint MUST treat this as a connection error of type PROTOCOL_ERROR.
        Expected: GOAWAY frame (ErrorCode: PROTOCOL_ERROR)
                  RST_STREAM frame (ErrorCode: PROTOCOL_ERROR)
                  Connection close
          Actual: DATA frame (Length: 4541, Flags: 0)
```

Lukasa:
This change force-rejects frames with invalid padding, treating them as a connection error of type PROTOCOL_ERROR.

Resolves #73.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -7,12 +7,14 @@ An implementation of a HTTP/2 connection.
 """
 from enum import Enum, IntEnum
 
+from hyperframe.exceptions import InvalidPaddingError
 from hyperframe.frame import (
     GoAwayFrame, WindowUpdateFrame, HeadersFrame, DataFrame, PingFrame,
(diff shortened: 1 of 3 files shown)


Lukasa:
FWIW, I believe the bump on hyperframe dependency means we need to bump the major version number of hyper-h2. Just a thing to be aware of here, but this makes me want to delay the hyper-h2 release until I've fixed a few more bugs, just to make the jump to v2 worth it.

PR #86, opened 2016-01-10
fix exception args

Kriechi:
This fixes the arguments to create a new `StreamIDTooLowError` exception.
I'm not sure how that got past this test: https://github.com/python-hyper/hyper-h2/blob/master/test/test_exceptions.py#L12

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -336,9 +336,7 @@ class H2Connection(object):
         )
 
         if stream_id <= highest_stream_id:
-            raise StreamIDTooLowError(
-                "Stream ID must be larger than %s", highest_stream_id
-            )
+            raise StreamIDTooLowError(stream_id, highest_stream_id)
 
         if allowed_ids != AllowedStreamIDs.ANY:
             if (stream_id % 2) != int(allowed_ids):


sigmavirus24:
:+1: but the test you link doesn't test the code you're modifying. A test that exercises `begin_new_stream` would *possibly* catch that, but that doesn't seem entirely likely since the test would probably have just tested that the exception was raised, not the exception message.

Still, this is a good catch!

sigmavirus24:
Actually, a test for `begin_new_stream` probably would have caught this as a result of trying to interpolate a string where `%d` is in the format string. Unfortunately there doesn't seem to be a test module for the connection module. I'll wait for @Lukasa to weigh in, otherwise I'd be in favor of merging this ASAP.

Lukasa:
Great catch!

The reason we didn't see this in testing is that we didn't ever actually print the object or test its values. That's because I suspect we actually only validate that a ValueError or ProtocolError is raised in that case, and so don't check the storage. Let me find the test that tests this logic so we can update it appropriately.

Lukasa:
Yup, so this line of code is tested by both `TestBasicClient.test_cannot_send_headers_on_lower_stream_id`, which tests that this exception is a ValueError, and `TestClosedStreams.test_receiving_low_stream_id_causes_goaway`, which tests that it's a ProtocolError. I think I'd like this PR to switch both of those tests to confirming that it's a StreamIDTooLowError, and validate the contents of that error (though there's no need to validate the string formatting, that's tested by the test you linked to @Kriechi). While we're here, let's remove the `ValueError` in the inheritance tree for the class, because I'm going to push a major release soon and so we can break that now.

@sigmavirus24, for the future, the reason it looks like there is no testing module for the `connection` module is that the code here is as much as possible tested by feeding it real byte strings or events, rather than testing the classes in isolation. Basically, if we can't reach a line of code by giving the public API the right inputs, then that code should almost certainly be removed. =)

sigmavirus24:
Ah @Lukasa. I haven't dug too much into the tests here. My first instinct is always to look for tests related to a module in a test module with a similar name. ;)

Lukasa:
See #95 for my follow-up to this with the changes I suggested. =)

PR #87, opened 2016-01-10
fix highest stream id check for push promise

Kriechi:
I think there is a problem with regard to PushPromise and the stream ids. Right now there is a value in `H2Connection` called: `highest_stream_id`. It is used to check if a new stream has a higher id than the previous highest value. As far as I understand the spec, this is only considered for the client and server direction independently.

The spec says:
> The identifier of a newly established stream MUST be numerically
> greater than all streams that the initiating endpoint has opened or
> reserved.

This PR is a rough fix, not sure if we get other problems - I'm welcome for alternatives or cleaner solutions. I'm not sure if this fixes everything, as I did not yet get PushPromises working in my code base.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -234,7 +234,7 @@ class H2Connection(object):
     def __init__(self, client_side=True):
         self.state_machine = H2ConnectionStateMachine()
         self.streams = {}
-        self.highest_stream_id = 0
+        self.highest_stream_id = [0, 0] # [server-initiated, client-initiated]
         self.encoder = Encoder()
         self.decoder = Decoder()
         self.client_side = client_side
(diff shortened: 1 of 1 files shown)


Lukasa:
Ah good spot, you're totally right! I think the general shape of this fix is fine but if you'd rather hold off on it I can take a run at this tomorrow. I think I'd split these out into two properties rather than use a list, just for some explicitness.

Kriechi:
I had the same idea at first, but then decided against it, because it of too many `if`s for my taste. 

Happy to wait - then I won't have to deal with fixing flake8 errors in this PR :smile:

Lukasa:
=P Ok, I'll take a swing at this tomorrow.

Lukasa:
Fixed in #92.

PR #93, opened 2016-01-11
Add get_next_available_stream_id function.

### LINKED ISSUE #91: Get next available stream_id to begin a new stream
Hi,

I want to create a new stream without knowing or generating a new stream_id myself. 
The current problem is that:
* `begin_new_stream`
* `get_or_create_stream`
* `send_headers`

all require a `stream_id` - which I don't know at this point. 

I would expect hyper-h2 to provide something like: `send_headers(headers, end_stream=True)` (note there is no `stream_id` argument!) or simply `stream_id = h2conn.get_next_stream_id()` as well, as long as it is the next valid id available.

Lukasa:
This function returns the next available stream ID for the given endpoint type, client or server. It makes it moderately easier to write a hyper-h2 server, even though it's not strictly necessary, and it also provides useful information about overly large stream IDs.

Resolves #91.

/cc @Kriechi

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -89,6 +89,9 @@ Protocol Errors
 .. autoclass:: h2.exceptions.InvalidSettingsValueError
    :members:
 
+.. autoclass:: h2.exceptions.NoAvailableStreamIDError
+   :members:
+
 
 HTTP/2 Error Codes
 ------------------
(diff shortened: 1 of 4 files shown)


Kriechi:
LGTM, thanks!

One question about PushPromise:
I think I need a similar mechanism for inbound stream ids as well, not sure yet...

Lukasa:
I don't think you do. With pushed streams, the server assigns them a new stream ID from its pool of stream IDs. This code will correctly return the next one allowed for a server. You can see one of the new tests I added for code that shows how this'd work for push promises.

Kriechi:
Yes, but in the case of mitmproxy, I am the server.
I will give it a try later today.

PR #94, opened 2016-01-11
Report the proper last stream ID in GOAWAY frames.

### LINKED ISSUE #76: Keep better track of last stream ID
For sending GOAWAY frames we do an ad-hoc calculation of the last stream ID based on the active streams at the time. This is obviously a flawed calculation, and we should be able to do better.

Lukasa:
Resolves #76.


Lukasa:
Hurrah for negative diffs!

Issue #90, opened 2016-01-11
Restrict the public API of H2Connection


Lukasa:
Right now H2Connection has a few too many methods as part of its public API, including `begin_new_stream` and `get_or_create_stream`. These are really artifacts of the internal structure of the class, and so we should "privatise" them by adding a leading underscore.

Lukasa:
Resolved in #97.

PR #97, opened 2016-01-11
Privatise some of the public API.

Lukasa:
This change makes a number of confusing methods "private" to avoid confusion.

In return, the documentation on the remaining public methods is greatly improved.


Lukasa:
Resolves #90.

PR #99, opened 2016-01-11
Fully support CONTINUATION frames

Lukasa:
This change is a bit of a monster to review, but looking at each commit in isolation should tell the story. In order to get this to work I had to push some of the frame length checking logic into the frame buffer where previously it was in the connection, but that's probably just fine.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,3 +7,4 @@ show_missing = True
 exclude_lines =
     pragma: no cover
     .*:.* # Python \d.*
+    assert False, "Should not be reachable"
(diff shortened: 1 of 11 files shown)


Lukasa:
There's one thing we're missing here: if a CONTINUATION frame is received *without* a preceding header block then it slips through into the state machine, which barfs on it. We should probably enforce that that doesn't happen: another validation in the FrameBuffer might be the way to go.

Lukasa:
Ugh, no, that won't work, because we need to reject them on CLOSED streams too. New plan: catch it on the H2Connection object itself.

PR #104, opened 2016-01-12
do not return empty event lists

### LINKED ISSUE #103: Consider removing the empty event lists returned by most function calls.
Initially we were being very defensive about the possibility that the user-level function calls (e.g. `send_headers` and friends) might want to return events. None of them do, and it seems really unlikely that any will. We should consider abandoning that pattern.

Kriechi:
fixes #103

### DIFF
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -534,7 +534,13 @@ class H2Stream(object):
         """
         # Because encoding headers makes an irreversible change to the header
         # compression context, we make the state transition *first*.
-        events = self.state_machine.process_input(StreamInputs.SEND_HEADERS)
+
+        # This does not trigger any events.
+        events = self.state_machine.process_input(
+            StreamInputs.SEND_HEADERS
+        )
+        assert not events
+
         hf = HeadersFrame(self.stream_id)
         frames = self._build_headers_frames(headers, encoder, hf)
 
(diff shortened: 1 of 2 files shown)


h2/stream.py
```diff
@@ -572,10 +572,9 @@ def locally_pushed(self):
         immediately after initialization. Sends no frames, simply updates the
         state machine.
         """
-        events = self.state_machine.process_input(
-            StreamInputs.SEND_PUSH_PROMISE
-        )
-        return [], events
+        # This does not trigger any events.
+        self.state_machine.process_input(StreamInputs.SEND_PUSH_PROMISE)
```
Lukasa: For any call into the state machine, can we assert on the return value? This'll stop us getting this wrong or accidentally throwing away anything we need.

Lukasa:
This mostly looks reasonable. Just one note. =)

h2/stream.py
Kriechi: better now?

h2/stream.py
```diff
@@ -534,7 +534,13 @@ def send_headers(self, headers, encoder, end_stream=False):
         """
         # Because encoding headers makes an irreversible change to the header
         # compression context, we make the state transition *first*.
-        events = self.state_machine.process_input(StreamInputs.SEND_HEADERS)
+
+        # This does not trigger any events.
+        events = self.state_machine.process_input(
+            StreamInputs.SEND_HEADERS
+        )
+        assert len(events) == 0
```
Lukasa: Let's keep this idiomatic: `assert not events`.

Lukasa:
Better! One small style change. Also, the doc strings might be better if rather than deleting the `:returns:` stanza we changed it to ’:returns: Nothing`. Thoughts?

Lukasa:
Looks great, thanks!

PR #121, opened 2016-01-20
Remove need to acknowledge settings changes.

### LINKED ISSUE #88: Auto-ACK Settings
Hi,

In your docs it [is stated](https://github.com/python-hyper/hyper-h2/blob/ec5fb452e3f79a29dc36c6cdd77a3d3d50afd710/h2/events.py#L97-L99):
> In HTTP/2, settings changes need to be acknowledged. hyper-h2 does not
> automatically acknowledge them, because it is possible that the caller may
> not be happy with the changed setting (or would like to know about it).

Could you elaborate on why the caller "might not be happy" with the changes?
There is no negotiation of settings, either you ACK them, or you get a GoAway or something, as far as I understand it? Creating a `RemoteSettingsChanged` by itself is nice to have, but manually having to ACK them is IMHO not necessary, it would be cleaner to handle that the same way pings are auto-handled.

Or am I missing some important edge-case?
Thanks for the explanation.

P.S. I know it is only 2 LOCs, but still - if we could put these lines into `hyper-h2` I think it is easier to use.

Lukasa:
Resolves #88.

/cc @Kriechi

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,10 @@ API Changes (Breaking)
 - Settings values set by both the user and the remote peer are now validated
   when they're set. If they're invalid, a new ``InvalidSettingsValueError`` is
   raised and, if set by the remote peer, a connection error is signaled.
+- Settings changes no longer require user action to be acknowledged: hyper-h2
+  acknowledges them automatically. This moves the location where some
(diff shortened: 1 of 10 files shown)


Kriechi:
LGTM! Thanks very much!

PR #123, opened 2016-01-21
Police content lengths when provided.

### LINKED ISSUE #114: Consider policing content-length.
This is an interesting idea: should we reject invalid content-lengths? RFC 7540 § 8.1.2.6 says so:

> A request or response that includes a payload body can include a content-length header field.  A request or response is also malformed if the value of a content-length header field does not equal the sum of the DATA frame payload lengths that form the body.
>
> Malformed requests or responses that are detected MUST be treated as a stream error (Section 5.4.2) of type PROTOCOL_ERROR.

This suggests that we should be keeping track of this. However, I'm a bit nervous that tools like mitmproxy may want to let the invalid stream stand, despite RFC 7540 warning against this:

> Intermediaries that process HTTP requests or responses (i.e., any intermediary not acting as a tunnel) MUST NOT forward a malformed request or response.

The other option would be to require that the policing of the content length be done by the tools that wrap hyper-h2.

@Kriechi @mhils thoughts?

Lukasa:
Resolves #114.

/cc @Kriechi @mhils It'd be great if you would confirm that this really is ok for mitmproxy. If you're currently unsure I'd like to err on the side of including this, and we can always add a flag to disable it if it turns out to be problematic.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -115,6 +115,10 @@ Protocol Errors
    :show-inheritance:
    :members:
 
+.. autoclass:: h2.exceptions.InvalidBodyLengthError
+   :show-inheritance:
+   :members:
+
 
 HTTP/2 Error Codes
 ------------------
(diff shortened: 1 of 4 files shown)


mhils:
As indicated before, I'm OK with this. :smiley:

@Kriechi: this may actually make it easier to implement mitmproxy's max_body_len.

Kriechi:
@mhils right now I'm just receiving data frames and check the total length. Works fine as well.
I think this should be ok as well - merge it and we will see what happens! :shipit:

PR #138, opened 2016-01-31
Don't allow SETTINGS_INITIAL_WINDOW_SIZE to update the connection flow control window.

### LINKED ISSUE #137: flow control impl is different from the RFC
While trying to write a tornado http2 client with hyper-h2==2.0.0, nginx==1.9.10, 
H2Connection throws FlowControlError('May not increment flow control window past 2147483647',) after nginx sends two frame at the beginning:

1. SettingsFrame(Stream: 0; Flags: None):  {'body_len': 18, 'stream_id': 0, 'flags': , 'settings': {3: 4096, 4: 2147483647, 5: 16777215}}
2. WindowUpdateFrame(Stream: 0; Flags: None): 7fff0000 {'body_len': 4, 'stream_id': 0, 'flags': , 'window_increment': 2147418112}

The implementation was:
https://github.com/python-hyper/hyper-h2/blob/master/h2/connection.py#L846
which applies SETTINGS_INITIAL_WINDOW_SIZE to both connection and every each existing streams.

After reading the RFC, I found some points:
https://http2.github.io/http2-spec/#SETTINGS_INITIAL_WINDOW_SIZE

1. Both endpoints can adjust the initial window size for **new streams** by including a value for SETTINGS_INITIAL_WINDOW_SIZE in the SETTINGS frame that forms part of the connection preface. **The connection flow-control window can only be changed using WINDOW_UPDATE frames**.
1. A SETTINGS frame cannot alter the connection flow-control window.
1. SETTINGS frames always apply to a connection, never a single stream. The stream identifier for a SETTINGS frame MUST be zero (0x0).

And I think SETTINGS_INITIAL_WINDOW_SIZE should NOT applies to existing streams' window size, which should be updated only WINDOW_UPDATE frame.

Lukasa:
Resolves #137.

@boyxuper Thanks so much for reporting this. Please take a quick look at this diff, particularly at CONTRIBUTORS.rst to check that you're happy with how you're being credited. Then feel free to test your code against the `master` branch when this change is merged to confirm that your initial set of problems are resolved.

I'm looking at releasing 2.1.0 sometime early next week.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -25,3 +25,8 @@ In chronological order:
 - Thomas Kriechbaumer (@Kriechi)
 
   - Fixed incorrect arguments being passed to ``StreamIDTooLowError``.
+
+- WeiZheng Xu (@boyxuper)
+
+  - Reported a bug relating to hyper-h2's updating of the connection window in
+    response to SETTINGS_INITIAL_WINDOW_SIZE.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -16,6 +16,9 @@ Bugfixes
 
 - Hyper-h2 now correctly accounts for padding when maintaining flow control
   windows.
+- Resolved a bug where hyper-h2 would mistakenly apply
+  SETTINGS_INITIAL_WINDOW_SIZE to the connection flow control window in
+  addition to the stream-level flow control windows.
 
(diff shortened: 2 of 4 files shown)


boyxuper:
Thanks for the fast fix, It's my pleasure to contribute.

PR #147, opened 2016-02-03
Debug representation for all events.

### LINKED ISSUE #146: Implement __repr__ for events
We're currently debugging our mitmproxy http2 stuff - it would be very handy if the hyper-h2 events would have a human-readable `__repr__` that includes all their attributes.

@Lukasa, any chance to get that implemented? :smiley:

Lukasa:
Resolves #146.

### DIFF
--- a/h2/events.py
+++ b/h2/events.py
@@ -25,6 +25,11 @@ class RequestReceived(object):
         #: The request headers.
         self.headers = None
 
+    def __repr__(self):
+        return "<RequestReceived stream_id:%s, headers:%s>" % (
+            self.stream_id, self.headers
+        )
+
 
 class ResponseReceived(object):
     """
@@ -38,6 +43,11 @@ class ResponseReceived(object):
 
         #: The response headers.
         self.headers = None
+
+    def __repr__(self):
+        return "<ResponseReceived stream_id:%s, headers:%s>" % (
+            self.stream_id, self.headers
+        )
 
 
 class TrailersReceived(object):
@@ -55,6 +65,11 @@ class TrailersReceived(object):
 
         #: The trailers themselves.
         self.headers = None
+
+    def __repr__(self):
+        return "<TrailersReceived stream_id:%s, headers:%s>" % (
+            self.stream_id, self.headers
+        )
 
 
 class DataReceived(object):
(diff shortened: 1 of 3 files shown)


h2/events.py
```diff
@@ -149,6 +183,9 @@ def __init__(self):
         #: The data included on the ping.
         self.ping_data = None
 
+    def __repr__(self):
+        return "<PingAcknowledged ping_data:%s>" % self.ping_data
```
Kriechi: Is `ping_data` supposed to be text or raw bytes? Maybe encode in hex?

Lukasa: Mm, that's a good call.

h2/events.py
```diff
@@ -212,6 +267,11 @@ def __init__(self):
         #: the changed settings.
         self.changed_settings = {}
 
+    def __repr__(self):
+        return "<SettingsAcknowledged changed_settings:%s>" % (
+            self.changed_settings,
```
Kriechi: A Settings-ACK is not supposed to have any `changed_settings`:
> When this bit is set, the payload of the SETTINGS frame MUST be empty.

Lukasa: That's true, but that's not the settings that are carried on this event: this event carries the settings that were acknowledged, not what was in the SETTINGS frame. =)

Kriechi: Ok, but then the name is misleading. `self.acknowledged_settings` seems more intuitive to me.

h2/events.py
```diff
@@ -76,6 +91,15 @@ def __init__(self):
         #: than ``len(data)``.
         self.flow_controlled_length = None
 
+    def __repr__(self):
+        return (
+            "<DataReceived stream_id:%s, "
+            "flow_controlled_length:%s, "
+            "data:%s>" % (
+                self.stream_id, self.flow_controlled_length, self.data[:20]
```
Kriechi: Probably encode `data` here as well. Might be gzipped or something binary.

h2/events.py
Lukasa: I'm on the fence there. The settings you change don't actually take effect until they're acknowledged: that is to say, the change is not applied until this event is received. I can see either way, and right now I'm inclined to want to leave the name unchanged rather than change it and rev hyper-h2 to 3.0.0.

Kriechi:
Thanks! This makes debugging so much easier for us!

h2/events.py
Lukasa: See #148.

PR #152, opened 2016-02-05
Specify the extra requirements in setup.cfg

### LINKED ISSUE #151: Cannot install hyper-h2 on Ubuntu LTS
Doing some tests for mitmproxy, I noticed that h2 doesn't install on either Ubuntu 12.04 or 14.04 by default: https://gist.github.com/mhils/d97ca425c33df644722f
We currently recommend mitmproxy users to just `pip install mitmproxy` in order to get a somewhat recent release - which doesn't work anymore with h2 as a dependency.

I know the whole thing sucks, but how can we address it best?

sigmavirus24:
This allows hyper-h2 to be installed with old versions of pip and
setuptools

Closes #151

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,6 @@
 [wheel]
 universal = 1
+
+[metadata]
+requires-dist =
+    enum34>=1.0.4,<1.1; python_version<"3.4"
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,4 @@ setup(
         'hyperframe~=3.1',
         'hpack~=2.0',
     ],
-    extras_require={
-        ':python_version<"3.4"': ['enum34~=1.0.4'],
-    }
 )


sigmavirus24:
Also, @Lukasa you might recall that we did this for [urllib3](https://github.com/shazow/urllib3/blob/master/setup.cfg#L16)

Lukasa:
Yeah, this does seem vaguely familiar.

Lukasa:
Of course, the failing tests suggest that maybe it doesn't work? =P

sigmavirus24:
Hm. I think this works mainly for wheels and that may be the problem with the tests.

Lukasa:
It'd be great if @dstufft could step in and share his wisdom at some point. =)

mhils:
Ubuntu 12 does not support wheels, so fixing them doesn't help much in this case.

PR #153, opened 2016-02-05
Support old setuptools versions

mhils:
This fixes the installation on Ubuntu 14.04 (but not 12.04). We must neither use `python_version<"3.4"` (that is setuptools 17.1 or so) nor `~=`.


Lukasa:
Thanks @mhils, and thanks for the assistance on this @sigmavirus24. :heart:

PR #155, opened 2016-02-07
GOAWAY additional data in ConnectionTerminated event

etataurov:
I think, it's handy to have access to additional data. i.e. Apple puts JSON with disconnect reason into GOAWAY frame https://developer.apple.com/library/ios/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/Chapters/APNsProviderAPI.html

RFC states
>  Endpoints MAY append opaque data to the payload of any GOAWAY frame.
   Additional debug data is intended for diagnostic purposes only and
   carries no semantic value.  Debug information could contain security-
   or privacy-sensitive data.  Logged or otherwise persistently stored
   debug data MUST have adequate safeguards to prevent unauthorized
   access.

so I didn't add additional_data to `__repr__` 

BTW, thanks for the repo. I really like that it is transport-agnostic

### DIFF
--- a/h2/events.py
+++ b/h2/events.py
@@ -334,6 +334,9 @@ class ConnectionTerminated(object):
         #: peer and so can safely be resent.
         self.last_stream_id = None
 
+        #: Additional debug data that can be appended to GOAWAY frame.
+        self.additional_data = None
+
     def __repr__(self):
         return "<ConnectionTerminated error_code:%s, last_stream_id:%s>" % (
             self.error_code, self.last_stream_id
--- a/h2/events.py
+++ b/h2/events.py
@@ -338,8 +338,15 @@ class ConnectionTerminated(object):
         self.additional_data = None
 
     def __repr__(self):
-        return "<ConnectionTerminated error_code:%s, last_stream_id:%s>" % (
-            self.error_code, self.last_stream_id
+        return (
+            "<ConnectionTerminated error_code:%s, last_stream_id:%s, "
+            "additional_data:%s>" % (
+                self.error_code,
+                self.last_stream_id,
+                _bytes_representation(
+                    self.additional_data[:20]
+                    if self.additional_data else None)
+            )
         )
 
 
@@ -352,6 +359,9 @@ def _bytes_representation(data):
     mainline of the code. It's safe to use in things like object repr methods
     though.
     """
+    if data is None:
+        return None
+
     hex = binascii.hexlify(data)
 
     # This is moderately clever: on all Python versions hexlify returns a byte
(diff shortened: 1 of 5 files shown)


Lukasa:
This looks really good, thanks so much! I'm happy to add this functionality.

Firstly, however, we have to get it through Travis, which has apparently lost the pull request. Let me see if I can re-trigger it.

Lukasa:
Ok, looks like you have some linter errors. =) Mind fixing those up for me?

etataurov:
I've fixed it.
I hadn't have python3.4 on my local machine, so testenv:list did not worked for me

Kriechi:
I would argue `__repr__` is mostly used for debugging purposes, and should therefore also show `additional_data` if available.

Lukasa:
Yeah, let's go with @Kriechi here: let's take a similar bit of logic to the `DATA` frame and show the first few bytes of additional data, or at least indicate that there is some to be seen.

etataurov:
Yeah, I agree.

I have doubt about `additional_data` None value. All other fields in other events have None as initial value in `__init__`, but set to specific value when event created. But `additional_data` will be None if frame's `additional_data` is empty.
I think it might be confusing and I'll better change it. So additional_data will be passed as-is from frame to event. What do you think?

Lukasa:
I think having `additional_data` as `None` is fine. =)

h2/events.py
```diff
@@ -334,9 +334,18 @@ def __init__(self):
...
     def __repr__(self):
-        return "<ConnectionTerminated error_code:%s, last_stream_id:%s>" % (
-            self.error_code, self.last_stream_id
+        return (
+            "<ConnectionTerminated error_code:%s, last_stream_id:%s, "
+            "additional_data:%s>" % (
+                self.error_code,
+                self.last_stream_id,
+                (_bytes_representation(self.additional_data[:20])
+                 if self.additional_data else None)
```
Lukasa: This has gotten just a little bit confusing. It might be better, and more robust, if we drag the `if None` check into `_bytes_representation`, such that calling `_bytes_representation(None)` returns `None` (which is about what you'd expect anyway).

Lukasa:
Alrighty this is nearly perfect. One small note and I'll be happy to merge!

h2/events.py
etataurov: But we are slicing self.additional_data here, so it needs to be checked here. 

Or _bytes_representation should accept some kind of `max_len` parameter, which may lead to confusion: Which string is this length of: input or output?

Or we can add check into _bytes_representation
```
if data is None:
    return None
```
and move parentheses a bit
```
"additional_data:%s>" % (
    self.error_code,
    self.last_stream_id,
    _bytes_representation(
        self.additional_data[:20]
        if self.additional_data else None)
)
```

IMHO last option is more explicit

Lukasa: Heh, yeah, good spot: I missed that we sliced first. Yeah, ok, let's just adjust the indentation as you suggest.

Lukasa:
Alright perfect, thank you so much @etataurov!

etataurov:
Thanks for merging!

PR #158, opened 2016-02-08
Use an SVG for Travis CI status

brettcannon:
Makes the badge look nicer on HDPI screens.


Lukasa:
Thanks a lot @brettcannon! :sparkles: :cake: :sparkles:

PR #170, opened 2016-02-15
Allow all enum34 1.x series

felixonmars:
Tests passed here with enum34 1.1.1 w/Python 2.7.11.


Lukasa:
In a strictly semantically versioned world we should be able to loosen this to any version less than 2. Want to give that a shot?

felixonmars:
I agree. Updated accordingly :)

Lukasa:
Thanks so much!

PR #173, opened 2016-02-16
Hoist PRIORITY handling to connection scope.

### LINKED ISSUE #168: Reconsider PRIORITY frame state machine flow.
Right now hyper-h2 passes PRIORITY frames to the stream state machine. This is an entirely understandable decision: PRIORITY frames technically are stream-specific and so it makes some degree of sense to want to pass them to that state machine.

However, PRIORITY frames don't actually *affect* the stream state machine in any way: they can be received and sent on streams in any state (even CLOSED), they cause no state transitions, and in general don't affect the state of any of the stream-level objects. However, they *do* complicate our stream-level state management because they can inadvertently cause us to resurrect dead streams and generally screw with us.

For this reason, I think we may actually want to hoist the logic for PRIORITY frames *up* one level to treat them as connection-only frames. This weirdly makes somewhat more sense: the connection has a much better understanding of the way the streams are configured, and while PRIORITY frames are stream-scoped they affect connection state.

This also tentatively affects #157, as the change in handling frames on reset streams applies here.

Lukasa:
Resolves #168.

This stops passing PRIORITY frames through the stream-level state machine. Instead, they are handled at the connection-level, without affecting stream state machines. This avoids problems with handling PRIORITY frames, as discussed in #168.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -19,7 +19,7 @@ from hpack.exceptions import HPACKError
 from .errors import PROTOCOL_ERROR, REFUSED_STREAM
 from .events import (
     WindowUpdated, RemoteSettingsChanged, PingAcknowledged,
-    SettingsAcknowledged, ConnectionTerminated
+    SettingsAcknowledged, ConnectionTerminated, PriorityUpdated
 )
 from .exceptions import (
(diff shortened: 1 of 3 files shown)


Lukasa:
Strictly this requires no notes in the changelog and can safely be backported.

PR #174, opened 2016-02-22
Update test_requirements.txt

felixonmars:
They are really outdated :)


Lukasa:
Thanks for this! It raises some questions about what requires.io is up to, since requires.io is supposed to be making pull requests to keep this up to date.

Regardless, thanks for spotting and improving it. =) :sparkles: :cake: :sparkles:

PR #178, opened 2016-03-08
Fix up example to use newer curio API.

kdart:
Also fix some bugs. Change remote flow control window check with local
one.

Now it doesn't throw an exception. However, it does stall on large
files so this is not a complete fix yet.

I also changed it so that you can specify alternate certs for testing from different interfaces and hosts.


examples/curio/curio-server.py
```diff
@@ -190,6 +191,13 @@ def __init__(self, sock, root):
...
-    kernel.add_task(h2_server(('', 8080), sys.argv[1]))
+    host = sys.argv[2] if len(sys.argv) > 2 else "localhost"
+    kernel = Kernel(with_monitor=True)
+    kernel.add_task(h2_server((host, 5000),
+                              sys.argv[1],
+                              "{}.crt.pem".format(host),
+                              "{}.key".format(host)))
+    print("Try GETting:")
+    print("    OSX after 'brew install curl --with-c-ares --with-libidn --with-nghttp2 --with-openssl':")
+    print("/usr/local/Cellar/curl/7.47.1/bin/curl --tlsv1.2 --http2 -k https://localhost:5000/bundle.js")
```
Lukasa: These notes are a little too specific. Maybe suggest the user point a browser at the host instead?

Lukasa: (Unless of course these are from your code and you didn't meant to commit them, in which case never mind!)

examples/curio/curio-server.py
```diff
@@ -190,6 +191,13 @@ def __init__(self, sock, root):
...
 
 if __name__ == '__main__':
-    kernel = Kernel()
-    kernel.add_task(h2_server(('', 8080), sys.argv[1]))
+    host = sys.argv[2] if len(sys.argv) > 2 else "localhost"
+    kernel = Kernel(with_monitor=True)
+    kernel.add_task(h2_server((host, 5000),
+                              sys.argv[1],
+                              "{}.crt.pem".format(host),
+                              "{}.key".format(host)))
```
Lukasa: This change adjusts the example to stop working with the cert and key that are stored in the repository. I think it'd be better if it didn't.

examples/curio/curio-server.py
kdart: Ok, but it can be hard to tell what is actually happening with browser. In fact, Safari also hangs with this, and it's hard to tell on what file it got hung.

Lukasa: Oh, sure, no question that that's true, but having instructions that are so platform specific isn't substantially better IMO.

examples/curio/curio-server.py
Lukasa: Oh, I just noticed that you changed the filenames. Ignore this feedback!

kdart: Really? I tried it (after renaming them, part of this change) and it worked. At least with curl it does. I can't check it now with a browser (since the cert is for localhost and now I'm remote).

Lukasa:
@kdart Ok, this looks great. Do you have an idea of how large a file causes the stall? If you do, I'll replicate it here and dive into it.

kdart:
The file is this:
-rw-r--r--  1 keith_dart  staff  1055730 Mar  4 17:42 bundle.js

About a megabyte of text.  I'm not sure yet how far into it it gets.

kdart:
Ah, stalls at right about 65K

```
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  6 1030k    6 65535    0     0   4006      0  0:04:23  0:00:16  0:04:07     0^C

-rw-r--r--  1 keith_dart  staff  61440 Mar  8 02:35 bundle_test.js
```

Lukasa:
@kdart Yeah, that seems somewhat inevitable. Let's see what's going on there.

Lukasa:
Alright, I've worked it out. The problem is in `window_updated`, specifically [here](https://github.com/python-hyper/hyper-h2/blob/master/examples/curio/curio-server.py#L187). The bug is that, once again, we yield the flow of control. This can cause the event we just set to be replaced with a new event, but then we throw the entire dictionary away, causing us to lose the event we need to set to unblock the stream.

It's enough to simply remove that, but it might be better to refactor that `elif` block entirely to:

```python
blocked_streams = list(self.flow_control_events.keys())
for stream_id in blocked_streams:
    event = self.flow_control_events.pop(stream_id)
    await event.set()
```

That should resolve your bug @kdart.

kdart:
Awesome, thanks! That does fix it. I also tried the alternative you suggested, but that doesn't work. It hangs again. It seems that dictionary is mutated while being looped over in this coroutine, right?

Lukasa:
@kdart My suggested alternative really does require that you get a *list* of keys in the dictionary right at the start, before popping anything out of that dictionary. If the code is identical to my suggested code it works fine.

kdart:
oh, right. Let me try that.

kdart:
So how does that look?

Lukasa:
Looks perfect! Thank you so much @kdart! :sparkles: :cake: :sparkles:

PR #181, opened 2016-03-12
Add intermediate-level docs.

Lukasa:
This branch tracks adding an intermediate-level worked example: specifically, building a WSGI server. The first step is to actually build such a thing: the next is to work backwards from the finished product into a description of how to build it.

Task items:

- [X] Write working example WSGI server
- [x] Write explanatory documentation

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -15,6 +15,7 @@ Example Servers
    twisted-example
    eventlet-example
    curio-example
+   wsgi-example
 
 Example Clients
 ---------------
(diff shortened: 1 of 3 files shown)


kdart:
This is really great! Thanks for the example.

Lukasa:
So, at this stage I am actually planning to leave the example in but remove the explanatory documentation requirement. Documenting, in detail, the building of this application looks like being upwards of 5000 words long *at least*, particularly if it's done in a way that is evolutionary (e.g. where the thing "works" at every stage). That, combined with the number of concepts introduced here, just makes the burden of writing and maintaining that document too much.

For that reason, I'm going to merge this as-is.

Issue #183, opened 2016-03-16
Remove deprecated fallback for dictionaries as headers.


Lukasa:
Remove the fallback that was deprecated in #184.

LuRsT:
Can anyone review this change please?

Lukasa:
@LuRsT This change looks good to me: however, there's currently no roadmap for the 3.0.0 release (it'll be released when we have other breaking changes).

You're welcome to open a PR with this change but it won't get merged until we're ready to go, so it'll likely drift from the current master pretty painfully over time. So instead, I recommend posting that patch in this issue (the output from `git diff`) will be fine, so that the patch doesn't get lost. =)

LuRsT:
Sure thing, also, part of what I did in that commit (changing the dict headers in the test) should be merged to avoid confusion right (if yes, I'll send a PR with just that change, if not, I'll keep it here with the rest)?

Lukasa:
@LuRsT How about we go one step further. Can you make the dict headers change, but also add a *new* test that explicitly tests that dicts are accepted? That gives us a single targetted test that can also be removed along with this change.

LuRsT:
@Lukasa Of course! Great idea, will do that!

LuRsT:
@Lukasa done, could you check if that's what you wanted?

Lukasa:
\o/ Fab!

Lukasa:
This is happening in #500.

PR #184, opened 2016-03-16
Deprecate the use of dictionaries for headers.

Lukasa:
Following on from the discussion on python-hyper/hpack#32, this change deprecates the conversion of headers from dictionaries into iterables of two-tuples. The issue for actually removing this deprecation is in #183.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -21,6 +21,12 @@ Bugfixes
   explicitly overridden. This is a safe defensive initial value for this
   setting.
 
+Deprecations
+~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


derekchentrendmicro:
now I need to pass headers this way
```
headers =  [(':status', 200),
            ('server', 'basic-h2-server/1.0'),
            ('content-type','application/json'),
            ('content-length', len(response_data))]

conn.send_headers(stream_id=stream_id,headers=headers)
```
is this the expected usage?

Lukasa:
Yes, it is. =)

PR #192, opened 2016-04-05
Allow informational responses in the HALF_CLOSED_* states.

Lukasa:
This must have just been an oversight, but we were forbidding 1XX responses in the half closed local and half closed remote states. There's no reason for that: they're entirely allowed.

This patch isn't complete yet: we need to tighten up our processing to ensure that informational headers are never sent/received after the actual response headers.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,14 @@ Release History
 dev
 ---
 
+Bugfixes
+~~~~~~~~
+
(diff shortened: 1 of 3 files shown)


Lukasa:
Alright, that now completes the patch.

Lukasa:
This patch was backported to 2.2.X, but not to 2.1.X as the fix applies to the informational responses that were added in 2.2.0.

Issue #154, opened 2016-02-07
Consider having received headers be bytes, not unicode strings.


Lukasa:
Currently the `RequestReceived`, `ResponseReceived`, and `TrailersReceived` events all return the headers as a list of tuples of unicode strings. This happens because HPACK decodes the headers as UTF-8 before use. While this is enormously convenient when it works, which is almost all the time, it introduces the risk that a senseless HTTP/2 implementation may send non-UTF-8-encoded data and have hyper-h2 barf on it.

For the 2.x series we're stuck with this design decision, but when we come to look at doing a 3.0.0 we may want to revisit this decision.

Kriechi:
I think this is fully covered in #198 - and the only remaining thing to discuss/decide is to change the default encoding from `utf-8` to *no encoding* in a 3.0 release.

Personally I have no strong opinion, since changing it is a simple flag on the connection. However I'm slightly leaning toward going raw bytes-only for all internals, just so API users need to think about what they expect to get and what they actually want.

Lukasa:
Agreed, this issue is basically whether or not we change the default. I remain +1 on changing the default: I think unicode-by-default was the wrong choice, though it seemed sensible back when I first made that call in 2014.

PR #198, opened 2016-04-13
Allow users to obtain raw headers.

Lukasa:
This change adds a new flag that allows users to request 'raw' headers: that is, headers that have not been decoded to unicode and are still in bytes form. This was originally requested by the Twisted developers who use a bytes internal headers representation, though it may well be useful for others.

This change involves quite a substantial internal change: we now have headers move through the body of hyper-h2 in bytes form, and only get encoded at the edge.

This change introduces one unfortunate problem: we need to bump the minimum version of HPACK we can use. I believe this is safe to do in a semver-minor release, as we're only bumping the minor version, so the upgrade should be safe.

The default output for headers *remains* unicode. That may change in 3.0.0 (per #154), but for now we're leaving it as-is.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,16 @@ Release History
 dev
 ---
 
+API Changes (Backward-Compatible)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Added a new flag to the ``H2Connection`` constructor: ``header_encoding``,
+  that controls what encoding is used (if any) to decode the headers from bytes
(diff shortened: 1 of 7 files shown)


hawkowl:
This looks pretty good to me :+1:

mhils:
Our [netlib.http.Headers](https://github.com/mitmproxy/mitmproxy/blob/96782407833a7f827e79fa7aa5f2576f1bee3547/netlib/http/headers.py#L32) class works on bytes as well, so this change is a very welcome addition. :smiley:

Looks good to me as well! :+1:

Lukasa:
Ok, I'm going to land this now.

PR #199, opened 2016-04-14
Disallow custom pseudo-headers

LuRsT:
I had a look at `hypothesis` and realized that the test that was already present in `TestFilter` is testing the "good case", so I only created a test for an invalid pseudo-header, correct me if I'm wrong, but I found no way to do something like what I did in the test in a `hypothesis` way.

### DIFF
--- a/test/test_invalid_headers.py
+++ b/test/test_invalid_headers.py
@@ -102,3 +102,8 @@ class TestFilter(object):
             assert headers == h2.utilities.validate_headers(headers)
         except h2.exceptions.ProtocolError:
             assert True
+
+    def test_invalid_pseudo_headers(self):
+        headers = [(b':custom', b'value')]
+        with pytest.raises(h2.exceptions.ProtocolError):
+            h2.utilities.validate_headers(headers)
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -22,6 +22,15 @@ CONNECTION_HEADERS = set([
 ])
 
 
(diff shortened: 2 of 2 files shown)


test/test_invalid_headers.py
```diff
@@ -102,3 +102,8 @@ def test_range_of_acceptable_outputs(self, headers):
             assert headers == h2.utilities.validate_headers(headers)
         except h2.exceptions.ProtocolError:
             assert True
+
+    def test_invalid_pseudo_headers(self):
+        headers = ((b':custom', b'value'))
```
Lukasa: I don't think this actually creates a tuple within a tuple. Consider changing this to a list with one tuple entry. That's why your test is failing.

Lukasa:
One small note regarding the tests.

test/test_invalid_headers.py
LuRsT: Damn, fail, I did fix this but then I undid it, I'll fix this :+1:

LuRsT:
I force pushed it, don't know if that's the way to go, but wanted a cleaner commit, could you ask travis-ci to re-check it please?

LuRsT:
Disregard, travis-ci is smart :)

Lukasa:
Hooray! Looks great to me, thank you so much! :sparkles: :cake: :sparkles:

LuRsT:
Great! Thanks for your help @Lukasa, and expect more contributions for me :D

PR #206, opened 2016-04-22
Also never-index proxy-authorization

rbtcollins:


Lukasa:
Thanks for this, I definitely missed this.

I'm going to merge this manually, because I'll also add some tests to ensure that we don't regress this. And then I'm going to consider adding a new API to allow this to be configurable, to allow clients/servers that know better than us to ensure that nothing falls through the cracks.

PR #209, opened 2016-04-23
Use code-block in readme

kragniz:
This looks at least 10% cooler.


Lukasa:
More like 20%! Thanks! :sparkles: :cake: :sparkles:

PR #215, opened 2016-04-29
[requires.io] dependency update on master branch

Lukasa:


rbtcollins:
So since this is automated, should we just hit merge?

rbtcollins:
Oh... I can't hit merge, nvm :)

Lukasa:
@rbtcollins You should have been able to merge this if it passes tests. Were they passing?

sigmavirus24:
@Lukasa they were

rbtcollins:
I could 'Close Pull Request' or 'Comment', but not 'Merge'. Merge is
apparently restricted to folk with write access.

Lukasa:
@rbtcollins You should have write access, but let me double check.

PR #217, opened 2016-04-30
Fixed test using dict headers

LuRsT:
Also added test to ensure we allow dictionaries implicitly

Related to #183


Lukasa:
\o/ This is magnificent, thank you @LuRsT! :sparkles:

PR #220, opened 2016-05-02
Use set literals where applicable

berkerpeksag:
2.7 is the minimum Python version listed in setup.py
so set([...]) syntax can be replaced with {...}
syntax safely.


Lukasa:
Hooray! Thank you so much. :sparkles: :cake: :sparkles:

PR #221, opened 2016-05-03
Remove redundant condition in settings.py

alexwlchan:
* If value is negative, then the first condition is True. The second
  condition is never checked, and we return an error.
* If value is zero or positive, then both conditions are False, and
  we return 0.

In either case, we gain no new information from the second check, so
it can be removed.


Lukasa:
Nice catch @alexwlchan, thanks! :sparkles:

PR #231, opened 2016-05-26
add more data to close_connection

Kriechi:


test/helpers.py
```diff
@@ -110,7 +110,7 @@ def build_ping_frame(self, ping_data, flags=None):
         return f
 
     def build_goaway_frame(self,
-                           last_stream_id,
+                           last_stream_id=0,
```
Lukasa: Can we leave this passed explicitly? I know it's an annoyance in the test, but I'd rather have the explicitness.

test/helpers.py
```diff
@@ -119,7 +119,7 @@ def build_goaway_frame(self,
         f = GoAwayFrame(0)
         f.error_code = error_code
         f.last_stream_id = last_stream_id
-        f.additional_data = additional_data
+        f.additional_data = additional_data or b''
```
Lukasa: This fallback should be unnecessary.

Lukasa:
I've got a couple of small notes, but this otherwise looks really good! :sparkles:

test/helpers.py
Kriechi: Done.

test/helpers.py
Kriechi: I wanted to test all three cases: https://github.com/python-hyper/hyper-h2/pull/231/files#diff-19d167383ddbd75543abdbc92e6b0698R853
Or is that also not necessary?

Lukasa: Ah, I see what's happened here. It might be better, rather than adding this fallback, to parametrize two arguments: the input value, and the expected result. That makes the test more explicit about what's going on, rather than hiding this in the frame factory.

h2/connection.py
```diff
@@ -860,19 +860,28 @@ def reset_stream(self, stream_id, error_code=0):
...
+
         """
         Close a connection, emitting a GOAWAY frame.
 
         :param error_code: (optional) The error code to send in the GOAWAY
             frame.
+        :param additional_data: (optional) Additional debug data indicating
+            a reason for closing the connection.
+        :param last_stream_id: (optional) The last stream which was processed
+            by the sender. Defaults to `highest_inbound_stream_id`.
```
Lukasa: This needs a double-backtick.

h2/connection.py
```diff
@@ -860,19 +860,28 @@ def reset_stream(self, stream_id, error_code=0):
...
         :returns: Nothing
         """
         self.state_machine.process_input(ConnectionInputs.SEND_GOAWAY)
 
-        f = GoAwayFrame(0)
-        f.error_code = error_code
-        f.last_stream_id = self.highest_inbound_stream_id
+        f = GoAwayFrame(
+            stream_id=0,
+            last_stream_id=(last_stream_id or self.highest_inbound_stream_id),
```
Lukasa: This doesn't *quite* work as expected: if you explicitly pass `last_stream_id=0`, this will cause hyper-h2 to emit the frame with `last_stream_id = self.highest_inbound_stream_id`. We may need a more complex statement outside the initializer, and then also some tests for this new behaviour (which is arguably unrelated to the test for additional data).

Lukasa:
Arg, sorry, found a subtlety here that needs addressing. Sorry I didn't see it the first time around!

h2/connection.py
Kriechi: I think I fixed it. Please take another look.

Lukasa:
Awesome, this looks great, thanks so much @Kriechi! :sparkles:

PR #233, opened 2016-05-28
Emit priority data

### LINKED ISSUE #229: Allow depends_on and weight settings in send_headers
Currently the `send_headers` function on the connection (and stream) object does not allow us to specify a `depends_on` and `weight` setting. It would be useful to create a single HeadersFrame with these values, instead of creating an additional PriorityFrame.

A related problem is that a `RequestReceived` event does not carry these values - but instead a second event (`PriorityUpdated`) is issued - which makes it unnecessary difficult to reassemble the original HeadersFrame. This might also be related to #228 for handling multiple events of the same stream etc.

Right now, once I receive a `RequestReceived` event, I'm looping through all remaining events to find possible `PriorityUpdated` which I would like then to merge all values into a new `send_headers` call.

Let me know if this is a feasible request - or maybe I'm just missing a nifty API feature.
Thanks.

**edit**
How can change the priority of a stream? I couldn't find a way to generate a PriorityFrame - or do I have to do that manually with raw frames?

Lukasa:
Resolves #229.

/cc @Kriechi 

I'm a bit nervous about the fact that the kwargs are different in `send_headers` and `prioritize`, but otherwise I think this is ok.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -11,6 +11,10 @@ API Changes (Backward-Compatible)
   user to send additional debug data on the GOAWAY frame.
 - Adds ``last_stream_id`` to ``H2Connection.close_connection``, allowing the
   user to manually control what the reported last stream ID is.
+- Add new method: ``prioritize``.
+- Add support for emitting stream priority information when sending headers
+  frames using three new keyword arguments: ``priority_weight``,
+  ``priority_depends_on``, and ``priority_exclusive``.
 
 Bugfixes
 ~~~~~~~~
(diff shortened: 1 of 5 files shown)


Kriechi:
LGTM! Thanks for the fast response and implementation! 🎉 

Should we also add Priority information upon receiving (i.e. generating a `RequestReceived`)?
https://github.com/python-hyper/hyper-h2/blob/3e3ce61824e4f5d9bc4732aae72cf44b15f3fbc8/h2/stream.py#L160

Lukasa:
Yeah, probably. =D Let's use a new ticket.

PR #237, opened 2016-06-09
Run Travis in a single test script.

Lukasa:
Turns out our tests have been failing on Python 3 for quite a while, and we failed to notice because Travis wasn't failing early. Let's change Travis to do the right thing, and then fix the scripts.


rbtcollins:
Whoops :). The set's were what we needed?

Lukasa:
Yeah, we needed the sets, but it seems like Travis has some concerns if you `set -e` in the YAML file.

Issue #234, opened 2016-05-30
RequestReceived should contain priority information


Kriechi:
A follow-up on #229 and #233:

It would be nice if the `RequestReceived` event contains priority information, if the HEADERS frame carries them. This way all available information about a new request is in the same event. `PriorityUpdated` is then only used for actual `PRIORITY` frames.

Lukasa:
So I'm not sure exactly how we want to do this yet. 

We could put the priority information in `RequestReceived`, but that means that priority information isn't obtained isomorphically regardless of the source: instead it needs to be found in different places depending on how the information was sent. I don't think I love that much.

I think I'd rather we decided on a solution for #228 that allows understanding this information without forcing people to know about it if they just want to handle the priority information wherever it came from.

Kriechi:
Maybe I misunderstood you, but you sounded more enthusiastic about it in https://github.com/python-hyper/hyper-h2/pull/233#issuecomment-222327671.

#228 sounds more like a 3.0 kinda thing. As a backward-compatible solution, putting priority information into `RequestReceived` is a good thing IMHO.

Lukasa:
@Kriechi We can *duplicate* the priority information in there, but we can't *move* it there, because that would also be a breaking API change.

Kriechi:
Yes, copying the values to the `RequestReceived` is fine.

Lukasa:
So then, the question becomes which of the following APIs we want to go with:

1. Add new data to `RequestReceived` that contains duplicate information from `PriorityUpdated`.
2. Add a link from `RequestReceived` to a `PriorityUpdated` event if they were emitted simultaneously.
3. Add support for emitting events as `frozenset` objects.

Doing 3 is a major version bump which we're considering anyway, so let's restrict ourselves to 1) and 2). Do you have a preference here?

Kriechi:
*1.* is probably the simplest and requires only minimal changes / additions to the API.
*2.* sounds like a nice idea - not sure how efficient this can be done. Ideally all events with the same `stream_id` are linked to each other.

Unless you really dislike duplicating the values between two events, I would prefer *1.*
But I'm also happy with *2.*

Lukasa:
Ok, I'm prototyping a solution based on (2), because it has the handy property of being a good stepping stone to potentially implementing @rbtcollins' "frozenset" based idea for #228: that is, it'll get people used to the idea that they may want to check a set for other things before processing too far.

Lukasa:
Resolved by #240.

Issue #228, opened 2016-05-18
[RFC]: Lazy state changes.


Lukasa:
Inspired by @njsmith in njsmith/h11#4.

Right now there is a wart in the h2 API when working on a busy connection. The problem exists when working with `receive_data`. For example, it is possible that you might encounter the following set of events returned from `receive_data`: [DataReceived(stream_id=1), WindowUpdated(stream_id=1), StreamReset(stream_id=1)].

In response to that WindowUpdated you may want to send another data frame on stream 1, but if you do so without processing the entire event list h2 will throw a `ProtocolError` at you because the stream is now in the CLOSED state. However, the user doesn't know that: they're processing the events serially, and haven't gotten to the event that actually closed the stream.

The outcome of this is not *too* bad: while the `ProtocolError` does get thrown, the connection state is still fine and no damage has been done to your ability to continue to use the connection. However, the exception may cause an incautious user to run into trouble: if they're not expecting and handling this exception it'll quite probably cause them to lose a lot of application state.

In njsmith/h11#4, @njsmith has proposed a split API for receiving data: a single `receive_data` call returns no events, but instead you need to call `next_event` to advance the state machine. This proposal is essentially a "split" API much like we have for sending.

Making that change is a pretty substantial API change and I don't know how I feel about it. *However*, it turns out that hyper-h2 is capable of emulating that change with a very simple change to the logic of `receive_data`.

Right now, `receive_data` returns a list of events. However, it would be a trivial change for `receive_data` to return a generator instead. That generator would then iterate across events, one event at a time. The effect of doing that would be to avoid updating the state machines until the user has an event in hand, and would essentially immediately change hyper-h2 to an API that *lazily* updates stream state. This would ensure that a user cannot see a `ProtocolError` for sending data on a closed stream unless they have iterated past an event that closed the stream (or they closed it themselves).

I'm interested in what @python-hyper/contributors think, as well as our wider user-base. I think making such a change would be extremely valuable, but it represents a breaking API change (`receive_data` is [documented](http://python-hyper.org/h2/en/stable/api.html#h2.connection.H2Connection.receive_data) as returning a `list`).

Do people think this change is sensible?

Lukasa:
[...]

PR #238, opened 2016-06-10
[RFC] Implement "related events" API.

Lukasa:
This is a potential solution to #234.

The problem we're trying to address here is twofold. Firstly, users like mitmproxy (represented by @Kriechi and @mhils) have a desire to reproduce frames as accurately as possible when they proxy the connection. This means that it's important to them to tell the difference between the two frame sequence of frames that goes `[HEADERS(END_HEADERS), PRIORITY]` and the one frame one that goes `[HEADERS(PRIORITY, END_HEADERS)]`.

Secondly, we want to begin to introduce to hyper-h2 this notion of events happening "simultaneously": that is, where multiple state transitions happen in such a way that they all occur at the same time, meaning that the user cannot reliably determine the allowed actions unless they have processed all the events that were emitted at that time. Mostly this is a pretty benign issue, but it would be nice to lazily consume events rather than emit them all at once (see #228 for more on this).

That makes this change both an attempt to address the first point cleanly *and* a potential precursor to the second. In essence, you can think of this change as something of a "beta" API for one of the proposed changes in #228.

In this specific change, events that can semantically occur *simultaneously* with each other now carry a `related_events` property. This property is a `frozenset` (immutability is good!) of all the events that occurred simultaneously.

This PR is a RFC because I'm not 100% sure I'm happy with the API as it stands. In particular, there are a few warts that I'd like feedback on.

1. Only some events have `related_events` fields. This is done primarily for clarity (to suggest that only some events can be affected by simultaneity), but it potentially means that code that works with the events gets pretty ugly, because sometimes accessing `event.related_events` will raise an `AttributeError`, based only on `type(event)`. So we may want to add `related_events` to *all* events.
2. `related_events` contains a `frozenset` of *all* the events that occurred at once, meaning that `event in event.related_events` is *always* `True`. That's simple and requires the least work in the lower-level library (which is good, because this is a hot section of code), but it leads to the possibility of surprises if the event is handled incautiously (e.g. if you process `event` and then process `event.related_events` you may process `event` *twice*). So we may want to create different frozensets for each event.

Anyway, I'd like people's feedback on this. @python-hyper/contributors, can I get your input here?

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1296,7 +1296,16 @@ class H2Connection(object):
 
         try:
             for frame in self.incoming_buffer:
-                events.extend(self._receive_frame(frame))
+                frame_events = self._receive_frame(frame)
+
+                # If more than one event popped out here, we want to process
+                # them as "related events".
+                if len(frame_events) > 1:
+                    event_set = frozenset(frame_events)
+                    for event in frame_events:
+                        event.related_events = event_set
+
+                events.extend(frame_events)
         except InvalidPaddingError:
             self._terminate_connection(PROTOCOL_ERROR)
             raise ProtocolError("Received frame with invalid padding.")
--- a/h2/events.py
+++ b/h2/events.py
@@ -23,6 +23,9 @@ class RequestReceived(object):
     .. versionchanged:: 2.3.0
        Changed the type of ``headers`` to :class:`HeaderTuple
        <hpack:hpack.HeaderTuple>`. This has no effect on current users.
+
+    .. versionchanged:: 2.4.0
+       Added ``related_events`` property.
     """
     def __init__(self):
         #: The Stream ID for the stream this request was made on.
@@ -30,6 +33,13 @@ class RequestReceived(object):
 
         #: The request headers.
         self.headers = None
+
+        #: Any events "related" to this one: that is, any events that occurred
+        #: at the exact same time as this one because their data was carried on
+        #: the same frame. All of these events need to be processed at once.
+        #:
+        #: .. versionadded:: 2.4.0
+        self.related_events = frozenset()
 
     def __repr__(self):
         return "<RequestReceived stream_id:%s, headers:%s>" % (
@@ -46,6 +56,9 @@ class ResponseReceived(object):
     .. versionchanged:: 2.3.0
        Changed the type of ``headers`` to :class:`HeaderTuple
        <hpack:hpack.HeaderTuple>`. This has no effect on current users.
+
+    .. versionchanged:: 2.4.0
+       Added ``related_events`` property.
     """
     def __init__(self):
         #: The Stream ID for the stream this response was made on.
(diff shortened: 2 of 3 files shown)


Kriechi:
Thanks for taking a swing at this! Really appreciated! 👍 

> related_events contains a frozenset of all the events that occurred at once

Why did you choose this behaviour? Sounds strange to me - on first thought...
Not sure why I would want to the the `event` again while looping through `related_events`

I have my API-user hat on at the moment:
If I get a `RequestReceived` event, I have to loop through `related_events` and call `isinstance` each time to check if I might have gotten a `StreamEnded` event. That's kinda annoying - unless you have a clean idea on how this API is intended to be used?

Since you already have a very nice and comprehensive definition of which events can be related with each other (in your test comments) - how about baking this directly into the events?
E.g. A `RequestReceived` event could have a `stream_endend = True/False` and `priority_weight`, ... attribute.

Or maybe I just have a very simplistic view for a general purpose API design? 😄

Lukasa:
> Since you already have a very nice and comprehensive definition of which events can be related with each other (in your test comments) - how about baking this directly into the events?

Yeah, so this is a thing I've been thinking about. What I'm most worried about is the question of aliasing. Specifically, if we emit `DataReceived` with `stream_ended=True`, should we then emit `StreamEnded`? If not, users are exposed to a subtle bug wherein they may not notice state changes. If we do, then users need to be alert for the possibility that they may receive multiple instances of redundant information.

This isn't so bad with `StreamEnded`, but it's potentially somewhat more problematic with `PriorityUpdated`, which could cause incautious users to process the same priority update twice, churning a whole chunk of state for no good reason.

One possibility is to have events have specific flags for their possible related events (e.g. a `DataReceived` event could have a `stream_ended_event` property that is optionally filled with the event that will follow `DataReceived`). This has all the same issues as above, but you can at least keep a temporary set around and check whether you've processed an event already.

Otherwise, I can't see a system based on "aggregating" events that isn't awkward. Frankly, the least awkward mode here might be to hang the triggering frame off any event fired by that frame. That at least gives mitmproxy what it needs to rebuild equivalent frames, though we *again* need to worry about the wacky internal gymnastics h2 performs with regard to HEADERS frames.

Kriechi:
> events have specific flags for their possible related events (e.g. a DataReceived event could have a stream_ended_event property that is optionally filled

Sounds good!

I'm not sure how other people are receiving data and passing it to h2, but in mitmproxy we only read one frame at a time, and process it. This makes dealing with events much simpler, because it keeps the dependencies between streams limited.

Lukasa:
How are you ensuring you read only one frame?

Kriechi:
We are parsing the frame header and then read the exact length of the frame from the socket.
https://github.com/mitmproxy/mitmproxy/blob/c801f81373d5935f60bf950b1f266911a94ecf60/netlib/http/http2/framereader.py#L6

Lukasa:
Ok, that makes sense. I think that's a relatively unusual approach: most implementations will just feed bytes directly into the state machine without checking where frame boundaries are. Similarly, however, most implementations don't really care about simultaneity except as it might cause unexpected `ProtocolError`s to show up.

So I remain a bit unsure about how best to move this forward, but we can also prototype the interface based on set properties and see how the implementation there looks.

Lukasa:
Alright, we decided to go with #240 instead.

Issue #246, opened 2016-06-29
Reject mismatched :authority and Host headers


Lukasa:
Related to #245, and we may want to merge the work together. Read that issue before tackling this one.

HTTP/2 allows for both the :authority and Host header to be present in a request header block. If they *are* both present, however, they absolutely should be the same. We should add a stage to the header validation pipeline (possibly the same as the stage in #245) that confirms that, if present, these are both identical. If they are not identical, that's a ProtocolError.

For @python-hyper/contributors, see the message in #245 for guidance on tackling this.

Lukasa:
This is partly resolved by #252. We now need only to reject *sending* headers that are mismatched in this way.

njsmith:
Interesting fact, recording here for posterity: traditionally servers like Apache have been known to dispatch based on the request line / `:authority` equivalent, while applications like Django have trusted the `Host:` header, and this has lead to real security bugs: http://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

So thanks to #252, this particular attack is not possible on h2-based servers.

Lukasa:
Yup, it's fairly important to police this. It's one of the advantages of the newness of HTTP/2: we can tighten this stuff up pretty easily without breaking too much!

mhils:
Just to clarify: We can still _send_ mismatched header/:authority values?

Lukasa:
For the moment, but we're working on a PR to stop that too.

mhils:
Hm - that is actually an ability we would like to retain for mitmproxy. Can we cater for this somehow?

Doesn't have to sacrifice the API usability, I'd be fine with us calling `h2._dont_check_host_mismatch = True` somewhere after import.

alexwlchan:
@mhils 

1. I think (I don’t have a working setup to try) it would be possible to skip this particular validation step by stubbing out the validation function in `h2.utilities`, something like:

    ```python
    import h2
    h2.utilities. _check_sent_host_authority_header = lambda headers: headers
    ```

    Which is a fiddly solution, but should work to disable any part of the validation pipeline (ingoing or outgoing) that you wanted to skip.

    I don’t know if @Lukasa is interested in offering anything less-fiddly than that. My gut feeling is against it, because it feels like it would add code complexity for what seems like a niche use case – but I don’t feel strongly.

    Out of curiosity, what’s the use case for being able to send mismatched :authority/Host headers?

[...]

Issue #245, opened 2016-06-29
Reject request header blocks with neither :authority nor Host headers.


Lukasa:
RFC 7540 Section 8.1.2.3 has complex rules regarding :authority and Host headers, but the basic gist of it is that there is very nearly no situation when you can have *neither* of them. Basically the only way this could happen is if a HTTP/1.0 request with no Host header was proxied into a HTTP/2 request. I'm pretty disinterested in serving this use-case at this time, so I think we should forbid both sending and receiving header blocks of this form.

Lukasa:
For @python-hyper/contributors, if you want to tackle this, you'll want to write another stage in the `validate_headers` header filter pipeline defined in `utilities.py`. That'll work for received headers. Sending headers have a much smaller filter pipeline, but at this point we should probably define a `validate_sent_headers` header filter pipeline that takes the two current stages of the sent pipeline (`_lowercase_header_names` and `secure_headers`) and adds some more.

This is a pretty good easy-to-intermediate task: it's a real problem that requires you to write real code, but it's pretty self-contained and easy enough to understand. If someone wants to work on it, assign it to yourself. Feel free to ask me to sketch out my proposed design in more detail.

alexwlchan:
I'm going to have a look at this today.

Lukasa:
\o/ Sounds good @alexwlchan, hit me up if you have questions!

Lukasa:
This is partly resolved by #252. We now only need to reject *sending* such blocks.

Lukasa:
Resolved.

PR #252, opened 2016-07-08
Reject request header blocks with bad :authority/Host headers

alexwlchan:
This is a WIP attempt at #245 and #246. So far I’ve done the case where we receive header blocks that don’t work under these rules; I haven’t tried to enforce it for header blocks that we send.

It turns out a lot of the example responses in the tests don’t include these headers – given what was said in the original issue, I’m assuming that was just an oversight?

Specific questions:

* Does it look okay so far?
* What's the appropriate action when we try to send header blocks that don’t meet these rules? Is there a way to auto-fill that information, or should we be raising an exception?
* Since this is part of the HTTP/2 spec, and not a hyper-specific feature, does it need any documentation?

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -35,7 +35,7 @@ from .settings import (
     MAX_CONCURRENT_STREAMS
 )
 from .stream import H2Stream
-from .utilities import validate_headers, guard_increment_window
+from .utilities import guard_increment_window
 
 
 class ConnectionState(Enum):
@@ -1387,7 +1387,6 @@ class H2Connection(object):
             # compatibility, catch all of them.
             raise ProtocolError("Error decoding header block: %s" % e)
 
-        headers = validate_headers(headers)
         events = self.state_machine.process_input(
             ConnectionInputs.RECV_HEADERS
         )
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -5,6 +5,7 @@ h2/utilities
 
 Utility functions that do not belong in a separate module.
 """
+import collections
 import re
 
 from hpack import NeverIndexedHeaderTuple
@@ -147,9 +148,19 @@ def authority_from_headers(headers):
     return None
 
 
-def validate_headers(headers):
+# Flags used by the validate_headers pipeline to determine which checks
+# should be applied to a given set of headers.
+HeaderValidationFlags = collections.namedtuple(
+    'HeaderValidationFlags',
+    ['is_client', 'is_trailer']
+)
+
+
+def validate_headers(headers, hdr_validation_flags):
     """
     Validates a header sequence against a set of constraints from RFC 7540.
+
+    :param hdr_validation_flags: An instance of HeaderValidationFlags.
     """
     # This validation logic is built on a sequence of generators that are
     # iterated over to provide the final header list. This reduces some of the
@@ -160,14 +171,26 @@ def validate_headers(headers):
     # For example, we avoid tuple upacking in loops because it represents a
     # fixed cost that we don't want to spend, instead indexing into the header
     # tuples.
-    headers = _reject_uppercase_header_fields(headers)
-    headers = _reject_te(headers)
-    headers = _reject_connection_header(headers)
-    headers = _reject_pseudo_header_fields(headers)
+    headers = _reject_uppercase_header_fields(
+        headers, hdr_validation_flags
+    )
+    headers = _reject_te(
+        headers, hdr_validation_flags
+    )
+    headers = _reject_connection_header(
+        headers, hdr_validation_flags
+    )
+    headers = _reject_pseudo_header_fields(
+        headers, hdr_validation_flags
+    )
+    headers = _check_host_authority_header(
+        headers, hdr_validation_flags
+    )
+
     return list(headers)
 
 
-def _reject_uppercase_header_fields(headers):
+def _reject_uppercase_header_fields(headers, hdr_validation_flags):
     """
     Raises a ProtocolError if any uppercase character is found in a header
     block.
@@ -179,7 +202,7 @@ def _reject_uppercase_header_fields(headers):
         yield header
 
 
-def _reject_te(headers):
+def _reject_te(headers, hdr_validation_flags):
     """
     Raises a ProtocolError if the TE header is present in a header block and
     its value is anything other than "trailers".
@@ -195,7 +218,7 @@ def _reject_te(headers):
         yield header
 
 
-def _reject_connection_header(headers):
+def _reject_connection_header(headers, hdr_validation_flags):
     """
     Raises a ProtocolError if the Connection header is present in a header
     block.
@@ -209,7 +232,7 @@ def _reject_connection_header(headers):
         yield header
 
 
-def _reject_pseudo_header_fields(headers):
+def _reject_pseudo_header_fields(headers, hdr_validation_flags):
     """
     Raises a ProtocolError if duplicate pseudo-header fields are found in a
     header block or if a pseudo-header field arrives in a block after an
@@ -242,3 +265,55 @@ def _reject_pseudo_header_fields(headers):
             seen_regular_header = True
 
         yield header
+
+
+def _check_host_authority_header(headers, hdr_validation_flags):
+    """
+    Raises a ProtocolError if a header block arrives that does not contain
+    :authority or a Host header, or if a header block contains both fields,
+    but their values do not match.
+    """
+    # We only expect to see :authority and Host headers on request header
+    # blocks that aren't trailers, so skip this validation if we're on the
+    # server side or looking at trailer blocks.
+    if hdr_validation_flags.is_client or hdr_validation_flags.is_trailer:
+        for header in headers:
+            yield header
+        return
+
+    # We use None as a sentinel value.  Iterate over the list of headers,
+    # and record the value of these headers (if present).  We don't need
+    # to worry about receiving duplicate :authority headers, as this is
+    # enforced by the _reject_pseudo_header_fields() pipeline.
+    #
+    # TODO: We should also guard against receiving duplicate Host headers.
+    authority_header_val = None
+    host_header_val = None
+
+    for header in headers:
+        if header[0] == b':authority':
+            authority_header_val = header[1]
+        elif header[0] == b'host':
+            host_header_val = header[1]
+
+        yield header
+
+    # If we have not-None values for these variables, then we know we saw
+    # the corresponding header.
+    authority_present = (authority_header_val is not None)
+    host_present = (host_header_val is not None)
+
+    # It is an error for a request header block to contain neither
+    # an :authority header nor a Host header.
+    if not authority_present and not host_present:
+        raise ProtocolError(
+            "Did not receive an :authority or Host header."
+        )
+
+    # If we receive both headers, they should definitely match.
+    if authority_present and host_present:
+        if authority_header_val != host_header_val:
+            raise ProtocolError(
+                "Received mismatched :authority and Host headers: %r / %r" %
+                (authority_header_val, host_header_val)
+            )
--- a/test/test_invalid_headers.py
+++ b/test/test_invalid_headers.py
@@ -41,6 +41,9 @@ class TestInvalidFrameSequences(object):
         base_request_headers + [('transfer-encoding', 'gzip')],
         base_request_headers + [('upgrade', 'super-protocol/1.1')],
         base_request_headers + [('te', 'chunked')],
+        base_request_headers + [('host', 'notexample.com')],
+        [header for header in base_request_headers
+         if header[0] != ':authority'],
     ]
 
     @pytest.mark.parametrize('headers', invalid_header_blocks)
@@ -59,7 +62,7 @@ class TestInvalidFrameSequences(object):
             c.receive_data(data)
 
         expected_frame = frame_factory.build_goaway_frame(
-            last_stream_id=0, error_code=h2.errors.PROTOCOL_ERROR
+            last_stream_id=1, error_code=h2.errors.PROTOCOL_ERROR
         )
         assert c.data_to_send() == expected_frame.serialize()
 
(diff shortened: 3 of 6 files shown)


Lukasa:
Thanks for getting started on this @alexwlchan!

The first thing to note is that :authority and Host are only defined on *request* header blocks: we don't need either on *response* blocks (and in fact we'd be quite confused if we received one on responses). That will hopefully reduce the churn in the tests: in fact, I think all the current tests should continue to pass in the face of that change.

This may mean that we need a change to the API of the pipeline to pass a boolean (or possibly an enum, actually) indicating whether the header block is for a request or response. 

As to your question about sending blocks that don't match these rules, I think probably the best thing to do is to define a new exception (a subclass of `ProtocolError`) for this kind of thing (`InvalidHeaderBlock`, maybe?) which we throw both when the remote peer screws this up and when a user screws this up. That provides somewhat more clarity about what's going on.

And yeah, I think we should document this. We can do part of the documentation in the docstring for the new exception, but we can also do it in [this section of the prose docs](http://python-hyper.org/projects/h2/en/stable/advanced-usage.html#headers).

alexwlchan:
Adding an enum to the pipeline API seems sensible, but I’m not sure how we’d use it. Tests aside, the one place `validate_headers` is called is in `H2Connection._receive_headers_frame`. Depending on whether hyper is being used as a server or client, presumably this method could be getting requests or responses?

There’s the `state_machine` attribute on the `H2Connection` object, so perhaps we could use that to work out what we expect to be receiving:

* If we're in the `ConnectionState.SERVER_OPEN` state, then we’re expecting to receive requests
* If we’re in the `ConnectionState.CLIENT_OPEN` state, then we’re expecting to receive responses
* If we’re in the `ConnectionState.IDLE` or `ConnectionState.CLOSED` states, then we don’t expect anything – but I assume that gets policed before we get to this method?

Does that sound sensible?

alexwlchan:
@Lukasa Could I get a quick yay/nay on whether my idea for working out if we’ve got a request/response sounds plausible?

Lukasa:
We can do that, but we can also do better. The connection knows if it's a client or not: see `H2Connection.client_side`. That can be used to do the needful validation. =)

alexwlchan:
OK, the `validate_headers()` now takes a `client_side` parameter, same as the `H2Connection` class. Much less churn required in the tests, yay.

Two tests are still failing:

* `TestBasicServer.test_can_receive_trailers` – the trailer header block only contains a `content-length: 0` header. I’m not totally clued-up on the subtleties of HTTP/2 trailers: does it count as a request in this context? Should we be including :authority or Host on trailer header blocks?
* `TestBasicServer.test_close_connection_with_last_stream_id` – this test sends an empty header block into the server, so it doesn’t find the :authority or Host blocks. This header block is being sent right at the end of the connection – should we be sending these headers here, or does being at the end of a connection make it special?

Lukasa:
1. No, neither `:authority` nor `Host` should appear in trailers. This was a subtlety I wondered about when I was writing my last comment but decided not to mention. This means we probably need to move *validate headers* to `H2Stream.recieve_headers`, and then we need to consider passing a more complex state object to the validators. Probably we need to construct a state object out of a namedtuple with a number of fields on it: `client` and `block_is_trailers` are probably two that we want to accurately fill out before validating. For that, you'll want to use the `client` flag on the `H2StreamStateMachine`.

   Alternatively, if you want to pursue a different approach to this (because this does start seeming to be complex), you could provide a new validation function that checks only `RequestReceived` events and validates that their headers are acceptable. That would probably be an appropriate change as well.

2. No, this header block is supposed to fail validation. Probably all that's required is to change the `receive_data` call to be inside a `with pytest.raises()`.

alexwlchan:
Got slightly caught up on other things, so couldn't look at this for a few days. Back on it now:

1. I've added a call to `validate_headers()` in `H2Stream.receive_headers` which allows us to decide whether we're looking at a trailer block. I think this is the first approach you suggested. For now I'm passing the state as extra keyword arguments, but it can be reduced to a namedtuple quite easily. Does this look appropriate?

2. Looking at the description of that test, I'm not convinced that's correct:

    > Closing the connection with last_stream_id set emits a GOAWAY frame
        with that value.

    If we throw a ProtocolError when we receive the malformed request headers, will we still emit the GOAWAY frame correctly? (Should we?)

Lukasa:
@alexwlchan You'll need to actually push those changes up for me to review them. =)

alexwlchan:
@Lukasa \*facepalm*

test/test_invalid_headers.py
```diff
@@ -92,18 +95,24 @@ class TestFilter(object):
     HTTP/2 and so may never hit the function, but it's worth validating that it
     behaves as expected anyway.
     """
-    @given(HEADERS_STRATEGY)
-    def test_range_of_acceptable_outputs(self, headers):
+    @given(HEADERS_STRATEGY, booleans())
+    def test_range_of_acceptable_outputs(self, headers, client_side):
```
Lukasa: This test is missing a parametrize.

h2/connection.py
```diff
@@ -1387,7 +1387,9 @@ def _receive_headers_frame(self, frame):
             # compatibility, catch all of them.
             raise ProtocolError("Error decoding header block: %s" % e)
 
-        headers = validate_headers(headers)
+        headers = validate_headers(headers,
+                                   client_side=self.client_side,
+                                   is_trailer=False)
```
Lukasa: So I appreciate the plan here, but I definitely think we want to swap this to a namedtuple before we continue. We'll likely have to add more filter criteria here, and so avoiding the proliferation of kwargs would be ideal.

Lukasa: While we're here, this call is probably unnecessary: it's a duplicate of the call in the stream, so you can probably lose it.

h2/utilities.py
```diff
@@ -164,6 +169,12 @@ def validate_headers(headers):
     headers = _reject_te(headers)
     headers = _reject_connection_header(headers)
     headers = _reject_pseudo_header_fields(headers)
+
+    # We only expect to see :authority and Host headers on request header
+    # blocks, so only perform this validation if we're on the server side.
+    if (not client_side) and (not is_trailer):
+        headers = _check_host_authority_header(headers)
```
Lukasa: So ideally we'd have a single state namedtuple here, and then each filter would accept both the headers and the state tuple as an argument. That lets us define essentially a single logical "filter function". The pre-existing filters just ignore the state, but then `_check_host_authority_header` can use it to make a decision (and in this case, decide to be a no-op).

alexwlchan:
@Lukasa Pushed a new set of changes based on your comments. In particular, we now use a namedtuple `ConnectionState` as part of the `validate_headers()` API.

This seems to have broken all the tests in `test_invalid_headers.py`. I’m not sure why my changes have caused this particular failure. Any thoughts?

Lukasa:
So the reason all the tests failed is because we moved the header validation until *after* we processed the header block. That means the last stream ID is now logically 1, not 0: we've affected our internal state with data from stream 1. Those tests just need fixing up to use `last_stream_id=1`.

I'll do more specific review on the diff directly.

h2/utilities.py
```diff
@@ -147,7 +148,11 @@ def authority_from_headers(headers):
     return None
 
 
-def validate_headers(headers):
+ConnectionState = collections.namedtuple('ConnectionState',
+                                         ['is_client', 'is_trailer'])
```
Lukasa: I don't love the name `ConnectionState` here. Maybe `HeaderValidationFlags`?

h2/utilities.py
```diff
@@ -242,3 +249,41 @@ def _reject_pseudo_header_fields(headers):
...
+        raise ProtocolError(
+            "Did not receive an :authority or Host header."
+        )
+
+    if (authority_header_val is not None) and (host_header_val is not None):
+        if authority_header_val != host_header_val:
+            raise ProtocolError(
+                "Received mismatched :authority and Host headers: %r / %r" %
+                (authority_header_val, host_header_val)
+            )
```
Lukasa: This doesn't have to be dealt with in this PR, but we'll also want to reject this kind of header block when our users send it. That means we'll probably want to factor the core of this logic (checking the header block) out to a function we can call in a separate situation to this one. You can either do that now, or we can do it in a follow-up PR: your call.

h2/utilities.py
```diff
@@ -242,3 +249,41 @@ def _reject_pseudo_header_fields(headers):
...
+            host_header_val = header[1]
+
+        yield header
+
+    if (authority_header_val is None) and (host_header_val is None):
+        raise ProtocolError(
+            "Did not receive an :authority or Host header."
+        )
+
+    if (authority_header_val is not None) and (host_header_val is not None):
```
Lukasa: We probably don't need this outer check. `None` won't compare equal to a string under any circumstances.

Lukasa:
Looking really good! I have only a relatively few comments here. =)

h2/utilities.py
alexwlchan: Two hard problems in computer science, etc. :smile: 

(Also, it turns out there already ~is something called `ConnectionState` in hyper. What could possibly go wrong from a second object with the same name?)

Fixed.

h2/utilities.py
alexwlchan: I'm confused. This is checking that if we receive both the :authority and the Host header, they match. We can only check that they match if we receive both headers (i.e., both `authority_header_val` and `host_header_val` are not-`None`).

If we remove this check and receive a header block with only one of :authority and Host, this will always fail -- precisely because a string will never match `None`.

Lukasa: Oh, that's quite right. In that case, rather than have the nested `if` blocks (which have already confused me once!), let's have sequential `if` blocks. That lets us more easily have a comment that explains what the `None` checks are.

alexwlchan: K, I've tidied up the `if` blocks and added some comments to this function.

alexwlchan:
There's a weird interaction between pytest fixtures and Hypothesis that I haven't got working yet, but I think that's the last failing test.

I'll defer enforcing these rules on sent header blocks for a separate PR. This is starting to get quite messy, and it would be good to have a clean slate for that work.

h2/utilities.py
```diff
@@ -242,3 +249,56 @@ def _reject_pseudo_header_fields(headers):
...
+        for header in headers:
+            yield header
+        return
+
+    # We use None as a sentinel value.  Iterate over the list of headers,
+    # and record the value of these headers (if present).  We don't need
+    # to worry about receiving duplicate :authority headers, as this is
+    # enforced by the _reject_pseudo_header_fields() pipeline.
+    #
+    # @@AWLC: Do we have mechanisms to enforce non-duplicate Host headers?
```
alexwlchan: I'm assuming it is the case that getting duplicate Host headers would be a bug, but I can't find anything that enforces that.

Lukasa: Duplicate Host headers should be rejected too, but that's a pretty strict reading of the specs. Let's hold off on it for now: maybe put it as a `#TODO`?

h2/utilities.py
```diff
@@ -147,7 +148,11 @@ def authority_from_headers(headers):
     return None
 
 
-def validate_headers(headers):
+HeaderValidationFlags = collections.namedtuple('HeaderValidationFlags',
+                                               ['is_client', 'is_trailer'])
```
Lukasa: Can we get a comment here to explain what this structure is for?

h2/utilities.py
```diff
@@ -160,14 +165,16 @@ def validate_headers(headers):
...
     # tuples.
-    headers = _reject_uppercase_header_fields(headers)
-    headers = _reject_te(headers)
-    headers = _reject_connection_header(headers)
-    headers = _reject_pseudo_header_fields(headers)
+    headers = _reject_uppercase_header_fields(headers, hdr_validation_flags)
+    headers = _reject_te(headers, hdr_validation_flags)
+    headers = _reject_connection_header(headers, hdr_validation_flags)
+    headers = _reject_pseudo_header_fields(headers, hdr_validation_flags)
+    headers = _check_host_authority_header(headers, hdr_validation_flags)
```
Lukasa: So the indentation policy in h2 is this:

```python
headers = _check_host_authority_header(
    headers, hdr_validation_flags
)
```

Mind updating?

h2/utilities.py
```diff
@@ -242,3 +249,56 @@ def _reject_pseudo_header_fields(headers):
...
+    :authority or a Host header, or if a header block contains both fields,
+    but their values do not match.
+    """
+    # We only expect to see :authority and Host headers on request header
+    # blocks that aren't trailers, so skip this validation if we're on the
+    # server side or looking at trailer blocks.
+    if hdr_validation_flags.is_client or hdr_validation_flags.is_trailer:
+        for header in headers:
+            yield header
+        return
```
Lukasa: We can just return `headers` here, I think,

h2/utilities.py
```diff
@@ -242,3 +249,56 @@ def _reject_pseudo_header_fields(headers):
...
+            host_header_val = header[1]
+
+        yield header
+
+    # If both of these variables are None, then we didn't find
+    # an :authority or a Host header.
+    if (
+        (authority_header_val is None) and
+        (host_header_val is None)
+    ):
```
Lukasa: Rather than create this rather ghastly `if` statement, why not define the conditional on its own logical line? That is:

```python
both_headers_none = (
    authority_header_val is None and
    host_header_val is None
)

if both_headers_none:
    # Do whatever
```

h2/utilities.py
```diff
@@ -242,3 +249,56 @@ def _reject_pseudo_header_fields(headers):
...
+        raise ProtocolError(
+            "Did not receive an :authority or Host header."
+        )
+
+    # If we receive both headers, they should definitely match.
+    if (
+        (authority_header_val is not None) and
+        (host_header_val is not None) and:
+        (authority_header_val != host_header_val)
+    ):
```
Lukasa: Sorry, I wasn't clear enough. I meant to have a separate `if` block. But now that I see both these conditionals together, I think there's a better way to structure this:

```python
authority_present = (authority_header_val is not None)
host_present = (host_header_val is not None)

if not authority_present and not host_present:
    raise ProtocolError(
        "Did not receive an :authority or Host header."
    )

if authority_present and host_present:
    if authority_header_val != host_header_val:
       raise ProtocolError("Mismatched")
```

This seems to me to most clearly communicate what's actually being tested. Do you agree with that assessment?

Lukasa: Note that we can combine the two if conditionals at the bottom into one. I'm not sure that it improves clarity much, but that's up to you.

h2/utilities.py
```diff
@@ -147,9 +148,17 @@ def authority_from_headers(headers):
     return None
 
 
-def validate_headers(headers):
+# Flags used by the validate_headers pipeline to determine which checks
+# should be applied to a given set of headers.
+HeaderValidationFlags = collections.namedtuple('HeaderValidationFlags',
+                                               ['is_client', 'is_trailer'])
```
Lukasa: Lets do some indentation cleanups here. Rather than constrain ourselves to only having a tiny number of characters to work with, let's reflow to:

```python
HeaderValidationFlags = collections.namedtuple(
    'HeaderValidationFlags',
    ['is_client', 'is_trailer']
)
```

test/test_invalid_headers.py
```diff
@@ -92,18 +95,29 @@ class TestFilter(object):
...
 
-    def test_invalid_pseudo_headers(self):
+    @pytest.mark.parametrize('is_client', [True, False])
+    def test_invalid_pseudo_headers(self, is_client):
         headers = [(b':custom', b'value')]
         with pytest.raises(h2.exceptions.ProtocolError):
-            h2.utilities.validate_headers(headers)
+            conn_state = h2.utilities.HeaderValidationFlags(
+                is_client=is_client,
+                is_trailer=False
```
Lukasa: We should probably parametrize this as well.

test/test_invalid_headers.py
```diff
@@ -92,18 +95,29 @@ class TestFilter(object):
...
+    def test_range_of_acceptable_outputs(self, headers, is_client):
         """
         validate_headers either returns the data unchanged or throws a
         ProtocolError.
         """
         try:
-            assert headers == h2.utilities.validate_headers(headers)
+            conn_state = h2.utilities.HeaderValidationFlags(
+                is_client=is_client,
+                is_trailer=False
```
Lukasa: Same here, we should probably parametrize that too.

h2/utilities.py
```diff
@@ -242,3 +265,55 @@ def _reject_pseudo_header_fields(headers):
...
+    Raises a ProtocolError if a header block arrives that does not contain
+    :authority or a Host header, or if a header block contains both fields,
+    but their values do not match.
+    """
+    # We only expect to see :authority and Host headers on request header
+    # blocks that aren't trailers, so skip this validation if we're on the
+    # server side or looking at trailer blocks.
+    if hdr_validation_flags.is_client or hdr_validation_flags.is_trailer:
+        for header in headers:
+            yield
```
Lukasa: This naked yield isn't right. =(

alexwlchan: ...oops

alexwlchan:
At last, a clean test run!

I need to write some documentation and squash these commits, but I’ll save that for the morning. I probably shouldn't rebase when I’m tired. 😛

Lukasa:
@alexwlchan Sounds good to me, thank you so much! :sparkles:

alexwlchan:
Rebased against master, squashed, and change log entry added. I think this is ready to go.

test/test_invalid_headers.py
```diff
@@ -92,18 +95,56 @@ class TestFilter(object):
...
         """
         validate_headers either returns the data unchanged or throws a
         ProtocolError.
         """
         try:
-            assert headers == h2.utilities.validate_headers(headers)
+            conn_state = h2.utilities.HeaderValidationFlags(
+                is_client=is_client,
+                is_trailer=is_trailer,
+            )
```
Lukasa: We're at the height of pedanticness now, but seeing as you've been so willing to tinker...can you hoist the construction of the `HeaderValidationFlags` both here and in `test_invalid_pseudo_headers` out of their containing exception scopes? It makes it a bit less clear what statement we're trying to protect. =) Small `try` blocks are good.

alexwlchan: Super-pedantry, but fixed. :wink:

Lukasa:
Beautiful work @alexwlchan, thank you so much for putting up with that extensive code review! :sparkles: :cake: :sparkles:

PR #258, opened 2016-07-14
initialise content length to zero if HEAD request

aidylewis:


aidylewis:
The PR has failed the checks. How do I run all the unit tests locally?

Lukasa:
Sure! We run the unit tests using [tox](https://pypi.python.org/pypi/tox). You'll need to `pip install` tox, and then the tests can be run by simply running `tox`. If you don't have some Python versions installed tox will complain at you by printing angrily in red, but you can ignore that for now: the CI rig will cover all the Pythons even if you can't.

h2/stream.py
```diff
@@ -668,6 +668,7 @@ def __init__(self, stream_id):
         self.state_machine = H2StreamStateMachine(stream_id)
         self.stream_id = stream_id
         self.max_outbound_frame_size = None
+        self.headers = []
```
Lukasa: There's no need to store all the headers. We can avoid it simply by storing the actual value we need: specifically, the `:method` header.

aidylewis:
Guess, I am failing on not writing a unit test. I was using the twisted/request examples to verify the fix!!!

Lukasa:
So this is beginning to look very good indeed, thank you!

Yeah, as you've noticed, we'll need unit tests for this: probably not just one, but several. Additionally, I think we'll want to write some more code. The code you've written has a few changes that I'd like:

1. Creating a temporary dictionary for a lookup and then throwing it away is a pretty gratuitous wasted of memory, and it's also costly in terms of extra allocations that cannot be made more efficient. It'd be better, I think, to have a linear scan of the header block. This is *particularly* true because a valid HTTP/2 request header block *must* have `:method` in the first four headers in the block, so while the *nominal* runtime of such a loop would be `O(n)` the *actual* runtime of the loop will be `O(1)` in any case where the request would succeed. Of course, this is enough code that we'll need to write a new function for this (to put in `utils.py`, probably), and that leads on neatly to my next problem.
2. For my sins, hyper-h2 allows both bytestring and unicode strings for header names and values. That means we need to support both. The complexity of doing that is relatively simple: we can check `if header_name in (b':method', u':method')` and then we can check the type of the value and optionally encode it to a bytestring if we need to when we store it.
3. Of course, if we're going to do all of that, it may be better to avoid storing the string entirely. We could instead have an `enum`, `HeaderMethods`, that includes `HEAD`, and then store that. That moves the check into somewhere logical and avoids the weirdness of using a string here. This last step is entirely optional: if I were writing the code myself I'd probably do it this way, but if you write it you can do as you see fit.

I'm aware that there's quite a lot of change there, so please let me know if you feel like that's too much work and I can help out by doing some of it, or taking this over from you. You should do only as much work as you feel comfortable doing. Of course, if you'd like to write the code but not the tests, that's fine, and if you think the extra complexity here isn't worth it I understand completely. Thank you so much for the work you've already done! :sparkles:

aidylewis:
I am finding this really enjoyable, so I don't mind TDD'ing.

Lukasa:
I'm glad you're enjoying it! Work like this is really valuable, so it's much appreciated. As always, feel free to ask for clarification on anything, or any other questions you might need answered.

aidylewis:
Before I go down a rabbit hole ...

I have kept the request method as a tuple for consistency  ```(':method', 'HEAD')```. 

I have created a ```test_extract_header_method``` which is implemented in ```utilities.py```. The only place I know we can extract the headers is in ```send_headers```, which is used in ```_initialize_content_length```. The code coverage utility doesn't like the non-tested if statement, but I am not sure how to test this.

h2/stream.py
```diff
@@ -735,6 +737,9 @@ def send_headers(self, headers, encoder, end_stream=False):
         except AttributeError:
             headers = headers
 
+        # store request method for _initialize_content_length
+        self.request_method = extract_method_header(headers)
+
```
Lukasa: So it may be worth moving this towards the bottom of this method. That allows us to check `self.state_machine.client` and `input_` to work out whether or not we're likely to find a method, and so we can avoid doing unnecessary searches of the header block.

h2/utilities.py
```diff
@@ -67,6 +67,20 @@ def secure_headers(headers):
...
+    request_header = (
+            'GET', 'HEAD',
+            'POST', 'PUT'
+            'DELETE', 'TRACE',
+            'CONNECT',
+            )
+
+    for k, v in headers:
+        if k == ':method' and v in request_header:
+            return (k, v)
```
Lukasa: So we can change this function up a bit.

Firstly, let's not be too fussy about what `:method` we get. If we get a `:method`, we use it, even if it's not something we recognise.

Secondly, let's not return the tuple: just the method name.

Thirdly, this doesn't handle the possibility of receiving unicode and bytes headers. You'll want to look at some of the other functions in this module to see what they do: see the use of `_SECURE_HEADERS`, for example.

Lukasa:
Let's fix up the functions first, and then worry about test coverage. =)

aidylewis:
I believe I made your suggested changes.

Lukasa:
Alright @aidylewis, this is getting really good! We're into the realm of small changes now. =)

I'm going to leave some notes inline for things I'd like added or changed. As always, at any point feel free to say you'd like help or you'd like me to do something specific.

h2/stream.py
```diff
@@ -723,6 +724,7 @@ def send_headers(self, headers, encoder, end_stream=False):
         Returns a list of HEADERS/CONTINUATION frames to emit as either headers
         or trailers.
         """
+
```
Lukasa: The general style in this codebase is to not have a blank line following a docstring. =)

h2/utilities.py
```diff
@@ -67,6 +67,13 @@ def secure_headers(headers):
             yield header
 
 
+def extract_method_header(headers):
+
```
Lukasa: This method needs a docstring. =)

h2/utilities.py
```diff
@@ -67,6 +67,13 @@ def secure_headers(headers):
...
+    for k, v in headers:
+        if k in (b':method', u':method'):
+            return v
```
Lukasa: Let's save ourselves time down the road by always ensuring we have a bytes header value.

```python
if not isinstance(v, bytes):
    return v.encode('utf-8')
else:
    return v
```

h2/stream.py
```diff
@@ -1050,6 +1055,11 @@ def _initialize_content_length(self, headers):
         _expected_content_length field from it. It's not an error for no
         Content-Length header to be present.
         """
+
+        if self.request_method in (b'HEAD', u'HEAD'):
```
Lukasa: If we make the change suggested in `extract_method_header`, this can become a straight bytestring comparison.

test/test_utility_functions.py
```diff
@@ -151,3 +152,17 @@ def test_does_not_increment_without_stream_send(self):
...
+    example_request_headers = [
+        (':authority', 'example.com'),
+        (':path', '/'),
+        (':scheme', 'https'),
+        (':method', 'GET'),
+    ]
+
+    def test_extract_header_method(self):
+        assert extract_method_header(self.example_request_headers) \
+            == u'GET'
```
Lukasa: In addition to this, I think I'd like two tests in the main tests to confirm that we use this information correctly. Firstly, one test showing a HEAD request and response with a non-zero content length and no body that works just fine. Then a test showing a HEAD request and a response with a non-zero content length and a body that is correctly rejected.

Do you feel up to writing those two tests? You can take inspiration from some of the other tests in`test_basic_logic`.

h2/utilities.py
aidylewis:
```python
return v.encode('utf-8')
```
Shouldn't we be encoding it to bytes?

Lukasa: That's exactly what that method call does.

test/test_utility_functions.py
aidylewis: I've made the suggested changes but not included the additional core tests. I'll do that now as I have work to do tomorrow.

aidylewis:
Is https://github.com/python-hyper/hyper-h2/pull/258/commits/c4037514fe6dd6da37fb74bdfe823af42d3604a2 OK for the test: 'HEAD request and response with a non-zero content length and no body'? 

Not sure about the test: 'HEAD request and a response with a non-zero content length and a body that is correctly rejected'. From my understanding a HEAD request doesn't have a body.

Lukasa:
The response to a HEAD request should not have a body, that's right. That's why we want a test for when we receive a response like that that *does* have a body: we need to confirm that we behave correctly in the error situation as well as the success.

test/test_head_request.py
```diff
@@ -0,0 +1,42 @@
...
+        ]
+
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, example_request_headers)
+
+        f = frame_factory.build_headers_frame(
+            example_response_headers
+        )
+        events = c.receive_data(f.serialize())
```
Lukasa: I'd like this test to make sure it sets `END_STREAM` on the headers frame too, to ensure that we are testing the full state machine. That'll increase our event count to two.

aidylewis:
Will I need to build a data frame ```frame_factory.build_data_frame(data=b'\x01')``` to force a response body?

aidylewis:
I've squashed the last 4 commits in https://github.com/python-hyper/hyper-h2/pull/258/commits/3d8ce1fb6c6056b93b48b4be4d5ceaadf56a7bd5

test/test_head_request.py
```diff
@@ -0,0 +1,56 @@
...
+
+        def test_reject_non_zero_content_and_body(self, frame_factory):
+            c = h2.connection.H2Connection()
+            c.initiate_connection()
+            c.send_headers(1, self.example_request_headers)
+
+            headers = frame_factory.build_headers_frame(
+                self.example_response_headers
+            )
+            data = frame_factory.build_data_frame(data=b'\x01')
```
Lukasa: Can we add the `'END_STREAM'` flag here too?

Lukasa:
Awesome, this is looking very close to done! :sparkles:

test/test_head_request.py
aidylewis: Why do we want to add an ```END_STREAM``` flag now we are also sending data?

aidylewis:
Squashed all commits. All tests passing locally.

test/test_head_request.py
Lukasa: Because the goal here is for this test response to be wrong in only one specific way. In all other ways it should be a perfectly valid response. That means, ideally, we'd like to correctly terminate the stream.

It also protects us if we want to evolve the function in future to, for example, only raise when the response is complete. =)

Lukasa:
Hrm, Travis seems to have lost this PR. Let me try to kick it.

Lukasa:
Well that's extremely perplexing. I'll try to reach out to some Travis-CI folks and see if I can work out what's happening.

test/test_head_request.py
aidylewis: If I add the ```END_STREAM``` flag, the exception is not raised.

From pytest: 
```python
    def test_reject_non_zero_content_and_body(self, frame_factory):
        c.initiate_connection()

        headers = frame_factory.build_headers_frame(
            self.example_response_headers,
            flags=['END_STREAM']
        )
        data = frame_factory.build_data_frame(data=b'\x01')

        c.receive_data(headers.serialize())
        with pytest.raises(h2.exceptions.InvalidBodyLengthError):
>           c.receive_data(data.serialize())
E           Failed: DID NOT RAISE
```

Lukasa: Sorry, I want the `END_STREAM` flag on the data frame. =)

aidylewis:
I can fork again and apply a patch?

Lukasa:
@aidylewis It's up to you really, but it'd be enough just to cut a new branch from here and open a pull request from that. =)

aidylewis:
I forked to the BBC repo, but the patch is failing. I'll just ``git show head`` on my old fork and manually include the changes,

aidylewis:
I have created a new PR from the BBC repo. I'll close this and delete my personal repo fork.

Lukasa:
For the sake of future me trying to keep this straight, this has been superseded by #264.

PR #259, opened 2016-07-15
Fix example in the getting started

Mec-iS:
Headers has to be defined with an ordered data structure or it can
happen the `ProtocolError` to be raised by the `_reject_pseudo_header_fields`
procedure (`seen_regular_header` condition). `dict` does not guarantee the
ordering of keys. This can be corrected by implementing `collections.OrderedDict`
or `list`. To not brake the API, I corrected the dictionary into an
ordered dictionary.

### DIFF
--- a/docs/source/basic-usage.rst
+++ b/docs/source/basic-usage.rst
@@ -437,7 +437,10 @@ signaled the request. Let's define it.
         stream_id = event.stream_id
         conn.send_headers(
             stream_id=stream_id,
-            headers={':status': '200', 'server': 'basic-h2-server/1.0'},
+            headers=OrderedDict([
+                (':status', '200'), 
+                ('server', 'basic-h2-server/1.0')
+            ]),
         )
         conn.send_data(
             stream_id=stream_id,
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this change! In fact, the using of dictionaries for this has been deprecated (see [this page](http://python-hyper.org/projects/h2/en/stable/advanced-usage.html#headers)) in the docs, and I clearly just missed that part. Would you like to change this to use lists, rather than OrderedDicts?

Mec-iS:
Cool, didn't reach that page in the doc yet.

For sure dictionaries provide a better interface both for readability and fluency (but are quite more inefficient for small set of keys), tuples are harder to read (but more efficient). There is also the `namedtuple` alternative, but it's quite hard to read as well. 
Maybe it can be better to implement a 'dict-to-list of tuples' casting in the `_reject_pseudo_header_fields` before checking the condition. This way the higher level interface (the `H2Connection.send_headers` method) remains clean using dictionaries and the 'tupling' takes place in the private method to order the pseudo-keys. Sounds good? Any alternative?

Thanks for the work done until now, I will keep exploring and trying to implement more complex examples. Hope to be of more help.

Lukasa:
For the moment, we've just *entirely* deprecated the use of dicts internally. The list of tuples approach is a bit more annoying to handle but ultimately cleaner. We get away with that because this is a low-level library.

Mec-iS:
ok, I will do it with list of tuples then.

PR #260, opened 2016-07-17
Fix headers' declaration in 'base-usage.rst' with lists

Mec-iS:
Headers has to be declared as lists instead of dictionaries, because
of the decision of deprecating entirely dictionaries usage.


Lukasa:
Thanks so much for this! :sparkles: :cake: :sparkles:

aidylewis:
Have you deprecated dictionaries for lists of tuples because of mutability?

Lukasa:
Nope, it's because of ordering. Specifically, HTTP/2 enforces certain ordering constraints on headers that mean that using a dictionary for them as input is unwise.

PR #261, opened 2016-07-20
Reject sent header blocks with bad :authority/Host headers

alexwlchan:
Quick something I cooked up in my lunch break.

This continues the work started in #252. Intention is that if the user tries to send :authority or Host headers that don't conform, we should raise an `InvalidHeaderBlock` exception.

For now, I've just refactored the existing send_headers pipeline into `utilities.py`, ready to add the new logic. I'll look at the rest of it this evening/tomorrow.

### DIFF
--- a/h2/config.py
+++ b/h2/config.py
@@ -32,19 +32,19 @@ class H2Configuration(object):
         ``False`` or the empty string.
     :type header_encoding: ``str``, ``False``, or ``None``
 
-    :param validate_outbound_headers: Controls whether the headers emitted
+    :param validate_sent_headers: Controls whether the headers emitted
         by this object are validated against the rules in RFC 7540.
         Disabling this setting will cause outbound header validation to
         be skipped, and allow the object to emit headers that may be illegal
         according to RFC 7540. Defaults to ``True``.
-    :type validate_outbound_headers: ``bool``
+    :type validate_sent_headers: ``bool``
 
-    :param normalize_outbound_headers: Controls whether the headers emitted
+    :param normalize_sent_headers: Controls whether the headers emitted
         by this object are normalized before sending.  Disabling this setting
         will cause outbound header normalization to be skipped, and allow
         the object to emit headers that may be illegal according to
         RFC 7540. Defaults to ``True``.
-    :type normalize_outbound_headers: ``bool``
+    :type normalize_sent_headers: ``bool``
 
     :param validate_inbound_headers: Controls whether the headers received
         by this object are validated against the rules in RFC 7540.
@@ -56,13 +56,13 @@ class H2Configuration(object):
     def __init__(self,
                  client_side=True,
                  header_encoding='utf-8',
-                 validate_outbound_headers=True,
-                 normalize_outbound_headers=True,
+                 validate_sent_headers=True,
+                 normalize_sent_headers=True,
                  validate_inbound_headers=True):
         self._client_side = client_side
         self._header_encoding = header_encoding
-        self._validate_outbound_headers = validate_outbound_headers
-        self._normalize_outbound_headers = normalize_outbound_headers
+        self._validate_sent_headers = validate_sent_headers
+        self._normalize_sent_headers = normalize_sent_headers
         self._validate_inbound_headers = validate_inbound_headers
 
     @property
@@ -108,42 +108,42 @@ class H2Configuration(object):
         self._header_encoding = value
 
     @property
-    def validate_outbound_headers(self):
+    def validate_sent_headers(self):
         """
         Whether the headers emitted by this object are validated against
         the rules in RFC 7540. Disabling this setting will cause outbound
         header validation to be skipped, and allow the object to emit headers
         that may be illegal according to RFC 7540. Defaults to ``True``.
         """
-        return self._validate_outbound_headers
+        return self._validate_sent_headers
 
-    @validate_outbound_headers.setter
-    def validate_outbound_headers(self, value):
+    @validate_sent_headers.setter
+    def validate_sent_headers(self, value):
         """
         Enforces validation of outbound headers.
         """
         if not isinstance(value, bool):
-            raise ValueError("validate_outbound_headers must be a bool")
-        self._validate_outbound_headers = value
+            raise ValueError("validate_sent_headers must be a bool")
+        self._validate_sent_headers = value
 
     @property
-    def normalize_outbound_headers(self):
+    def normalize_sent_headers(self):
         """
         Whether the headers emitted by this object are normalized before
         sending.  Disabling this setting will cause outbound header
         normalization to be skipped, and allow the object to emit headers
         that may be illegal according to RFC 7540. Defaults to ``True``.
         """
-        return self._normalize_outbound_headers
+        return self._normalize_sent_headers
 
-    @normalize_outbound_headers.setter
-    def normalize_outbound_headers(self, value):
+    @normalize_sent_headers.setter
+    def normalize_sent_headers(self, value):
         """
         Enforces normalization of outbound headers.
         """
         if not isinstance(value, bool):
-            raise ValueError("normalize_outbound_headers must be a bool")
-        self._normalize_outbound_headers = value
+            raise ValueError("normalize_sent_headers must be a bool")
+        self._normalize_sent_headers = value
 
     @property
     def validate_inbound_headers(self):
--- a/h2/config.py
+++ b/h2/config.py
@@ -31,10 +31,39 @@ class H2Configuration(object):
         to force them to be returned as bytestrings), this can be set to
         ``False`` or the empty string.
     :type header_encoding: ``str``, ``False``, or ``None``
+
+    :param validate_outbound_headers: Controls whether the headers emitted
+        by this object are validated against the rules in RFC 7540.
+        Disabling this setting will cause outbound header validation to
+        be skipped, and allow the object to emit headers that may be illegal
+        according to RFC 7540. Defaults to ``True``.
+    :type validate_outbound_headers: ``bool``
+
+    :param normalize_outbound_headers: Controls whether the headers emitted
+        by this object are normalized before sending.  Disabling this setting
+        will cause outbound header normalization to be skipped, and allow
+        the object to emit headers that may be illegal according to
+        RFC 7540. Defaults to ``True``.
+    :type normalize_outbound_headers: ``bool``
+
+    :param validate_inbound_headers: Controls whether the headers received
+        by this object are validated against the rules in RFC 7540.
+        Disabling this setting will cause inbound header validation to
+        be skipped, and allow the object to receive headers that may be illegal
+        according to RFC 7540. Defaults to ``True``.
+    :type validate_inbound_headers: ``bool``
     """
-    def __init__(self, client_side=True, header_encoding='utf-8'):
+    def __init__(self,
+                 client_side=True,
+                 header_encoding='utf-8',
+                 validate_outbound_headers=True,
+                 normalize_outbound_headers=True,
+                 validate_inbound_headers=True):
         self._client_side = client_side
         self._header_encoding = header_encoding
+        self._validate_outbound_headers = validate_outbound_headers
+        self._normalize_outbound_headers = normalize_outbound_headers
+        self._validate_inbound_headers = validate_inbound_headers
 
     @property
     def client_side(self):
@@ -77,3 +106,61 @@ class H2Configuration(object):
         if value is True:
             raise ValueError("header_encoding cannot be True")
         self._header_encoding = value
+
+    @property
+    def validate_outbound_headers(self):
+        """
+        Whether the headers emitted by this object are validated against
+        the rules in RFC 7540. Disabling this setting will cause outbound
+        header validation to be skipped, and allow the object to emit headers
+        that may be illegal according to RFC 7540. Defaults to ``True``.
+        """
+        return self._validate_outbound_headers
+
+    @validate_outbound_headers.setter
+    def validate_outbound_headers(self, value):
+        """
+        Enforces validation of outbound headers.
+        """
+        if not isinstance(value, bool):
+            raise ValueError("validate_outbound_headers must be a bool")
+        self._validate_outbound_headers = value
+
+    @property
+    def normalize_outbound_headers(self):
+        """
+        Whether the headers emitted by this object are normalized before
+        sending.  Disabling this setting will cause outbound header
+        normalization to be skipped, and allow the object to emit headers
+        that may be illegal according to RFC 7540. Defaults to ``True``.
+        """
+        return self._normalize_outbound_headers
+
+    @normalize_outbound_headers.setter
+    def normalize_outbound_headers(self, value):
+        """
+        Enforces normalization of outbound headers.
+        """
+        if not isinstance(value, bool):
+            raise ValueError("normalize_outbound_headers must be a bool")
+        self._normalize_outbound_headers = value
+
+    @property
+    def validate_inbound_headers(self):
+        """
+        Whether the headers received by this object are validated against
+        the rules in RFC 7540. Disabling this setting will cause inbound
+        header validation to be skipped, and allow the object to receive
+        headers that may be illegal according to RFC 7540.
+        Defaults to ``True``.
+        """
+        return self._validate_inbound_headers
+
+    @validate_inbound_headers.setter
+    def validate_inbound_headers(self, value):
+        """
+        Enforces validation of inbound headers.
+        """
+        if not isinstance(value, bool):
+            raise ValueError("validate_inbound_headers must be a bool")
+        self._validate_inbound_headers = value
--- a/h2/config.py
+++ b/h2/config.py
@@ -32,19 +32,19 @@ class H2Configuration(object):
         ``False`` or the empty string.
     :type header_encoding: ``str``, ``False``, or ``None``
 
-    :param validate_sent_headers: Controls whether the headers emitted
+    :param validate_outbound_headers: Controls whether the headers emitted
         by this object are validated against the rules in RFC 7540.
         Disabling this setting will cause outbound header validation to
         be skipped, and allow the object to emit headers that may be illegal
         according to RFC 7540. Defaults to ``True``.
-    :type validate_sent_headers: ``bool``
+    :type validate_outbound_headers: ``bool``
 
-    :param normalize_sent_headers: Controls whether the headers emitted
+    :param normalize_outbound_headers: Controls whether the headers emitted
         by this object are normalized before sending.  Disabling this setting
         will cause outbound header normalization to be skipped, and allow
         the object to emit headers that may be illegal according to
         RFC 7540. Defaults to ``True``.
-    :type normalize_sent_headers: ``bool``
+    :type normalize_outbound_headers: ``bool``
 
     :param validate_inbound_headers: Controls whether the headers received
         by this object are validated against the rules in RFC 7540.
@@ -56,13 +56,13 @@ class H2Configuration(object):
     def __init__(self,
                  client_side=True,
                  header_encoding='utf-8',
-                 validate_sent_headers=True,
-                 normalize_sent_headers=True,
+                 validate_outbound_headers=True,
+                 normalize_outbound_headers=True,
                  validate_inbound_headers=True):
         self._client_side = client_side
         self._header_encoding = header_encoding
-        self._validate_sent_headers = validate_sent_headers
-        self._normalize_sent_headers = normalize_sent_headers
+        self._validate_outbound_headers = validate_outbound_headers
+        self._normalize_outbound_headers = normalize_outbound_headers
         self._validate_inbound_headers = validate_inbound_headers
 
     @property
@@ -108,42 +108,42 @@ class H2Configuration(object):
         self._header_encoding = value
 
     @property
-    def validate_sent_headers(self):
+    def validate_outbound_headers(self):
         """
         Whether the headers emitted by this object are validated against
         the rules in RFC 7540. Disabling this setting will cause outbound
         header validation to be skipped, and allow the object to emit headers
         that may be illegal according to RFC 7540. Defaults to ``True``.
         """
-        return self._validate_sent_headers
+        return self._validate_outbound_headers
 
-    @validate_sent_headers.setter
-    def validate_sent_headers(self, value):
+    @validate_outbound_headers.setter
+    def validate_outbound_headers(self, value):
         """
         Enforces validation of outbound headers.
         """
         if not isinstance(value, bool):
-            raise ValueError("validate_sent_headers must be a bool")
-        self._validate_sent_headers = value
+            raise ValueError("validate_outbound_headers must be a bool")
+        self._validate_outbound_headers = value
 
     @property
-    def normalize_sent_headers(self):
+    def normalize_outbound_headers(self):
         """
         Whether the headers emitted by this object are normalized before
         sending.  Disabling this setting will cause outbound header
         normalization to be skipped, and allow the object to emit headers
         that may be illegal according to RFC 7540. Defaults to ``True``.
         """
-        return self._normalize_sent_headers
+        return self._normalize_outbound_headers
 
-    @normalize_sent_headers.setter
-    def normalize_sent_headers(self, value):
+    @normalize_outbound_headers.setter
+    def normalize_outbound_headers(self, value):
         """
         Enforces normalization of outbound headers.
         """
         if not isinstance(value, bool):
-            raise ValueError("normalize_sent_headers must be a bool")
-        self._normalize_sent_headers = value
+            raise ValueError("normalize_outbound_headers must be a bool")
+        self._normalize_outbound_headers = value
 
     @property
     def validate_inbound_headers(self):
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -512,6 +512,7 @@ class H2Connection(object):
             self.remote_settings.initial_window_size
         )
         s.inbound_flow_control_window = self.local_settings.initial_window_size
+        s.config = self.config
 
         self.streams[stream_id] = s
 
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -505,14 +505,13 @@ class H2Connection(object):
                 "Invalid stream ID for peer."
             )
 
-        s = H2Stream(stream_id)
+        s = H2Stream(stream_id, config=self.config)
         s.max_inbound_frame_size = self.max_inbound_frame_size
         s.max_outbound_frame_size = self.max_outbound_frame_size
         s.outbound_flow_control_window = (
             self.remote_settings.initial_window_size
         )
         s.inbound_flow_control_window = self.local_settings.initial_window_size
-        s.config = self.config
 
         self.streams[stream_id] = s
 
--- a/h2/events.py
+++ b/h2/events.py
@@ -136,6 +136,29 @@ class TrailersReceived(object):
         )
 
 
+class _HeadersSent(object):
+    """
+    The _HeadersSent event is fired whenever headers are sent.
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    pass
+
+
+class _TrailersSent(_HeadersSent):
+    """
+    The _TrailersSent event is fired whenever trailers are sent on a
+    stream. Trailers are a set of headers sent after the body of the
+    request/response, and are used to provide information that wasn't known
+    ahead of time (e.g. content-length).
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    pass
+
+
 class InformationalResponseReceived(object):
     """
     The InformationalResponseReceived event is fired when an informational
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -14,6 +14,7 @@ from hyperframe.frame import (
     RstStreamFrame, PushPromiseFrame, AltSvcFrame
 )
 
+from .config import H2Configuration
 from .errors import STREAM_CLOSED
 from .events import (
     RequestReceived, ResponseReceived, DataReceived, WindowUpdated,
@@ -690,6 +691,9 @@ class H2Stream(object):
         # The authority we believe this stream belongs to.
         self._authority = None
 
+        # The configuration for this stream.
+        self.config = H2Configuration()
+
     @property
     def open(self):
         """
@@ -926,11 +930,12 @@ class H2Stream(object):
             if not end_stream:
                 raise ProtocolError("Trailers must have END_STREAM set")
 
-        hdr_validation_flags = HeaderValidationFlags(
-            is_client=self.state_machine.client,
-            is_trailer=isinstance(events[0], TrailersReceived)
-        )
-        headers = validate_headers(headers, hdr_validation_flags)
+        if self.config.validate_inbound_headers:
+            hdr_validation_flags = HeaderValidationFlags(
+                is_client=self.state_machine.client,
+                is_trailer=isinstance(events[0], TrailersReceived)
+            )
+            headers = validate_headers(headers, hdr_validation_flags)
 
         if header_encoding:
             headers = list(_decode_headers(headers, header_encoding))
@@ -1053,8 +1058,10 @@ class H2Stream(object):
         """
         # We need to lowercase the header names, and to ensure that secure
         # header fields are kept out of compression contexts.
-        headers = normalize_sent_headers(headers, hdr_validation_flags)
-        headers = validate_sent_headers(headers, hdr_validation_flags)
+        if self.config.normalize_sent_headers:
+            headers = normalize_sent_headers(headers, hdr_validation_flags)
+        if self.config.validate_sent_headers:
+            headers = validate_sent_headers(headers, hdr_validation_flags)
 
         encoded_headers = encoder.encode(headers)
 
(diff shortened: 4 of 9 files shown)


h2/utilities.py
```diff
@@ -41,7 +41,7 @@
 ])
 
 
-def secure_headers(headers):
+def secure_headers(headers, hdr_validation_flags):
```
alexwlchan: This is a public function, although the only place it was used was in `stream.py`.

Is it a problem that the argument signature is changing? Or could we add an underscore and make it non-public, to match the other header pipeline functions in this file?

h2/utilities.py
```diff
@@ -160,6 +160,7 @@ def validate_headers(headers, hdr_validation_flags):
     """
     Validates a header sequence against a set of constraints from RFC 7540.
 
+    :param headers: An iterable of headers.
```
alexwlchan: Some naughty person wrote an incomplete argument list in the docstring, tut tut.

h2/utilities.py
Lukasa: It's not a public function. =)

hyper-h2 is very clear in its documentation: publicness is not signalled by underscores per-se, but by what is documented in the API docs. This is not documented in those docs (in fact, the entire `utilities` module is not). Therefore, we can do whatever we want here.

h2/utilities.py
```diff
@@ -267,9 +243,59 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
...
+    # If we are server-side, then we are receiving the block.
+    if hdr_validation_flags.is_client:
+        mode = 'sending'
+    else:
+        mode = 'receiving'
+
+    # It is an error for a request header block to contain neither
+    # an :authority header nor a Host header.
+    if not authority_present and not host_present:
+        if mode == 'receiving':
```
alexwlchan: I’m not super in love with adding ~another level of `if` statements to this function. Do we need the extra exception type?

Any suggestions for making this cleaner?

h2/utilities.py
```diff
@@ -298,22 +324,94 @@ def _check_host_authority_header(headers, hdr_validation_flags):
...
-        )
 
-    # If we receive both headers, they should definitely match.
-    if authority_present and host_present:
-        if authority_header_val != host_header_val:
-            raise ProtocolError(
-                "Received mismatched :authority and Host headers: %r / %r" %
-                (authority_header_val, host_header_val)
-            )
+def _secure_headers(headers, hdr_validation_flags):
```
alexwlchan: I moved this down just so all the sent-headers pipeline functions are nicely grouped together.

h2/utilities.py
Lukasa: Sorry, this got out of hand a bit.

I suggested in the prior PR that you pull the core of this validation logic out to a separate function, leaving just the checks in here and then a call to that function. That would resolve this problem for you: we can then have one function for receiving and one for sending, each with different checks, and then the sending function can simply wrap the error from the receiving one.

alexwlchan: So I ~think you’re suggesting a function something like:

    def core_validating_logic(authority, host):
        # check at least one is set
        # if both set, check authority == host

which raises a ProtocolError if it fails.

Then the function for sending would catch the `ProtocolError`, an re-raise as an `InvalidHeaderBlockError` with the same exception message. And we tweak the error messages so they don’t refer to receiving/sending, and the same error text can apply to both.

Have I got that right?

Lukasa: That can do it, or we can even pass the exception class in to that function and have it raise from that. But yeah, that's basically right.

alexwlchan:
So I’ve pulled out the header validation, and currently it gets called in `H2Stream._build_headers_frames`, where the original send-header-normalization logic lived. This is causing a whole bunch of tests to fail, which I can sort out in a bit. First let’s get the core function right.

Having now dug into this a bit, a few thoughts:

1. Is this the right place to be putting it? I wonder if `H2Stream.send_headers` is a better place for it – that’s where we raise other errors with the headers we’re about to send, e.g. sending trailers without END_STREAM set.
2. The current header validation errors in `H2Stream.send_headers` all raise `ProtocolError` when you try to send something invalid. Do we need a special `InvalidHeaderBlockError`?
3. There are other checks in the received header validation pipeline that we don’t enforce on outgoing headers (e.g. the Connection header shouldn’t be present in a header block). If I’ve understood correctly, you could have hyper on both ends of a connection, and client-hyper could send a header block that server-hyper would reject. That seems a bit odd.

Lukasa:
1. So the reason it's in _build_header_frames is if it applies to sending all possible headers frames, including PUSH_PROMISE. That's an easy trap to fall into when we put it elsewhere: the headers for PUSH_PROMISEs also need validation. In fact, we probably overlooked this in #252.
2. ProtocolError is fine.
3. It's definitely odd. I'm open to fixing that up in some cases. =)

alexwlchan:
1. Okay, so next question: within `build_headers_frames`, how do we know if we’re in a trailer frame? Looking at `send_headers()`, it looks like maybe we want to be looking at `self.state_machine.trailers_send`, and if that’s `True`, we’re on a trailer?

2. Switched to ProtocolError.

3. I’ll raise a separate issue to track that weirdness, so that discussion about which header checks we might want to apply doesn’t clutter up this thread.

alexwlchan:
Other Q: is there an equivalent of `_build_header_frames` for received headers, that we can apply to the work done in #252?

Lukasa:
@alexwlchan 

1. Nah, in this instance it would be better to emit an internal-only event from the state machine that we suppress and do not emit, then use that to detect. Given that the state machine knows, we should get it to tell us.

alexwlchan:
In the two places we use `_build_headers_frames`, we ask the state machine to process input, but then assert we don’t get any events back. So would it be sensible to emit the internal-only event here, and pass that event into `_build_headers_frames`? This is all internal, so changing the argument signatures shouldn’t be (too) much of a problem.

Lukasa:
Yes, that's absolutely fine. =)

h2/events.py
```diff
@@ -136,6 +136,51 @@ def __repr__(self):
         )
 
 
+class ResponseSent(object):
```
alexwlchan: These are just ResponseReceived and TrailerReceived redressed with different names. I omitted a few of the extra fields, because we don’t have the logic to fill them in. (Unless I missed something.) Was that right?

h2/utilities.py
```diff
@@ -291,29 +295,101 @@ def _check_host_authority_header(headers, hdr_validation_flags):
     host_header_val = None
 
     for header in headers:
-        if header[0] == b':authority':
+        if header[0] in (b':authority', u':authority'):
```
alexwlchan: I’m sneaking this change in here, because it slipped through in #252, and it caused some tests to fail when I was playing with the sent header blocks. Okay? This mirrors what we do in `authority_from_headers`.

alexwlchan:
First stab at this pushed. Once again broken a lot of the tests – I’m guessing I’ve bumped a stream ID somewhere? I’m not quite sure what I’ve broken, so I’m going to leave this for tonight.

h2/events.py
Lukasa: So I think we'd like to clarify in some way that these are internal events. It's probably enough to put appropriate docstrings on them. Otherwise this is fine.

Out of interest, why don't we need RequestSent?

Lukasa: If the answer is that we probably do, can we replace both with a singular `HeadersSent`? When `TrailersSent` exists, I think those two names are clear enough.

h2/stream.py
```diff
@@ -1018,14 +1018,22 @@ def stream_reset(self, frame):
...
         Helper method to build headers or push promise frames.
         """
         # We need to lowercase the header names, and to ensure that secure
         # header fields are kept out of compression contexts.
-        headers = _lowercase_header_names(headers)
-        headers = secure_headers(headers)
+        try:
+            is_trailer = isinstance(events[0], TrailersSent)
+        except IndexError:
+            is_trailer = False
```
Lukasa: Can we have a blank line after this block?

h2/utilities.py
```diff
@@ -291,29 +295,101 @@ def _check_host_authority_header(headers, hdr_validation_flags):
...
+        elif header[0] in (b'host', u'host'):
+            host_header_val = header[1]
+
+        yield header
+
+    _validate_host_authority_header(
+        authority_header_val, host_header_val)
+
+
+def validate_sent_headers(headers, hdr_validation_flags):
```
Lukasa: I think this function needs a better name. Alternatively, we should normalize *first*, and then validate. Keeping `_lowercase_header_names` as a separate method that applies before the validation step doesn't seem unreasonable to me.

h2/utilities.py
Lukasa: Can we undo that and consider it a refactor for a separate PR? It makes it hard to confirm that there aren't function changes. =)

h2/utilities.py
Lukasa: This shouldn't actually be needed here because received header blocks should always be bytes: they only get optionally decoded after validation. However, given that this loop is now basically identical to the one in `_check_send_authority_host_header`, I recommend hoisting this loop out to `_validate_host_authority_header` too. Let's duplicate as little code as possible.

h2/stream.py
alexwlchan: Done.

h2/events.py
```diff
@@ -136,6 +136,56 @@ def __repr__(self):
         )
 
 
+class HeadersSent(object):
```
Lukasa: Part of me wonders if we should prefix these class names with underscores, just to make *really sure* we don't get confused.

h2/events.py
```diff
@@ -136,6 +136,56 @@ def __repr__(self):
...
+    """
+    The HeadersSent event is fired whenever headers are received.  This event
+    carries the HTTP headers and the stream ID of the new stream.
+
+    This is an internal event, used to distinguish between headers
+    and trailers.
+
+    .. versionadded:: 2.5.0
```
Lukasa: We don't need a versionadded marker here: that's only necessary for public classes.

Lukasa: Indeed, it's at real risk of being wrong as I may well backport this change.

h2/utilities.py
alexwlchan: Done.

alexwlchan:
Applied markups and rebased against master.

h2/stream.py
```diff
@@ -787,15 +789,14 @@ def push_stream_in_band(self, related_stream_id, headers, encoder):
         # Because encoding headers makes an irreversible change to the header
         # compression context, we make the state transition *first*.
 
-        # This does not trigger any events.
```
Lukasa: You removed this comment, but it seems to still be accurate.

Lukasa: Note that push promise is going to put some tricky assumptions to you: in particular, the client/server dichotomy no longer applies quite so cleanly. The headers in a push-promise frame are effectively *client* headers sent *by* a server.

Lukasa: That may mean we need a third event here, which you can use to help calculate your state tuple.

h2/stream.py
Lukasa: That would be ideal. =)

h2/utilities.py
```diff
@@ -279,33 +280,30 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
...
-    # server side or looking at trailer blocks.
-    if hdr_validation_flags.is_client or hdr_validation_flags.is_trailer:
-        for header in headers:
-            yield header
-        return
+    Given the :authority and Host headers from a request block that isn't
+    a trailer, check that:
+     1. At least one of these headers is set.
+     2. If both headers are set, they match.
+    Raises ``exception_class`` if the header block is invalid.
```
Lukasa: This comment is present but not acted on.

h2/utilities.py
```diff
@@ -279,33 +280,30 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
...
 
+    :param headers: The HTTP header set.
+    """
     # We use None as a sentinel value.  Iterate over the list of headers,
     # and record the value of these headers (if present).  We don't need
     # to worry about receiving duplicate :authority headers, as this is
     # enforced by the _reject_pseudo_header_fields() pipeline.
     #
     # TODO: We should also guard against receiving duplicate Host headers.
+    # TODO: We should guard against sending duplicate headers.
```
Lukasa: We probably don't need this #TODO twice. ;)

h2/utilities.py
```diff
@@ -319,13 +317,82 @@ def _check_host_authority_header(headers, hdr_validation_flags):
...
+    :authority or a Host header, or if a header block contains both fields,
+    but their values do not match.
+    """
+    # We only expect to see :authority and Host headers on request header
+    # blocks that aren't trailers, so skip this validation if we're on the
+    # server side or looking at trailer blocks.
+    if hdr_validation_flags.is_client or hdr_validation_flags.is_trailer:
+        return headers
+
+    return _validate_host_authority_header(headers)
```
Lukasa: Shiny, nice change! I had some vague concerns about what we were going to do in the absence of `yield from`, but changing this function to always return a generator without having any `yield` statements in it is a great solution to that problem. Well done!

Lukasa:
Alright, fab. Aside from the problems with pushes and a few minor style notes, this is getting much better!

We'll want to add a chunk of tests, much like the ones we have for the current validation pipeline, to confirm that we now forbid emitting the same headers we've been rejecting. This should be easy enough to do, though, because you just need to fire them into the connection as the first client request and watch it explode. =)

h2/stream.py
```diff
@@ -787,15 +805,14 @@ def push_stream_in_band(self, related_stream_id, headers, encoder):
...
         events = self.state_machine.process_input(
             StreamInputs.SEND_PUSH_PROMISE
         )
-        assert not events
+        print("@@AWLC events is %r" % events)
```
alexwlchan: Clearly this debugging line should come out before this gets merged. (Oops!)

alexwlchan:
@Lukasa I'll continue making markups and try to get the existing tests passing when I get home from work.

Any ideas why I've broken so much of the header indexing? e.g. https://travis-ci.org/python-hyper/hyper-h2/jobs/146619142#L332

h2/stream.py
```diff
@@ -216,23 +224,31 @@ def send_new_pushed_stream(self, previous_state):
...
+        be a client.
         """
         assert self.client is None
         self.client = True
         self.headers_sent = True
-        return []
+
+        event = _PushPromiseSent()
+        event.stream_id = self.stream_id
+        return [event]
```
Lukasa: This doesn't seem right. When we receive a push promise we spin *two* state machine methods. Conceptually, one of them owns the pushed headers and the other one does not.

We don't need a `PushPromiseSent` event in this case, because we *receive* a `PushPromiseReceived` event.

h2/stream.py
```diff
@@ -132,7 +133,12 @@ def request_sent(self, previous_state):
         """
         self.client = True
         self.headers_sent = True
-        return []
+
+        # TODO: Is it correct that we'd never send trailers on a response,
+        # or have I misread RFC 7230?
```
Lukasa: You've misread it: you can send trailers whenever you want.

Lukasa:
Hey @alexwlchan, what's the status here?

alexwlchan:
Sorry @Lukasa, last week was tricky and I didn’t have time for OSS stuff. Hoping to get back to it in the next day or so.

Lukasa:
That's ok, there's no rush. =)

alexwlchan:
Made a few markups, now I need to sort out the internal events that we fire (in particular, I now have a `_PushPromiseSent()` event but never actually use it). Looking through the sending methods on `H2StreamStateMachine`, this is what I think I should be doing:

* `request_sent() / response_sent()`, the logic should be:

   - if we've never sent headers, emit `_HeadersSent()`
   - if we’ve already sent headers, this is a trailer, so emit `_TrailersSent()`

  This is already implemented for `response_sent()`.

  (Also, can a server send requests?)

* `send_new_pushed_stream()` -- per your comment, we do nothing, because we receive a `PushPromiseReceived`

* `send_push_promise()` -- we should emit a `_PushPromiseSent` event.

* `send_informational_response()` and `send_alt_svc()` -- not sure what I should be doing here.

Plus tidying up all the places we assert we don‘t receive events, and updating the state tuples.

@Lukasa Does that sound sensible, or have I missed/misunderstood something?

Lukasa:
This sounds generally sensible. Let me elaborate on some points, because they touch on a series of bits of trickiness that I think the code doesn't necessarily make entirely clear (sadly).

1. Can a server send requests? Yes, but they are *pushed* requests and are contained entirely in a PUSH_PROMISE frame. No HEADERS frame a server emits can ever be for a request.
2. `send_new_pushed_stream` does not lead to an event being emitted. Prior to this change, no sending events methods emit events.
3. `send_new_pushed_stream` vs `send_push_promise`. We don't actually need both of these to emit an internal event. 

    The reason for this can be seen in `H2Connection.push_stream`. In short, though: a PUSH_PROMISE frame actually causes *two* stream-level state transitions. The first is for the stream on which the PUSH_PROMISE frame is sent (logically the *parent* stream). This stream has `H2Stream.push_stream_in_band` called on it, leading to `send_push_promise` being called, which currently returns no event. The second is for the stream that the PUSH_PROMISE *creates* (logically the *child* stream). This stream has `H2Stream.locally_pushed` called on it, which leads to `send_new_pushed_stream` being called on it and emitting the `PushPromiseReceived` event.

    This is a very long-winded way of saying: only *one* of these two methods should to return an event that we search for. More importantly, though: we don't need them to! Because, when we push a stream, these are *always* request headers. They cannot be trailers, they cannot be response headers, they must *always* be request headers. So we don't need an event at all: we know exactly what's going on here.
4. You don't need to worry about ALT_SVC, it has nothing to do with headers.
5. Informational responses are 1XX responses. They should be treated as ordinary response headers, though in the future we may want to validate them separately as they technically have different rules (e.g. no content-length allowed).

Does that clarify?

alexwlchan:
Okay, I think that all makes sense. I think the latest commit emits the correct events based on what you’ve said – could you double-check please? Once that’s done, I can start writing tests to check we reject the correct header set.

I’m also confused by the problem in the header indexing tests; I’m going to play around with git bisect to see where that started happening.

alexwlchan:
Found the header indexing bug: at some point I swapped the order in which we normalise sent headers. It was originally _lowercase, then secure_, and I’d turned it into _secure, then lowercase_.

h2/events.py
```diff
@@ -136,6 +136,52 @@ def __repr__(self):
...
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    def __init__(self):
+        #: The Stream ID for the stream this response was made on.
+        self.stream_id = None
+
+        #: The headers.
+        self.headers = None
```
alexwlchan: I just copied this field from some of the other internal events. But I can’t actually see where it gets populated on, say, `HeadersReceived`. Do we actually need it for these internal events, given that all we care about is whether this is a trailer or not?

Lukasa: This ordinarily gets populated by the stream object itself, because the state machine doesn't get access to the headers. We don't need it here. =) Nor do we even really need the stream ID.

h2/events.py
```diff
@@ -136,6 +136,52 @@ def __repr__(self):
...
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    def __init__(self):
+        #: The Stream ID for the stream on which these trailers were received.
+        self.stream_id = None
+
+        #: The trailers themselves.
+        self.headers = None
```
Lukasa: Similarly we don't need these.

h2/stream.py
```diff
@@ -131,8 +132,13 @@ def request_sent(self, previous_state):
         Fires when a request is sent.
         """
         self.client = True
-        self.headers_sent = True
-        return []
+        if not self.headers_sent:
+            self.headers_sent = True
+            event = _HeadersSent()
+        else:
+            event = _TrailersSent()
```
Lukasa: Are you sure this if branch is necessary? If it is, I think we need `self.trailers_sent` to be populated as well.

h2/stream.py
```diff
@@ -1023,14 +1030,24 @@ def stream_reset(self, frame):
     def _build_headers_frames(self,
                               headers,
                               encoder,
-                              first_frame):
+                              first_frame,
+                              events):
```
Lukasa: Rather than pass the events in here, can we hoist the building of the validation flags up a level? Maybe we need to write a helper function to do that, but that shouldn't be hard, and it avoids this method doing too much.

h2/stream.py
Lukasa: uh...actually, I think this branch is un-hittable. It looks a bit like maybe clients can't send trailers at all. That might be a separate bug.

Lukasa: Ah, no, they can, but it happens in `response_sent`. So I'm right, this branch is unhittable and we don't need it.

Lukasa:
Ok, I've got a few smaller notes about an unreachable branch and some other stuff, but this is looking broadly right now. =D Great work so far!

h2/events.py
```diff
@@ -136,6 +136,52 @@ def __repr__(self):
         )
 
 
+class _HeadersSent(object):
+    """
+    The _HeadersSent event is fired whenever headers are received.  This event
```
alexwlchan: > whenever headers are received

No.

test/test_invalid_headers.py
```diff
@@ -95,30 +97,49 @@ class TestFilter(object):
...
         """
         hdr_validation_flags = h2.utilities.HeaderValidationFlags(
             is_client=is_client,
             is_trailer=is_trailer,
         )
         try:
-            assert headers == h2.utilities.validate_headers(
-                headers, hdr_validation_flags)
+            assert headers == list(validation_function(
+                headers, hdr_validation_flags))
```
alexwlchan: Possibly this is the wrong fix; this was an attempt to fix a failure when a client tries to send an empty list of headers which aren't trailers:

```
        try:
>           assert headers == validation_function(
                headers, hdr_validation_flags)
E               assert [] == <generator object _validate_...thority_header at 0x279b910>
E                 Full diff:
E                 - []
E                 + <generator object _validate_host_authority_header at 0x279b910>

test/test_invalid_headers.py:127: AssertionError
------------------------------------------------------------------------- Hypothesis -------------------------------------------------------------------------
Falsifying example: test_range_of_acceptable_outputs(self=<test_invalid_headers.TestFilter at 0x28f0510>, headers=[], validation_function=validate_sent_headers, is_client=True, is_trailer=False)
============================================================ 1 failed, 31 passed in 3.04 seconds =============================================================
```

test/test_invalid_headers.py
```diff
@@ -95,30 +97,49 @@ class TestFilter(object):
...
-        (False, True),
-        (False, False),
+    validation_functions = [
+        h2.utilities.validate_headers,
+        h2.utilities.validate_sent_headers
     ]
+    validator_combinations = list(itertools.product(
+        validation_functions,
+        (True, False),
+        (True, False),
```
alexwlchan: Notes to self:

1. Do you definitely need to cast to a list?
2. Better if this had the logic for constructing header validation flags, and those were passed directly into the tests; hoist a bit more logic out of the tests themselves.

alexwlchan:
Notes for me before I go back to work (thinking aloud more than anything else):

* How does this interact with the config options for skipping this validation?
* Eventually we'd like to reject all the invalid header blocks in `test_invalid_headers.py` when we try to send them, for now we'll need to skip at least some of them.
* There was talk of writing some documentation for this change. Write some!

alexwlchan:
Okay, final update from me for the weekend. Comments/questions for @Lukasa:

* We need a flag to skip this header validation for people who like to live on the edge. Do we need a generic "skip all outbound header validation" flag, or should we break it down into more granular flags for each check we perform? I lean towards the former.

  Once I know what this flag looks like, I can update the docs.

* Tests are failing because we don’t have 100% coverage of `authority_from_headers()` in the utilities module – specifically, the `return None` if we don’t find an :authority header. Because any headers we try to send without an :authority header are being rejected before we hit this function. 😄 

  Probably the correct thing here is to test the config object really will let you send invalid headers, including one set without an :authority header.

Lukasa:
> Do we need a generic "skip all outbound header validation" flag, or should we break it down into more granular flags for each check we perform? I lean towards the former.

The former, for sure. Probably we need two flags, `validate_outbound_headers` (defaults to `True`) and `normalize_outbound_headers` (defaults to `True`), but for this PR feel free to just implement the first.

We obviously also want symmetrical flags for inbound headers.

> Probably the correct thing here is to test the config object really will let you send invalid headers, including one set without an :authority header.

Yup, sounds good.

alexwlchan:
I’ve squashed some of the earlier commits, and rebased against master.

1. The latest commit adds three of the four config options suggested in your previous comment. I haven‘t added a `normalize_outbound_headers` flag because, AFAICT, that’s not something we ever do, so it can’t be turned off. Or have I missed something here?

2. The proliferation of boolean config options – all using very similar code – feels like a code smell. I’m tempted to wrap it all up in some sort of `BooleanConfigOption` descriptor, and cut out the repetitive code – but it feels a bit like I’m just wielding a descriptor-shaped hammer. What do you think?

3. I’m trying to work out where we should actually make use of these settings. The `H2Connection` class owns the config, but it’s `H2Stream` that actually calls the validation/normalization functions. I’m struggling to think of a good way to pass the config options down into `H2Stream`; unless there’s a way for `H2Stream` to peek "up" at its owner and read the config from there?

Lukasa:
> I haven‘t added a normalize_outbound_headers flag because, AFAICT, that’s not something we ever do, so it can’t be turned off.

I presume you mean *inbound* headers? In which case, that's correct. =)

> The proliferation of boolean config options – all using very similar code – feels like a code smell. I’m tempted to wrap it all up in some sort of BooleanConfigOption descriptor, and cut out the repetitive code – but it feels a bit like I’m just wielding a descriptor-shaped hammer. What do you think?

Let's open a new issue to track this, and then deal with it as a refactor in a separate PR. I'm happy to remove the extra code, just want to try to reduce scope creep in this PR. =)

> I’m trying to work out where we should actually make use of these settings. The H2Connection class owns the config, but it’s H2Stream that actually calls the validation/normalization functions. I’m struggling to think of a good way to pass the config options down into H2Stream; unless there’s a way for H2Stream to peek "up" at its owner and read the config from there?

We should pass the config object into the `H2Stream` on construction.

alexwlchan:
Oh so nearly there. We’re now passing the `config` object to the `H2Stream`, skipping the validation/normalization where appropriate, and have some tests for the new function. Shiny!

Tests are failing because of the issue fixed in #285; once that’s merged, I’ll rebase atop that. Applying the fix locally gets the tests passing.

The one last problem is the `normalize_sent_headers` parameter. I’m struggling to think of a good test for it, and without that test coverage is <100%. Ideally we’d verify that sent headers don’t get modified when you disable this setting, but AFAICT you can’t see what headers get sent after you’ve called `send_headers` – which seems perfectly sensible. Any ideas?

Lukasa:
@alexwlchan Feel free to use the `hpack` module to import a HPACK decoder that can decode the HEADERS frame we emit. =)

alexwlchan:
At last a green tick, hurrah! I believe this is now ready to merge – at the very least, there’s nothing else on my todo list for it.

The next few days are a bit busier, so progress might slow down again a bit. (Sorry, the last few weeks have been a bit up-in-the-air – obviously the best time to start a big patch. 😉 Hopefully the next few patches will be a bit smaller.)

Lukasa:
Heh, thanks @alexwlchan for all this work! I'll do a deep-dive review shortly, but I'm really pleased with how this patch has been shaping up.

h2/connection.py
```diff
@@ -512,6 +512,7 @@ def _begin_new_stream(self, stream_id, allowed_ids):
             self.remote_settings.initial_window_size
         )
         s.inbound_flow_control_window = self.local_settings.initial_window_size
+        s.config = self.config
```
Lukasa: This looks like it just gets casually patched onto the `H2Stream`. Can we update `H2Stream.__init__` to create this member variable initialized to `None`?

h2/config.py
```diff
@@ -77,3 +106,61 @@ def header_encoding(self, value):
         if value is True:
             raise ValueError("header_encoding cannot be True")
         self._header_encoding = value
+
+    @property
+    def validate_sent_headers(self):
```
Lukasa: Minor nit, but it would be nicer to rename this to `validate_outbound_headers` for symmetry with `validate_inbound_headers`. Alternatively, swap to `sent`/`received`.

h2/stream.py
```diff
@@ -684,6 +691,9 @@ def __init__(self, stream_id):
         # The authority we believe this stream belongs to.
         self._authority = None
 
+        # The configuration for this stream.
+        self.config = H2Configuration()
```
Lukasa: Don't create a brand new config object. In fact, ignore my prior comment: we should probably provide this on the constructor, as we cannot reasonably expect streams to exist without it.

h2/stream.py
```diff
@@ -754,12 +764,12 @@ def send_headers(self, headers, encoder, end_stream=False):
...
 
-        # This does not trigger any events.
         events = self.state_machine.process_input(input_)
-        assert not events
 
         hf = HeadersFrame(self.stream_id)
-        frames = self._build_headers_frames(headers, encoder, hf)
+        hdr_validation_flags = self._build_hdr_validation_flags(events)
+        frames = self._build_headers_frames(
+            headers, encoder, hf, hdr_validation_flags)
```
Lukasa: Sorry, the general project style for long lines like this is:

```python
frames = self._build_headers_frames(
    headers, encoder, hf, hdr_validation_flags
)
```

h2/stream.py
```diff
@@ -795,7 +805,9 @@ def push_stream_in_band(self, related_stream_id, headers, encoder):
 
         ppf = PushPromiseFrame(self.stream_id)
         ppf.promised_stream_id = related_stream_id
-        frames = self._build_headers_frames(headers, encoder, ppf)
+        hdr_validation_flags = self._build_hdr_validation_flags(events)
+        frames = self._build_headers_frames(
+            headers, encoder, ppf, hdr_validation_flags)
```
Lukasa: Same here.

h2/stream.py
```diff
@@ -918,11 +930,12 @@ def receive_headers(self, headers, end_stream, header_encoding):
...
-        conn_state = HeaderValidationFlags(
-            is_client=self.state_machine.client,
-            is_trailer=isinstance(events[0], TrailersReceived)
-        )
-        headers = validate_headers(headers, conn_state)
+        if self.config.validate_inbound_headers:
+            hdr_validation_flags = HeaderValidationFlags(
+                is_client=self.state_machine.client,
+                is_trailer=isinstance(events[0], TrailersReceived)
+            )
```
Lukasa: This can also use `_build_hdr_validation_flags` if we slightly broaden that method.

h2/stream.py
```diff
@@ -1020,17 +1033,36 @@ def stream_reset(self, frame):
 
         return [], events
 
+    def _build_hdr_validation_flags(self, events):
+        """
+        Constructs a set of header validation flags for use when normalizing
+        and validating header blocks.
+        """
+        try:
+            is_trailer = isinstance(events[0], _TrailersSent)
```
Lukasa: Let's add `TrailersReceived` to the list of trailery events, because we can then use this method in `receive_headers` too.

h2/utilities.py
```diff
@@ -319,13 +317,82 @@ def _check_host_authority_header(headers, hdr_validation_flags):
...
         )
 
     # If we receive both headers, they should definitely match.
     if authority_present and host_present:
         if authority_header_val != host_header_val:
             raise ProtocolError(
-                "Received mismatched :authority and Host headers: %r / %r" %
-                (authority_header_val, host_header_val)
+                "Request header block must have matching :authority and "
+                "Host headers: %r / %r"
```
Lukasa: Let's change this to a statement of what is wrong: `"Request header block has mismatched :authority and Host headers"`

Lukasa:
Awesome, this is looking really good! I have left some inline notes, but generally speaking this is looking pretty shiny.

h2/config.py
alexwlchan: I'll go for `outbound` -- that's more accurate, as we validate the headers when they're outbound, not necessarily sent.

alexwlchan:
Ding. Review markups made. And a mere 77 comments later, are we done? :smile:

h2/stream.py
```diff
@@ -1020,17 +1031,42 @@ def stream_reset(self, frame):
...
+        """
+        Constructs a set of header validation flags for use when normalizing
+        and validating header blocks.
+        """
+        try:
+            is_trailer = isinstance(
+                events[0], (_TrailersSent, TrailersReceived)
+            )
+        except IndexError:
+            is_trailer = False
```
Lukasa: Hrm. What's this fallback for, and when do we hit it?

Lukasa:
I have one question, and then we should be safe.

h2/stream.py
alexwlchan: There are cases where we don't emit an internal event &ndash; push promise is the example that springs to mind &ndash; and so `events` is empty. It's the places where we still have `assert not events`.

Lukasa: Aha, ok. This feels like an easy source of bugs, but for now let's put a comment here.

alexwlchan: Comment added, along with a TODO to make this unnecessary in future.

Lukasa:
\o/ Thanks @alexwlchan, this turned into a marathon patch, but it looks really good! :sparkles:

PR #264, opened 2016-07-22
initialise content length to zero if HEAD request

aidylewis:


Lukasa:
There we go, builds now! =) Seems we have a linting issue.

Lukasa:
This supersedes #258.

test/test_head_request.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestHeadRequest(object):
+
+        example_request_headers = [
+            (u':authority', u'example.com'),
+            (u':path', u'/'),
+            (u':scheme', u'https'),
+            (u':method', u'HEAD'),
+            (u':content_length', u'1'),
```
Lukasa: Ah, sorry: this test is not quite right.

Firstly, the `content-length` header should be in the *response*, not the request. Secondly, the header name is `content-length`, not `:content-length`.

aidylewis:
I've squashed the last two commits. This **should** be final https://github.com/python-hyper/hyper-h2/pull/264/commits/98e2cd6fa8fdadf96267b565fbd3b857fe415a92 commit

aidylewis:
Are we now OK to squash?

Lukasa:
Yup, looks good to me! :sparkles:

Lukasa:
Merged. Thank you so much @aidylewis! :sparkles: :cake: :sparkles:

aidylewis:
OK. I am off to lunch now. What next? I was thinking of creating a command line load tool from this library. WDYT?

Lukasa:
I think that's an extremely interesting idea! =) Why not prototype it, see how it looks?

PR #265, opened 2016-07-25
Add configuration object.

Lukasa:
As discussed in #246, this adds a basic configuration object. The goal here is to resist the proliferation of flags to the HTTP/2 connection object.

### DIFF
--- a/h2/config.py
+++ b/h2/config.py
@@ -9,7 +9,7 @@ Objects for controlling the configuration of the HTTP/2 stack.
 
 class H2Configuration(object):
     """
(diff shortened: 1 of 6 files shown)


h2/config.py
```diff
@@ -0,0 +1,79 @@
...
+h2/config
+~~~~~~~~~
+
+Objects for controlling the configuration of the HTTP/2 stack.
+"""
+
+
+class H2Configuration(object):
+    """
+    An object that controlls the way a single HTTP/2 connection behaves.
```
alexwlchan: /s/controlls/controls

PR #276, opened 2016-08-05
Explicitly request newer PyPy.

Lukasa:
Apparently this works now?


Lukasa:
Wow, it really does. That's amazing.

PR #277, opened 2016-08-06
Twisted import fixes

kracekumar:
- Remove unused import 
- Club twisted imports from the same module in single line


Lukasa:
Thanks @kracekumar, [welcome to the team](http://python-hyper.org/en/latest/one-of-the-team.html)!

PR #280, opened 2016-08-07
Fix link to Priority lib

drpoggi:
Not sure if the link is actually broken or if the documentation isn't building correctly, just thought I'd bring it your attention.

### DIFF
--- a/docs/source/advanced-usage.rst
+++ b/docs/source/advanced-usage.rst
@@ -218,4 +218,4 @@ version 3.0 of hyper-h2.
 
 
 .. _RFC 7540: https://tools.ietf.org/html/rfc7540
-.. _an implementation: http://python-hyper/priority/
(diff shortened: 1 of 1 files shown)


alexwlchan:
I suspect the link is actually meant to be <http://python-hyper.org/priority/>, because `http://python-hyper/` isn’t a valid address.

drpoggi:
Yup that would make more sense, updated.

Lukasa:
Thanks for that catch! :cookie:

Lukasa:
Thanks @drpoggi, [welcome to the team](http://python-hyper.org/en/latest/one-of-the-team.html)!

drpoggi:
Thanks @Lukasa, I look forward to making more substantial contributions in the future!

Lukasa:
We're looking forward to it too, @drpoggi. =D

Issue #271, opened 2016-08-02
Reject header fields with invalid whitespace characters.


Lukasa:
Relevant: golang/go#16572.

We should forbid implementations from emitting header fields with invalid whitespace characters. Generally, that means surrounding whitespace of any kind, either leading or trailing, from both header field names and header field values.

alexwlchan:
Is there anything left to do, or is this fixed by #281?

Lukasa:
I think this is fixed by #281.

PR #281, opened 2016-08-07
Reject header fields with surrounding whitespace

sunu:
Trying to write a fix for #271.
I was wondering if whitespace should be stripped before sending the headers in [_build_headers_frames](https://github.com/python-hyper/hyper-h2/blob/5d880ad3dc8daf721da25c9eb1dbb61f3f3f2c29/h2/stream.py#L1023).


alexwlchan:
> I was wondering if whitespace should be stripped before sending the headers in _build_headers_frames.

I think that’s sensible. I’d suggest adding it as a separate step in `_build_headers_frames`.

(I’m midway through pulling out those normalisation steps into their own function – see `normalize_sent_headers` in #261 – but your patch is much smaller than mine, so it’ll probably get merged first. I’ll pull your change into my patch after yours has been merged.)

test/test_invalid_headers.py
```diff
@@ -45,6 +45,10 @@ class TestInvalidFrameSequences(object):
         base_request_headers + [('upgrade', 'super-protocol/1.1')],
         base_request_headers + [('te', 'chunked')],
         base_request_headers + [('host', 'notexample.com')],
+        base_request_headers + [(' name', 'name with leading space')],
+        base_request_headers + [('name ', 'name with trailing space')],
+        base_request_headers + [('name', ' value with leading space')],
+        base_request_headers + [('name', 'value with trailing space ')],
```
alexwlchan: 👍 for helpful descriptions

sunu:
Thanks for the quick feedback @alexwlchan. Updated the PR accordingly :)

h2/stream.py
```diff
@@ -1111,6 +1112,20 @@ def _lowercase_header_names(headers):
             yield (header[0].lower(), header[1])
 
 
+def _strip_surrounding_whitespace(headers):
+    """
+    Given an iterable of header two-tuples, strip both leading and trailing
+    whitespace from header both header names and header values. This generator
```
Lukasa: Nitpicking: You have the word "header" here too many times. Consider changing to: "whitespace from both header names and header values."

h2/utilities.py
```diff
@@ -214,6 +217,22 @@ def _reject_uppercase_header_fields(headers, hdr_validation_flags):
...
+    """
+    for header in headers:
+        if header[0].strip() != header[0]:
+            raise ProtocolError(
+                "Received header name surrounded by whitespace %s" % header[0])
+        if header[1].strip() != header[1]:
+            raise ProtocolError(
+                "Received header value surrounded by whitespace %s" % header[1]
+            )
+        yield header
```
Lukasa: So for a first pass I think this is totally fine. If you want an extension task, it would be really interesting to prototype different ways we could do this along with benchmarks, to see how they perform. This code is in a quite hot section of the code, which means that any efficiencies we can eke out of each function are huge.

This doesn't block the PR at all, just a suggestion for extension work. :grin:

Lukasa:
@sunu This is beautiful, thank you! I have one tiny nit inline that would be nice to have addressed. Additionally, it would be good if you could either rebase onto the current master or merge in the current master. :smile:

Thanks as well to @alexwlchan for the excellent code review! :heart:

h2/utilities.py
alexwlchan: Nitpicking in the extreme, but do we want `%r` instead of `%s` as the format specifiers?

It’s much easier to see the trailing whitespace (and how much of it there is) if the entire name/value is surrounded by quote marks.

Lukasa: Yeah, let's do %r. %r was already *effectively* true on Python 3, so this works well.

sunu:
@Lukasa @alexwlchan Updated the PR. Here's something I did for benchmarking different methods of detecting surrounding whitespace. https://gist.github.com/sunu/207c1901ac051d8a8c603690c1808f57

It seems there is a slightly better way to do it. But I'm not sure if the way I'm benchmarking the methods is actually the proper way. Also, not sure if there are any other methods to do this that I have not benchmarked.

Lukasa:
Thanks for the benchmarking @sunu! I added a comment under the gist to suggesting something we could do to get a complete feel for the various timings.

sunu:
Hey @Lukasa,
I tried running timeit for each word and each function. Here is the results I got.
```
In [12]: for word in words:
    ...:     print("-"*80)
    ...:     print("for word {0}".format(word))
    ...:     for func in [check1, check2, check3, check4]:
    ...:         print(func.__name__)
    ...:         %timeit -n9999 -r99 func(word)
    ...:         
    ...:         
    ...:         
    ...:         
--------------------------------------------------------------------------------
for word  sdfsdf
check1
9999 loops, best of 99: 128 ns per loop
check2
9999 loops, best of 99: 173 ns per loop
check3
9999 loops, best of 99: 324 ns per loop
check4
9999 loops, best of 99: 1.16 µs per loop
--------------------------------------------------------------------------------
for word dfsf 
check1
9999 loops, best of 99: 168 ns per loop
check2
9999 loops, best of 99: 180 ns per loop
check3
9999 loops, best of 99: 205 ns per loop
check4
9999 loops, best of 99: 1.04 µs per loop
--------------------------------------------------------------------------------
for word sdfs
check1
9999 loops, best of 99: 167 ns per loop
check2
9999 loops, best of 99: 149 ns per loop
check3
9999 loops, best of 99: 333 ns per loop
check4
9999 loops, best of 99: 1.05 µs per loop
--------------------------------------------------------------------------------
for word  dsf sd 
check1
9999 loops, best of 99: 128 ns per loop
check2
9999 loops, best of 99: 178 ns per loop
check3
9999 loops, best of 99: 203 ns per loop
check4
9999 loops, best of 99: 996 ns per loop
--------------------------------------------------------------------------------
for word  dfsdfs
check1
9999 loops, best of 99: 125 ns per loop
check2
9999 loops, best of 99: 174 ns per loop
check3
9999 loops, best of 99: 326 ns per loop
check4
9999 loops, best of 99: 1.18 µs per loop
--------------------------------------------------------------------------------
for word 23dwe 
check1
9999 loops, best of 99: 168 ns per loop
check2
9999 loops, best of 99: 180 ns per loop
check3
9999 loops, best of 99: 203 ns per loop
check4
9999 loops, best of 99: 1.04 µs per loop
```

Seems like the first method is faster in all but one case.

Lukasa:
So that's fairly marginal, but good enough for me to want to pursue it. Do you mind changing the reject function to use the new mechanism?

You'll also need to merge in the current master branch (or rebase onto it). =)

sunu:
Huh, didn't consider indexing into byte strings produces ints on Python 3. @Lukasa should I just use `strip` then? Converting bytes to unicode and then back will introduce more overhead anyway which is probably more than the time we will save.

alexwlchan:
I think we might be better off sticking with the current approach, but replacing `whitespace` with a `frozenset()` of the whitespace code point integers if we're using Python 3, and comparing to that instead?

(I also wonder if checking `char in frozenset(string.whitespace)`, where we cache the frozenset, might be marginally faster. In general, checking for set inclusion is O(1) whereas checking for string inclusion is O(n), but as `string.whitespace` is short, it might not matter.)

Lukasa:
@alexwlchan has a good suggestion, we can create a frozenset that contains both the single char `bytes` values (for Python 2) and the individual hex values (for Python 3) at import time that we then index into.

alexwlchan:
@Lukasa I was thinking of something ever so slightly different: we have two frozensets, and choose which one we want based on the Python version, i.e.:

```python
if python_version <= 2:
    whitespace = string.whitespace
else:
    whitespace = hex_values_for(string.whitespace)
```

I'd think the check becomes cheaper if you have a smaller set to check with. OTOH, set lookups are _O(1)_ in the usual case, so perhaps I'm overthinking it.

Lukasa:
Set lookup is a linear search for small sets, so yeah, let's do that instead.

sunu:
Oh, I pushed with a single combined frozenset.

h2/utilities.py
```diff
@@ -40,6 +42,11 @@
     b'proxy-authorization', u'proxy-authorization',
 ])
 
+if sys.version_info[0] == 3:
```
alexwlchan: I marginally prefer to do comparisons against Python 2, and treat that as the special case, because it future-proofs against the major Python version being incremented in future (how does this cope with Python 4?).

sunu: Fair enough. I'll fix.

sunu:
Quick question: if look up is linear for small sets, do we need to use them instead of say a tuple or a list?

Lukasa:
It's better to use them for two reasons. Firstly, "small" is ill-defined, and we may end up crossing it one day. Second, sets are *semantically* better. Using a set says "we will use this structure for lookups", whereas using a list says "we will use this structure for iteration".

sunu:
@Lukasa @alexwlchan PR updated. But it seems python-hyper.org is down.

Lukasa:
I've restarted the build.

h2/utilities.py
```diff
@@ -40,6 +42,11 @@
     b'proxy-authorization', u'proxy-authorization',
 ])
 
+if sys.version_info[0] == 2:  # pragma: no cover
```
Lukasa: If you change this comment to `# Python 2.X`, it will still exclude the coverage, and be a bit clearer.

h2/utilities.py
```diff
@@ -40,6 +42,11 @@
...
+    _WHITESPACE = frozenset(whitespace)
+else:  # pragma: no cover
```
Lukasa: Similarly, this comment can be `# Python 3.3 and higher`.

h2/utilities.py
```diff
@@ -215,6 +225,24 @@ def _reject_uppercase_header_fields(headers, hdr_validation_flags):
...
+    """
+    Raises a ProtocolError if any header name or value is surrounded by
+    whitespace characters.
+    """
+    for header in headers:
+        if header[0] and ((header[0][0] in _WHITESPACE) or
+                          (header[0][-1] in _WHITESPACE)):
+            raise ProtocolError(
+                "Received header name surrounded by whitespace %r" % header[0])
+        if header[1] and ((header[1][0] in _WHITESPACE) or
```
Lukasa: Do you need to test the existence of the header fields? I suppose you're only testing whether they're empty, but empty header field names or values are a bad idea and so we shouldn't tolerate them at this time.

Lukasa:
Alright, we're down way to the nitpicky details. Almost there!

h2/utilities.py
sunu: @Lukasa If I don't check for empty headers the test at https://github.com/python-hyper/hyper-h2/blob/7a80186834fd344628941bc39e333d0ea108652e/test/test_invalid_headers.py#L211 fails. Should I raise a ProtocolError here if the headers are empty?

Lukasa: Hrm. RFC 7230 allows empty header field values (though I wish it didn't), so we have to allow those too. However, we don't have to allow empty header field names, so let's remove that check and then, if that test still fails, we can amend the Hypothesis testing strategy to forbid empty header field names.

sunu:
@Lukasa PR updated.

Lukasa:
Thanks so much for this patch @sunu, and for your patience moving through the process. Per our [one of the team](http://python-hyper.org/en/latest/one-of-the-team.html) policy, you'll find a python-hyper organisation invite in your inbox. Welcome on board!

@alexwlchan, as always, thanks so much for your help as an extra set of eyes on the review here.

Issue #284, opened 2016-08-08
Seemingly inconsistent behaviour around H2Configuration objects


alexwlchan:
Spotted while trying to write tests for #261. Spinning this out as a separate issue because I don’t think it’s related to my changes, and I can readily reproduce it on master (commit d840a63).

We have the following test in `test_invalid_headers.py`:

```python
def test_transfer_encoding_trailers_is_valid(self, frame_factory):
    """
    Transfer-Encoding trailers is allowed by the filter.
    """
    headers = (
        self.base_request_headers + [('te', 'trailers')]
    )

    c = h2.connection.H2Connection(client_side=False)
    c.receive_data(frame_factory.preamble())

    f = frame_factory.build_headers_frame(headers)
    data = f.serialize()

    events = c.receive_data(data)
    assert len(events) == 1
    request_event = events[0]
    assert request_event.headers == headers
```

Passing on CI and locally, which is fine. But then I replace this line:

```python
    c = h2.connection.H2Connection(client_side=False)
```

with this, which I think should have the equivalent behaviour:

```python
    config = h2.config.H2Configuration(client_side=False)
    c = h2.connection.H2Connection(config=config)
```

Splat:

```
================================================================ FAILURES =================================================================
___________________________________ TestInvalidFrameSequences.test_transfer_encoding_trailers_is_valid ____________________________________

self = <test_invalid_headers.TestInvalidFrameSequences object at 0x104b7e2e8>, frame_factory = <helpers.FrameFactory object at 0x104b7e278>

    def test_transfer_encoding_trailers_is_valid(self, frame_factory):
        """
            Transfer-Encoding trailers is allowed by the filter.
            """
        headers = (
            self.base_request_headers + [('te', 'trailers')]
        )

        config = h2.config.H2Configuration(client_side=False)
        c = h2.connection.H2Connection(config=config)
        c.receive_data(frame_factory.preamble())

        f = frame_factory.build_headers_frame(headers)
        data = f.serialize()

        events = c.receive_data(data)
>       assert len(events) == 1
E       assert 0 == 1
E        +  where 0 = len([])

test/test_invalid_headers.py:86: AssertionError
======================================================== 1 failed in 0.03 seconds =========================================================
```

I’m going to keep digging and see if I can get to the bottom of this, because I can’t see any obvious reason why the change in approach would lead to different behaviour.

[...]

PR #285, opened 2016-08-08
Use a consistent version of the `client_side` parameter

alexwlchan:
This fixes the strange behaviour I was seeing in #284.

I wasn’t sure what sort of test I could write that would catch this, so I haven’t written one. Happy to take suggestions!


Lukasa:
Looks good, nice spot @alexwlchan! :custard:

Issue #283, opened 2016-08-08
Tidy up the repeated code in H2Configuration


alexwlchan:
When we merge #261, we’re going to have four boolean config options on `H2Configuration`. Each of these options has a getter/setter, but the associated code is very similar – it enforces that you can only set boolean values here.

It would be good to use something like a descriptor to reduce the amount of repeated code here.

Splitting off from #261 to avoid that growing in size any further.

alexwlchan:
Fixed in #286.

PR #286, opened 2016-08-11
Tidy up repeated code in h2.config

alexwlchan:
Using a descriptor for the boolean config options lets us cut down
on repetitive code for defining new boolean settings.  Also ensure
we actually use the validation code when setting up a new
`H2Configuration` object.

---

Addresses #283. Adding new boolean options is now a single-line change, not a load of copy-and-pasting.

### DIFF
--- a/h2/config.py
+++ b/h2/config.py
@@ -14,14 +14,15 @@ class _BooleanConfigOption(object):
     """
     def __init__(self, name):
         self.name = name
+        self.attr_name = '_%s' % self.name
 
     def __get__(self, instance, owner):
-        return getattr(instance, '_%s' % self.name)
+        return getattr(instance, self.attr_name)
 
     def __set__(self, instance, value):
         if not isinstance(value, bool):
             raise ValueError("%s must be a bool" % self.name)
-        setattr(instance, '_%s' % self.name, value)
+        setattr(instance, self.attr_name, value)
 
 
 class H2Configuration(object):
--- a/h2/config.py
+++ b/h2/config.py
@@ -7,6 +7,23 @@ Objects for controlling the configuration of the HTTP/2 stack.
 """
 
 
+class _BooleanConfigOption(object):
+    """
+    Descriptor for handling a boolean config option.  This will block
+    attempts to set boolean config options to non-bools.
+    """
+    def __init__(self, name):
+        self.name = name
+
+    def __get__(self, instance, owner):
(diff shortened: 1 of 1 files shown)


h2/config.py
```diff
@@ -6,6 +6,27 @@
...
+
+
+class _BooleanConfigOption(object):
+    """
+    Descriptor for handling a boolean config option.  This will block
+    attempts to set boolean config options to non-bools.
+    """
+    def __init__(self, name, default):
+        self.name = name
+        self.default = default
```
Lukasa: Do we need a default value here?

h2/config.py
```diff
@@ -6,6 +6,27 @@
...
+        self.data = weakref.WeakKeyDictionary()
```
Lukasa: Is it really necessary to use a weakref here? This is doubly true given that we're only storing bools, which are definitionally singletons and so cannot possibly cause us trouble to just store directly.

h2/config.py
alexwlchan: I suppose not. Actually, given that we supply the defaults to the constructor of `H2Configuration`, this is really just duplicating information. I'll remove it.

Lukasa:
So I'm not sure I like the specific descriptor approach we're using here. In particular, I think I don't see any reason not to simply have the descriptor pass through to private vars on the object via getattr/setattr, rather than maintain its own dictionary. I also don't know that we need to bother with a default value.

alexwlchan:
I don't feel strongly about it, so I'll have a stab at a different approach using private variables later.

alexwlchan:
New version that just sets private variables on the config objects, no mucking around with weakrefs.

h2/config.py
```diff
@@ -7,6 +7,23 @@
...
+    def __init__(self, name):
+        self.name = name
+
+    def __get__(self, instance, owner):
+        return getattr(instance, '_%s' % self.name)
+
+    def __set__(self, instance, value):
+        if not isinstance(value, bool):
+            raise ValueError("%s must be a bool" % self.name)
+        setattr(instance, '_%s' % self.name, value)
```
Lukasa: Can we cache this computation?

Lukasa:
So this is really close, I just have a small note about caching some computation if we can.

alexwlchan:
Caching added.

Lukasa:
Cool, I'm happy with this. Thanks so much @alexwlchan! :sparkles:

Issue #262, opened 2016-07-20
It’s possible to send header blocks from hyper-as-server that hyper-as-client will reject


alexwlchan:
Picking up a comment from #261:

We do a series of checks in the `validate_headers` pipeline for incoming header blocks, in particular:

- [x] All fields should be lowercase.
- [x] A Transfer-Encoding header must have value "trailers"
- [x] No blocks with the Connection header.
- [x] No duplicate pseudo-header fields and/or pseudo-header fields coming after an ordinary header field.
- [x] No header blocks with bad :authority/Host headers.

But we don’t enforce all these rules on header blocks that we send, which suggests that it’s possible to send a header block with hyper that we’d reject if we received. Which seems odd.

(1) is achieved by the `_lowercase_header_names` pipeline, and I’m adding (5) in #261. I think we should be enforcing (2)-(4) on header blocks that we send as well.

Does that seem sensible?

Lukasa:
I've updated the original comment to turn the list into a list of checkboxes so we can keep track. Feel free to tackle each one separately.

alexwlchan:
Note: if you’re going to tackle this, you should read the discussion on #246 about being able to skip these checks in certain cases.

alexwlchan:
All addressed in #289.

PR #289, opened 2016-08-17
Reject outbound header blocks with invalid headers

alexwlchan:
There are several rules we use to reject inbound header blocks as
invalid, but we don't enforce them on headers we're sending:

* A Transfer-Encoding header must have value "trailers".
* No blocks should have connection-specific headers.

These are rules mandated by RFC 7540. We already have the code to
police these rules, so apply it to outbound as well as inbound
headers.

Inbound headers always have byte strings for names/values, whereas
we may still have Unicode strings for the outbound headers (they're
encoded after this validation step), so we need to be tolerant
to Unicode headers here.

---

This is a first attempt at tackling #262. (2) and (3) are now enforced on outgoing headers.

Slightly complicated by the Unicode issue, which makes this pipeline a tad slower throughout. Is this is a sensible way to handle it?

I haven’t done the checks for duplicate pseudo-header fields yet, because that’s a bit more complicated. Specifically, we do `if header[0].startswith(b':')`, which only works if `header[0]` is a bytestring. You get a `TypeError` if you call it on a Unicode string. Doesn’t have to be done here, but any suggestions?

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -44,10 +44,6 @@ Bugfixes
   ``trailers``, per RFC 7540 Section 8.1.2.2.
 - Correctly refuse to send header blocks with connection-specific headers,
   per RFC 7540 Section 8.1.2.2.
-
-Bugfixes (Breaking)
-~~~~~~~~~~~~~~~~~~~
-
 - Correctly refuse to send header blocks that contain duplicate pseudo-header
   fields, or with pseudo-header fields that appear after ordinary header fields,
   per RFC 7540 Section 8.1.2.1.
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -41,9 +41,21 @@ Bugfixes
 - Correctly strip leading/trailing whitespace from header field names and
   values.
 - Correctly refuse to send header blocks with a TE header whose value is not
-  ``trailers``, with connection-specific headers, with duplicate pseudo-header
-  fields, or with pseudo-header fields that appear after an ordinary header
-  field.
+  ``trailers``, per RFC 7540 Section 8.1.2.2.
+- Correctly refuse to send header blocks with connection-specific headers,
+  per RFC 7540 Section 8.1.2.2.
+
+Bugfixes (Breaking)
+~~~~~~~~~~~~~~~~~~~
+
+- Correctly refuse to send header blocks that contain duplicate pseudo-header
+  fields, or with pseudo-header fields that appear after ordinary header fields,
+  per RFC 7540 Section 8.1.2.1.
+
+  This may cause passing a dictionary as the header block to ``send_headers``
+  to throw a ``ProtocolError``, because dictionaries are unordered and so they
+  may trip this check.  Passing dictionaries here is deprecated, and callers
+  should change to using a sequence of 2-tuples as their header blocks.
 
 2.4.0 (2016-07-01)
 ------------------
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -40,6 +40,10 @@ Bugfixes
   or trailing whitespace.
 - Correctly strip leading/trailing whitespace from header field names and
   values.
+- Correctly refuse to send header blocks with a TE header whose value is not
+  ``trailers``, with connection-specific headers, with duplicate pseudo-header
+  fields, or with pseudo-header fields that appear after an ordinary header
+  field.
 
 2.4.0 (2016-07-01)
 ------------------
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -291,7 +291,7 @@ def _custom_startswith(test_string, bytes_prefix, unicode_prefix):
 def _reject_pseudo_header_fields(headers, hdr_validation_flags):
     """
     Raises a ProtocolError if duplicate pseudo-header fields are found in a
-    header block or if a pseudo-header field arrives in a block after an
+    header block or if a pseudo-header field appears in a block after an
     ordinary header field.
     """
     seen_pseudo_header_fields = set()
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -18,13 +18,13 @@ UPPER_RE = re.compile(b"[A-Z]")
 
 # A set of headers that are hop-by-hop or connection-specific and thus
 # forbidden in HTTP/2. This list comes from RFC 7540 § 8.1.2.2.
-CONNECTION_HEADERS = {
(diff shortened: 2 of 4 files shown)


h2/utilities.py
```diff
@@ -221,8 +221,8 @@ def _reject_te(headers, hdr_validation_flags):
     its value is anything other than "trailers".
     """
     for header in headers:
-        if header[0] == b'te':
-            if header[1].lower().strip() != b'trailers':
+        if header[0] in (b'te', u'te'):
+            if header[1].lower().strip() not in (b'trailers', u'trailers'):
```
Lukasa: Longer term we can think about removing this `strip` call by adding a normalization step that strips header field names and values. Worth considering.

Lukasa:
Generally speaking this looks good @alexwlchan, thanks! Addressing some of your specific questions:

> Slightly complicated by the Unicode issue, which makes this pipeline a tad slower throughout. Is this is a sensible way to handle it?

Yeah, the overhead is fairly minimal.

> Doesn’t have to be done here, but any suggestions?

Easiest option is to write a function that does the right thing. =) You could write a `_custom_startswith(test_string, bytes_prefix, unicode_prefix)` function that allows both kinds of string.

alexwlchan:
And now we reject bad pseudo-header fields as well. Along with another nail in the coffin of sending-dicts-as-headers, because anybody who does so now misses out on this lovely new header validation. How sad.

HISTORY.rst
```diff
@@ -27,6 +27,10 @@ Bugfixes
 - Correctly refuse to send header blocks that contain neither :authority nor
   Host headers, or header blocks which contain mismatched :authority and Host
   headers, per RFC 7540 Section 8.1.2.3.
+- Correctly refuse to send header blocks with a TE header whose value is not
+  ``trailers``, with connection-specific headers, with duplicate pseudo-header
+  fields, or with pseudo-header fields that appear after an ordinary header
+  field.
```
Lukasa: Can we split this into multiple entries please?

h2/utilities.py
Lukasa: Hrrrm.

I'm now wondering if, instead of longer term, we do this *shorter* term. That saves us obtaining the extra tech debt.

Feel like adding this as a normalization step?

test/test_basic_logic.py
```diff
@@ -439,8 +439,16 @@ def test_dict_headers(self):
...
         Sending headers using dict is deprecated but still valid.
         """
+        # One of the steps in the outbound header validation logic checks
+        # that we don't send a pseudo-header after we've sent a regular
+        # header.  We can't guarantee that if you send headers as a dict,
+        # because dicts are unordered.
+        config = h2.config.H2Configuration(
+            validate_outbound_headers=False
+        )
+
```
Lukasa: Given that `validate_outbound_headers` is currently defaulting to `True`, I think this represents a backward compatibility violation: dict-headers that previously "worked" now throw exceptions.

Now, I *think* that's acceptable here because we're tightening bugs, but it's worth us using some language about this in the changelog if you can.

Lukasa:
Cool, we're really close! A few small notes. =)

h2/utilities.py
alexwlchan: Isn’t that the change being added in #281?

We strip extra whitespace as a normalisation step sending, and normalisation precedes validation, so we don’t need the `strip()` call. <sup>1</sup>

And we’ll reject header fields that have bad whitespace, so even if the missing `strip()` allows something to slip through here, we’ll catch it earlier/later.

<sup>
<sup>1</sup> Somebody who disables normalisation but not validation could get bitten here and allowed to send invalid headers, but that seems like a weird set of config to use, and we trust them to behave sensibly.
</sup>

test/test_basic_logic.py
alexwlchan: Yes, I thought carefully about this, and I’m in two minds about whether to actually put this in now, or wait for 3.0.

Other checks in this block assert that you include :authority or Host headers, and if you choose the former then you’re stuffed: depending on your Python, this could randomly explode on you. Which feels less like deprecation and more like deletion.

I’ve put in a changelog entry for it, but I’m tempted to remove it now and revisit it for 3.0.

What do you think?

alexwlchan:
Changeling markups made.

h2/utilities.py
Lukasa: It is, yeah, so I think we can merge #281 *first* and then remove the `strip` call here.

test/test_basic_logic.py
Lukasa: I think I'm ok with doing this prior to 3.0. As with all the validation checks, it can be disabled, and issuing ill-formed header blocks is a bug that should be alerted.

HISTORY.rst
```diff
@@ -27,6 +27,22 @@ Bugfixes
 - Correctly refuse to send header blocks that contain neither :authority nor
   Host headers, or header blocks which contain mismatched :authority and Host
   headers, per RFC 7540 Section 8.1.2.3.
+- Correctly refuse to send header blocks with a TE header whose value is not
+  ``trailers``, per RFC 7540 Section 8.1.2.2.
+- Correctly refuse to send header blocks with connection-specific headers,
+  per RFC 7540 Section 8.1.2.2.
+
+Bugfixes (Breaking)
+~~~~~~~~~~~~~~~~~~~
```
Lukasa: I don't think we need a separate section for this, but I'll clean that up myself rather than block merge on it.

h2/utilities.py
Lukasa: Ah, that's right, I remember that this is why I haven't merged this yet.

Lukasa: Ok, @alexwlchan, we've merged #281, so this should be safe to remove now.

alexwlchan:
@Lukasa Removed the unnecessary `strip()` call, and rebased against master (just some conflicts in `HISTORY.rst`.

HISTORY.rst
Lukasa: Heh, actually, would you like to remove this section? Some of these bug fixes are breaking anyway.

Lukasa:
One tiny note about the changelog and I'm ready to merge!

alexwlchan:
Changelog updates made.

Lukasa:
Awesome, let's land this now. 2.5.0 is shaping up to be a pretty great release!

Issue #263, opened 2016-07-21
Does hyper reject trailers that contain pseudo-header fields?


alexwlchan:
Spotted in [RFC 7540 section 8.1.2.1](https://tools.ietf.org/html/rfc7540#section-8.1.2):

> Pseudo-header fields MUST NOT appear in trailers.

Does hyper enforce this rule? Should it?

I’d expect this rule to be enforced in the `validate_headers()` pipeline, but that’s only just learnt how to distinguish between trailers and non-trailers, so I don’t think it would have been able to enforce this rule.

Lukasa:
That's easily checked: write a test, see what happens!

nateprewitt:
It looks like this may have been patched in #289?

Lukasa:
I don't know that anyone actually wrote a test for this. @nateprewitt, want to contribute a test for this specific case so that we can really consider this closed?

nateprewitt:
Sure, I'd be happy to give it a go.

alexwlchan:
I’m not sure #289 does address this. At least, I can’t see it at a quick glance of the code, and I don’t remember writing anything to explicitly cover this case (test or otherwise).

Lukasa:
Resolved by #305.

alexwlchan:
@Lukasa Err, is this resolved or not?

Lukasa:
Yes. I have got exactly no idea why this exploded everywhere.

PR #305, opened 2016-08-31
check for pseudo header in trailer

nateprewitt:
So, I'm quickly realizing that I'm way out of my element here with HTTP2, so please excuse my naivety.

From the quick reading I did, I think this workflow *should* fail with a `ProtocolError`, but currently doesn't. I'm not sure how state works with the connection object here, so it's possible I'm not using the frames correctly.

@Lukasa or @alexwlchan, could I get some confirmation that this is along the right path?

### DIFF
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -322,6 +322,13 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
 
         yield header
 
+    # Pseudo-header fields MUST NOT appear in trailers - RFC 7540 § 8.1.2.1
+    if hdr_validation_flags.is_trailer and seen_pseudo_header_fields:
+        raise ProtocolError(
+            "Received pseudo-header in trailer %s" %
+            seen_pseudo_header_fields
+        )
+
 
 def _validate_host_authority_header(headers):
     """
--- a/test/test_complex_logic.py
+++ b/test/test_complex_logic.py
@@ -90,9 +90,9 @@ class TestComplexClient(object):
             assert c.open_outbound_streams == expected_outbound_streams
 
             # Pushed streams can only be closed remotely.
-            f = frame_factory.build_headers_frame(
+            f = frame_factory.build_data_frame(
                 stream_id=stream_id+1,
-                headers=self.example_response_headers,
+                data=b'the content',
                 flags=['END_STREAM'],
             )
             c.receive_data(f.serialize())
--- a/test/test_invalid_headers.py
+++ b/test/test_invalid_headers.py
@@ -113,6 +113,37 @@ class TestInvalidFrameSequences(object):
         request_event = events[0]
         assert request_event.headers == headers
 
+    def test_pseudo_headers_rejected_in_trailer(self, frame_factory):
+        """
+        Ensure we reject pseudo headers included in trailers
+        """
+        trailers = [(':path', '/'), ('extra', 'value')]
+
+        c = h2.connection.H2Connection(client_side=False)
+        c.receive_data(frame_factory.preamble())
+        c.clear_outbound_data_buffer()
+
+        header_frame = frame_factory.build_headers_frame(
+            self.base_request_headers
+        )
+        trailer_frame = frame_factory.build_headers_frame(
+            trailers, flags=["END_STREAM"]
+        )
+        head = header_frame.serialize()
+        trailer = trailer_frame.serialize()
+
+        c.receive_data(head)
+        # Raise exception if pseudo header in trailer
+        with pytest.raises(h2.exceptions.ProtocolError) as e:
+            c.receive_data(trailer)
+        assert "pseudo-header in trailer" in str(e)
+
+        # Test appropriate response frame is generated
+        expected_frame = frame_factory.build_goaway_frame(
+            last_stream_id=1, error_code=h2.errors.ErrorCodes.PROTOCOL_ERROR
+        )
+        assert c.data_to_send() == expected_frame.serialize()
+
 
 class TestSendingInvalidFrameSequences(object):
     """
@@ -217,6 +248,14 @@ class TestFilter(object):
         ]
     ]
 
+    hdr_validation_no_trailers = [
+        h2.utilities.HeaderValidationFlags(is_client, is_trailer)
+        for is_client, is_trailer in [
+            (True, False),
+            (False, False)
+        ]
+    ]
+
     @pytest.mark.parametrize('validation_function', validation_functions)
     @pytest.mark.parametrize('hdr_validation_flags', hdr_validation_combos)
     @given(headers=HEADERS_STRATEGY)
@@ -241,7 +280,9 @@ class TestFilter(object):
             h2.utilities.validate_headers(headers, hdr_validation_flags)
 
     @pytest.mark.parametrize('validation_function', validation_functions)
-    @pytest.mark.parametrize('hdr_validation_flags', hdr_validation_combos)
+    @pytest.mark.parametrize(
+        'hdr_validation_flags', hdr_validation_no_trailers
+    )
     def test_matching_authority_host_headers(self,
                                              validation_function,
                                              hdr_validation_flags):


Lukasa:
Hrm, right now I can't see what is going wrong because I'm flying and have no access to a proper shell (I'm on a phone in an airport), and Travis is failing the build on the linter. I'll follow this up tomorrow.

nateprewitt:
Sorry, I was expecting the test to fail so I didn't check the reason. It's now failing by NOT raising the `ProtocolError`. I don't know if this is because the conditions I've created are insufficient to reach that code, or because the functionality isn't implemented yet. No rush here, have a good flight!

alexwlchan:
> I don't know if this is because the conditions I've created are insufficient to reach that code, or because the functionality isn't implemented yet.

I had a proper look, and it’s because the functionality isn’t implemented yet. So you’ll need to add it – if you look in `utilities.py`, there are a series of functions that do header block validation. You’ll need to find an appropriate place to perform this new check.

Looking at the test already written: looks fine to me (but I’m far from an HTTP/2 expert), but you should also check that we send the appropriate response (with an HTTP error code) back down the connection. If you look at `test_headers_event`, you’ll see we do this check *after* ensuring we hit the `ProtocolError`.

test/test_invalid_headers.py
```diff
@@ -113,6 +113,39 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
...
+            [(':path', '/'),
+             ('extra-header', 'values')]
+        )
+
+        c = h2.connection.H2Connection(client_side=False)
+        c.receive_data(frame_factory.preamble())
+        c.clear_outbound_data_buffer()
+
+        head_flag = ["END_HEADERS"]
+        stream_flag = ["END_STREAM"]
```
Lukasa: There's no reason for us to hoist these out, they can go straight into the relevant frame factory calls.

test/test_invalid_headers.py
```diff
@@ -113,6 +113,39 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
...
+        Ensure we reject pseudo headers included in trailers
+        """
+        headers = (
+            [h for h in self.base_request_headers if h[0] != ':path'] +
+            [('trailers', 'extra-header')]
+        )
+        trailers = (
+            [(':path', '/'),
+             ('extra-header', 'values')]
+        )
```
Lukasa: You don't need the external parentheses: Python does implicit line continuation in any kind of bracket.

test/test_invalid_headers.py
```diff
@@ -113,6 +113,39 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
         request_event = events[0]
         assert request_event.headers == headers
 
+    def test_pseudo_headers_rejected_in_trailer(self, frame_factory):
+        """
+        Ensure we reject pseudo headers included in trailers
+        """
+        headers = (
+            [h for h in self.base_request_headers if h[0] != ':path'] +
```
Lukasa: There's no reason to exclude `:path` here.

nateprewitt:
Thanks for clarifying that @alexwlchan!

@Lukasa, sorry about all the junk, I didn't intend to have this reviewed yet. I just wanted to clarify that the test was working as intended before extending.

I'll start working on getting that exception raise :)

nateprewitt:
So I've got a conditional in place in the `validate_headers` sub-checks, but most of the `TestFilter` tests are now failing because we're passing a `True` value in the `is_trailer` flag with frames containing pseudo-headers. Is this something I should look at cleaning up?

nateprewitt:
@Lukasa, it looks like this test in test_complex_logic.py may need some tweaks. It is currently sending two sets of headers, both with a `:status`, down a stream. When it [sends the second set of headers](https://github.com/python-hyper/hyper-h2/blob/master/test/test_complex_logic.py#L65) to tear down the inbound stream, the frame is flagged as being a trailer which causes the validate_header check to fail. This doesn't fail for the what looks to be a [very similar call](https://github.com/python-hyper/hyper-h2/blob/master/test/test_complex_logic.py#L56) on the outbound stream above. I'm not sure I understand why one is failing while the other isn't, other than different checks depending on "direction" of the stream.

Some possible solutions are sending a different header set (sans pseudo-headers), or just a data frame. This may need some more investigating if I'm missing something deeper though.

nateprewitt:
Are there different rules from inbound v.s. outbound header frames? [`validate_outbound_headers`](https://github.com/python-hyper/hyper-h2/blob/6e00a935ad169584a8509d39e10d6626bd1613b0/h2/utilities.py#L389) seems to only do a subset of `validate_headers`.

Lukasa:
A PUSH_PROMISE frame contains *request* headers, not response headers. Essentially, the server pushes a *request* and then sends its own response to it.

Lukasa:
> Are there different rules from inbound v.s. outbound header frames?

No, there aren't. However, some of the checks on inbound header frames don't apply to outbound headers because we normalize them first. There's no point checking for whitespace or uppercase header fields because we got rid of any violations already.

nateprewitt:
Ok cool, thanks @Lukasa.

Is that [second header frame](https://github.com/python-hyper/hyper-h2/blob/master/test/test_complex_logic.py#L65) supposed to be treated as a new response, or a trailer? I'll need to tweak it, or my validation check to get tests passing.

Lukasa:
There is only one HEADERS frame. =) It is, therefore, not a trailer.

nateprewitt:
Gah, sorry I'm being so sloppy with my links. [This](https://github.com/python-hyper/hyper-h2/blob/master/test/test_complex_logic.py#L98) is the second HeaderFrame we're sending down the same stream. This is being flagged as a trailer presently because my link in the comment above is the first set of headers sent.

test/test_invalid_headers.py
```diff
@@ -256,8 +284,12 @@ def test_matching_authority_host_headers(self,
             (b':method', b'GET'),
             (b'host', b'example.com'),
         ]
-        assert headers == h2.utilities.validate_headers(
-            headers, hdr_validation_flags)
+        if not hdr_validation_flags.is_trailer:
```
nateprewitt: This fixes the failure condition but raises a question of if we really want to use `hdr_validation_combos` with this test. This will never be a valid trailer header and we're now checking for trailers with pseudo headers elsewhere, so this seems a bit redundant.

If this is a special case that requires all the checks though, I like this over a blanket `try`/`except ProtocolError` catch since this should be the only condition we fail under.

nateprewitt:
Ok, after some further reading, I'm *fairly* confident the `build_headers_frame` issue I poorly explained earlier is in fact an unintended frame sequence. I've updated the code to send a data frame instead. Tests are passing now, and I think this should cover #263.

* I left the simplified test unsquashed, in case you prefer the more verbose test. I'll squash if it looks OK.
* I left a note on the conditional addition in `test_matching_authority_host_headers`.

test/test_complex_logic.py
```diff
@@ -90,9 +90,9 @@ def test_correctly_count_server_streams(self, frame_factory):
             assert c.open_outbound_streams == expected_outbound_streams
 
             # Pushed streams can only be closed remotely.
-            f = frame_factory.build_headers_frame(
+            f = frame_factory.build_data_frame(
                 stream_id=stream_id+1,
-                headers=self.example_response_headers,
+                data=b'the content',
                 flags=['END_STREAM'],
             )
```
Lukasa: Ok, seeing this change in place I now understand what you were getting at. Yes, there is nothing wrong with closing that stream using trailers, but the trailers that were being sent were not actually valid trailers. This change is better. =)

h2/utilities.py
```diff
@@ -317,6 +317,10 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
                     "Received custom pseudo-header field %s" % header[0]
                 )
 
+            if hdr_validation_flags.is_trailer:
+                raise ProtocolError(
+                    "Received pseudo-header in trailer %s" % header[0]
+                )
```
Lukasa: Rather than do this check for every pseudo-header we see, it would be better to do it once, at the end of the loop. `if hdr_validation_flags.is_trailer and seen_pseudo_header_fields`.

test/test_invalid_headers.py
```diff
@@ -113,6 +113,34 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
         request_event = events[0]
         assert request_event.headers == headers
 
+    def test_pseudo_headers_rejected_in_trailer(self, frame_factory):
+        """
+        Ensure we reject pseudo headers included in trailers
+        """
+        headers = self.base_request_headers
```
Lukasa: Is there any reason for this temporary variable name?

test/test_invalid_headers.py
```diff
@@ -113,6 +113,34 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
...
+        trailers = [(':path', '/'), ('extra', 'value')]
+
+        c = h2.connection.H2Connection(client_side=False)
+        c.receive_data(frame_factory.preamble())
+        c.clear_outbound_data_buffer()
+
+        header_frame = frame_factory.build_headers_frame(headers)
+        trailer_frame = frame_factory.build_headers_frame(trailers,
+                                                          flags=["END_STREAM"])
```
Lukasa: Let's follow the indentation style for this project, which is

```python
trailer_frame = frame_factory.build_headers_frame(
    trailers, flags=["END_STREAM"]
)
```

test/test_invalid_headers.py
Lukasa: I think you just want to add a new set of header validation combos that doesn't include the trailer set, and use that here.

test/test_invalid_headers.py
nateprewitt: I believe it was because 127 is too long with the full declaration and I felt like this looked a little funky without a param on the first line. I can implement the line wrap though.

```python
header_frame = frame_factory.build_headers_frame(
    self.base_request_headers)
```

Lukasa: Let's do the line wrap. Trailing parenthesis on a line by itself please! =D

nateprewitt:
Alright, I think things are tidied. Gonna call it a night but I'll finish off any final tweaks in the morning.

h2/utilities.py
```diff
@@ -322,6 +322,13 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
 
         yield header
 
+    # Pseudo-header fields MUST NOT appear in trailers - RFC 7540 § 8.1.2.1
+    if hdr_validation_flags.is_trailer and seen_pseudo_header_fields:
+        raise ProtocolError(
+            "Received pseudo-header in trailer %s" %
+            seen_pseudo_header_fields.pop()
```
Lukasa: Eh, let's just show them all. We can just show the set directly. =)

test/test_invalid_headers.py
```diff
@@ -113,6 +113,36 @@ def test_transfer_encoding_trailers_is_valid(self, frame_factory):
...
+        trailer_frame = frame_factory.build_headers_frame(
+            trailers, flags=["END_STREAM"]
+        )
+        head = header_frame.serialize()
+        trailer = trailer_frame.serialize()
+
+        c.receive_data(head)
+        # Raise exception if pseudo header in trailer
+        with pytest.raises(h2.exceptions.ProtocolError):
+            c.receive_data(trailer)
```
Lukasa: Let's add a quick confirmation that we're getting the *right* exception message here: this can save us from situations where this test stops failing for the reason we expect.

test/test_invalid_headers.py
```diff
@@ -217,6 +247,14 @@ class TestFilter(object):
         ]
     ]
 
+    hdr_validation_client_combos = [
```
Lukasa: Let's call this `hdr_validation_no_trailers`. =)

nateprewitt:
Voilà!

Lukasa:
Alright, fantastic, thanks so much @nateprewitt! Per our [one of the team](http://python-hyper.org/en/latest/one-of-the-team.html) policy, you've got an invite to the Python Hyper organisation in your inbox. Welcome on board!

Issue #290, opened 2016-08-19
Reject response header blocks that don’t include the `:status` field


alexwlchan:
Quoting from [RFC 7540 section 8.1.2.4](https://tools.ietf.org/html/rfc7540#section-8.1.2.4):

> For HTTP/2 responses, a single `":status"` pseudo-header field is
   defined that carries the HTTP status code field (see [RFC7231],
   Section 6).  This pseudo-header field MUST be included in all
   responses; otherwise, the response is malformed (Section 8.1.2.6).

This would mean adding another check in `utilities.py` to cover the possibility that this pseudo-header field is missing, and adding some new test cases in `test_invalid_headers.py`.

alexwlchan:
@Lukasa I’m not going to do this myself, because (1) I don’t have the time right now, and (2) I think this might be a good first patch for somebody else to take on.

But I did have a brief look at it, and I think we might want another field on `HeaderValidationFlags`. Specifically – is this header block being sent or received?

The sensible place (performance-wise) to do this check would be `_reject_pseudo_header_fields` – we’ve already built up a set of pseudo-header field names, so we can just check `b':status' in seen_pseudo_header_fields`, but I don’t think the current header validation flags give us enough information to know if this is a response or not.

alexwlchan:
(Buttons, how do they work?)

Lukasa:
Thanks for raising this up @alexwlchan! @python-hyper/contributors, if any of you feel like picking this up *or* have someone in mind who needs a good start in OSS, please assign the issue to yourself and tag the person you want to do the work. If not, I'll also go looking to find a new OSS contributor.

aidylewis:
I dont mind picking this up when I get back to work on Monday.

aidylewis:
Can I ask what is the difference between a header and a pseudo header?

aidylewis:
When ```b':status'```  is set as a header, it seems to be stripped out in ```validate_headers```.  And what need to go in the ```HeaderValidationFlags``` tuple?

Lukasa:
Header *validation* should never strip a header. It may reject a header block (raise `ProtocolError`), but never strip a header. Do you have a test case that shows this behaviour?


As to in `HeaderValidationFlags`, I think the flag we need is `"is_response_header"`, which is set to `True` when we're sending a response header block.

aidylewis:
[...]

PR #311, opened 2016-09-12
Reject response header blocks that don’t include the :status field

sunu:
Fix for #290

### DIFF
--- a/h2/events.py
+++ b/h2/events.py
@@ -146,6 +146,17 @@ class _HeadersSent(object):
     pass
 
 
+class _ResponseHeadersSent(_HeadersSent):
+    """
+    The _ResponseHeadersSent event is fired whenever response headers are sent
+    on a stream.
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    pass
+
+
 class _TrailersSent(_HeadersSent):
     """
     The _TrailersSent event is fired whenever trailers are sent on a
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -766,8 +766,7 @@ class H2Stream(object):
         events = self.state_machine.process_input(input_)
 
         hf = HeadersFrame(self.stream_id)
-        hdr_validation_flags = self._build_hdr_validation_flags(
-            events, is_response=not self.state_machine.client)
+        hdr_validation_flags = self._build_hdr_validation_flags(events)
         frames = self._build_headers_frames(
             headers, encoder, hf, hdr_validation_flags
         )
@@ -1032,7 +1031,7 @@ class H2Stream(object):
 
         return [], events
 
-    def _build_hdr_validation_flags(self, events, is_response=False):
+    def _build_hdr_validation_flags(self, events):
         """
         Constructs a set of header validation flags for use when normalizing
         and validating header blocks.
@@ -1041,6 +1040,9 @@ class H2Stream(object):
             is_trailer = isinstance(
                 events[0], (_TrailersSent, TrailersReceived)
             )
+            is_response_header = isinstance(
+                events[0], ResponseReceived
+            ) and not self.state_machine.client
         except IndexError:
             # Some state changes don't emit an internal event (for example,
             # sending a push promise).  We *always* emit an event for trailers,
@@ -1049,8 +1051,7 @@ class H2Stream(object):
             # TODO: Find any places where we don't emit anything, and emit
             # an internal event, so we can do away with this branch.
             is_trailer = False
-
-        is_response_header = is_response and not is_trailer
+            is_response_header = False
 
         return HeaderValidationFlags(
             is_client=self.state_machine.client,
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -1047,8 +1047,8 @@ class H2Stream(object):
             # Some state changes don't emit an internal event (for example,
             # sending a push promise).  We *always* emit an event for trailers,
             # so the absence of an event means this definitely isn't a trailer.
-            # Similarly, we also emit an event whenever reponse headers are
-            # sent or received. So absent of those events means this is not an
+            # Similarly, we also emit an event whenever response headers are
+            # sent or received. So absence of those events means this is not an
             # response header either.
             #
             # TODO: Find any places where we don't emit anything, and emit
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -19,7 +19,7 @@ from .events import (
     RequestReceived, ResponseReceived, DataReceived, WindowUpdated,
     StreamEnded, PushedStreamReceived, StreamReset, TrailersReceived,
     InformationalResponseReceived, AlternativeServiceAvailable,
-    _HeadersSent, _TrailersSent
+    _HeadersSent, _ResponseHeadersSent, _TrailersSent
 )
 from .exceptions import (
     ProtocolError, StreamClosedError, InvalidBodyLengthError
@@ -146,7 +146,7 @@ class H2StreamStateMachine(object):
             if self.client is True or self.client is None:
                 raise ProtocolError("Client cannot send responses.")
             self.headers_sent = True
-            event = _HeadersSent()
+            event = _ResponseHeadersSent()
         else:
             assert not self.trailers_sent
             self.trailers_sent = True
@@ -1041,12 +1041,15 @@ class H2Stream(object):
                 events[0], (_TrailersSent, TrailersReceived)
             )
             is_response_header = isinstance(
-                events[0], ResponseReceived
-            ) and not self.state_machine.client
+                events[0], (_ResponseHeadersSent, ResponseReceived)
+            )
         except IndexError:
             # Some state changes don't emit an internal event (for example,
             # sending a push promise).  We *always* emit an event for trailers,
             # so the absence of an event means this definitely isn't a trailer.
+            # Similarly, we also emit an event whenever reponse headers are
+            # sent or received. So absent of those events means this is not an
+            # response header either.
             #
             # TODO: Find any places where we don't emit anything, and emit
             # an internal event, so we can do away with this branch.
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -766,7 +766,8 @@ class H2Stream(object):
         events = self.state_machine.process_input(input_)
 
         hf = HeadersFrame(self.stream_id)
-        hdr_validation_flags = self._build_hdr_validation_flags(events)
+        hdr_validation_flags = self._build_hdr_validation_flags(
+            events, is_response=not self.state_machine.client)
         frames = self._build_headers_frames(
             headers, encoder, hf, hdr_validation_flags
         )
@@ -1031,7 +1032,7 @@ class H2Stream(object):
 
         return [], events
 
-    def _build_hdr_validation_flags(self, events):
+    def _build_hdr_validation_flags(self, events, is_response=False):
         """
         Constructs a set of header validation flags for use when normalizing
         and validating header blocks.
@@ -1049,9 +1050,12 @@ class H2Stream(object):
             # an internal event, so we can do away with this branch.
             is_trailer = False
 
+        is_response_header = is_response and not is_trailer
+
         return HeaderValidationFlags(
             is_client=self.state_machine.client,
-            is_trailer=is_trailer
+            is_trailer=is_trailer,
+            is_response_header=is_response_header
         )
 
     def _build_headers_frames(self,
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -332,14 +332,15 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
     # If ':status' pseudo-header is not there in a response header, reject it
     # Relevant RFC section: RFC 7540 § 8.1.2.4
     # https://tools.ietf.org/html/rfc7540#section-8.1.2.4
-    seen_status_field = (
-        b':status' in seen_pseudo_header_fields or
-        u':status' in seen_pseudo_header_fields
-    )
-    if hdr_validation_flags.is_response_header and not seen_status_field:
-        raise ProtocolError(
-            "Response header block does not have a :status header"
+    if hdr_validation_flags.is_response_header:
+        seen_status_field = (
+            b':status' in seen_pseudo_header_fields or
+            u':status' in seen_pseudo_header_fields
         )
+        if not seen_status_field:
+            raise ProtocolError(
+                "Response header block does not have a :status header"
+            )
 
 
 def _validate_host_authority_header(headers):
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -171,7 +171,7 @@ def authority_from_headers(headers):
 # should be applied to a given set of headers.
 HeaderValidationFlags = collections.namedtuple(
     'HeaderValidationFlags',
-    ['is_client', 'is_trailer']
+    ['is_client', 'is_trailer', 'is_response_header']
 )
 
 
@@ -328,6 +328,13 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
             "Received pseudo-header in trailer %s" %
             seen_pseudo_header_fields
         )
+
+    # If ':status' pseudo-header is not there in a response header, reject it
+    if (hdr_validation_flags.is_response_header and
+        not any(field in seen_pseudo_header_fields
+                for field in (b':status', u':status'))):
+        raise ProtocolError(
+            "Couldn't find :status pseudo-header field in response header")
 
 
 def _validate_host_authority_header(headers):
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -330,11 +330,14 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
         )
 
     # If ':status' pseudo-header is not there in a response header, reject it
-    if (hdr_validation_flags.is_response_header and
-        not any(field in seen_pseudo_header_fields
-                for field in (b':status', u':status'))):
+    if (
+        hdr_validation_flags.is_response_header and
+        b':status' not in seen_pseudo_header_fields and
+        u':status' not in seen_pseudo_header_fields
+    ):
         raise ProtocolError(
-            "Couldn't find :status pseudo-header field in response header")
+            "Response header block does not have a :status header"
+        )
 
 
 def _validate_host_authority_header(headers):
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -330,6 +330,8 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
         )
 
     # If ':status' pseudo-header is not there in a response header, reject it
+    # Relevant RFC section: RFC 7540 § 8.1.2.4
+    # https://tools.ietf.org/html/rfc7540#section-8.1.2.4
     seen_status_field = (
         b':status' in seen_pseudo_header_fields or
         u':status' in seen_pseudo_header_fields
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -330,11 +330,11 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
         )
 
     # If ':status' pseudo-header is not there in a response header, reject it
-    if (
-        hdr_validation_flags.is_response_header and
-        b':status' not in seen_pseudo_header_fields and
-        u':status' not in seen_pseudo_header_fields
-    ):
+    seen_status_field = (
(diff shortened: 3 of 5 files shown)


h2/stream.py
```diff
@@ -1049,9 +1050,12 @@ def _build_hdr_validation_flags(self, events):
             # an internal event, so we can do away with this branch.
             is_trailer = False
 
+        is_response_header = is_response and not is_trailer
```
Lukasa: `is_response` should be completely determinable by the events. Do we really need the extra parameter?

h2/utilities.py
```diff
@@ -329,6 +329,13 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
             seen_pseudo_header_fields
         )
 
+    # If ':status' pseudo-header is not there in a response header, reject it
+    if (hdr_validation_flags.is_response_header and
+        not any(field in seen_pseudo_header_fields
+                for field in (b':status', u':status'))):
```
Lukasa: Can we hoist this tuple up to the module scope? If we don't, CPython will recreate this tuple each time we enter the function, which is just unacceptable on a function that is called so frequently.

h2/utilities.py
```diff
@@ -329,6 +329,13 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
...
+        raise ProtocolError(
+            "Couldn't find :status pseudo-header field in response header")
```
Lukasa: This project's style for line breaks is to do this:

```python
raise ProtocolError(
    "Couldn't find :status pseudo-header field in response header"
)
```

Mind changing this to match?

test/test_invalid_headers.py
```diff
@@ -239,20 +241,29 @@ class TestFilter(object):
...
-        for is_client, is_trailer in [
-            (True, True),
-            (True, False),
-            (False, True),
-            (False, False)
-        ]
+        h2.utilities.HeaderValidationFlags(
+            is_client, is_trailer, is_response_header)
+        for is_client, is_trailer, is_response_header in (
+            itertools.product([True, False], repeat=3))
```
Lukasa: This entire block also needs its closing brackets moved to their own lines.

test/test_invalid_headers.py
```diff
@@ -239,20 +241,29 @@ class TestFilter(object):
...
     ]
 
     hdr_validation_no_trailers = [
-        h2.utilities.HeaderValidationFlags(is_client, is_trailer)
-        for is_client, is_trailer in [
-            (True, False),
-            (False, False)
+        h2.utilities.HeaderValidationFlags(
+            is_client, is_trailer, is_response_header)
```
Lukasa: This closing bracket also needs to be moved to its own line.

test/test_invalid_headers.py
```diff
@@ -239,20 +241,29 @@ class TestFilter(object):
...
+        for is_client, is_trailer, is_response_header in [
+            (True, False, False),
+            (False, False, False),
+        ]
+    ]
+
+    hdr_validation_response_headers = [
+        h2.utilities.HeaderValidationFlags(
+            is_client, is_trailer, is_response_header)
```
Lukasa: And this closing bracket.

test/test_invalid_headers.py
```diff
@@ -300,6 +311,13 @@ def test_matching_authority_host_headers(self,
         assert headers == h2.utilities.validate_headers(
             headers, hdr_validation_flags)
 
+    @pytest.mark.parametrize('hdr_validation_flags',
+                             hdr_validation_response_headers)
```
Lukasa: This test will presumably fail: some configurations of the header validation flags *allow* for no :status header field to be present because they are for non-response header blocks.

Lukasa:
Great start @sunu! So far this looks really good, I've left a few notes inline for you to look over.

h2/utilities.py
alexwlchan: My original thinking was to do something more like:

```python
if (b':status' not in seen_pseudo_header_fields) and (u':status' not in seen_pseudo_header_fields)
```

subject to line length. I suspect that inlining this way is more efficient than creating tuples or using `any()`, but that's just a gut feeling.

h2/utilities.py
alexwlchan: I'd suggest a slightly different wording, to match that used in `_validate_host_authority_header()`:

> Response header block does not have a :status header.

h2/utilities.py
Lukasa: It almost certainly is.

sunu: Haha. I did this first and then tried doing fancy things.

test/test_invalid_headers.py
sunu: I don't really understand your comment here. Can you clarify? I think by header validation flags you meant `hdr_validation_combos` variable, but I'm using `hdr_validataion_response_headers` which has the `is_response_header` bit set to true in all cases.

sunu:
Thanks for the review @Lukasa @alexwlchan. I've updated the PR addressing most of the comments.

test/test_invalid_headers.py
Lukasa: As sorry, I had missed that. =)

h2/stream.py
```diff
@@ -1040,6 +1040,9 @@ def _build_hdr_validation_flags(self, events):
             is_trailer = isinstance(
                 events[0], (_TrailersSent, TrailersReceived)
             )
+            is_response_header = isinstance(
+                events[0], ResponseReceived
+            ) and not self.state_machine.client
```
Lukasa: Do we need this `and not` check? I believe that `ResponseReceived` can only fire when we are a client.

While we're here, I suspect we're going to need an internal-only `_ResponseSent` event, much like the `_TrailersSent` event you can see in the block above.

h2/stream.py
```diff
@@ -1048,10 +1051,12 @@ def _build_hdr_validation_flags(self, events):
             # TODO: Find any places where we don't emit anything, and emit
             # an internal event, so we can do away with this branch.
             is_trailer = False
+            is_response_header = False
```
Lukasa: Can you extend the comment above to mention that we always emit an event for responses, too?

h2/utilities.py
```diff
@@ -329,6 +329,16 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
             seen_pseudo_header_fields
         )
 
+    # If ':status' pseudo-header is not there in a response header, reject it
+    if (
+        hdr_validation_flags.is_response_header and
+        b':status' not in seen_pseudo_header_fields and
+        u':status' not in seen_pseudo_header_fields
+    ):
```
Lukasa: Rather than have this gigantic if statement with confusing indentation, can we save the boolean value into a temporary variable and then `if` on that?

test/test_invalid_headers.py
```diff
@@ -239,20 +241,33 @@ class TestFilter(object):
...
-            (True, False),
-            (False, False)
+        h2.utilities.HeaderValidationFlags(
+            is_client, is_trailer, is_response_header
+        )
+        for is_client, is_trailer, is_response_header in [
+            (True, False, False),
+            (False, False, False),
+        ]
+    ]
```
Lukasa: I feel like `no_trailers` *does* include responses, so we need two more with `is_response_header = True`.

test/test_invalid_headers.py
```diff
@@ -239,20 +241,33 @@ class TestFilter(object):
...
+
+    hdr_validation_response_headers = [
+        h2.utilities.HeaderValidationFlags(
+            is_client, is_trailer, is_response_header
+        )
+        for is_client, is_trailer, is_response_header in [
+            (True, False, True),
+            (False, True, True),
+            (True, True, True),
+            (False, False, True)
```
Lukasa: Is it worth generating these two lists by using `filter`? Both `no_trailers` and `response_headers` are sub-cases of the `hdr_validation_combos`, so maybe we should stop hand-coding them and instead define their condition as a filter predicate.

That is: `hdr_validation_response_headers = list(filter(lambda flags: flags.is_response_header, hdr_validation_combos))`, and `hdr_validation_no_trailers = list(filter(lambda flags: not flags.is_trailer, hdr_validation_combos))`. That way, adding more flags to the header validation flags won't force us to keep updating this test.

Lukasa:
Great! I have left some more notes.

test/test_invalid_headers.py
sunu: Hey @Lukasa,
Quick question before updating the PR. It looks like we we use this to test if `:authority` and `host` match or not. As far as I understand, that test doesn't need to run on response headers. So we can safely ignore response headers here, can't we?

nateprewitt: @sunu I added this because there were certain combinations (specifically `is_trailer==True`) that were no longer considered valid after #305. I don't think we *need* to test `is_response_header` here since these test headers will never be a valid response, however, I don't think it hurts to test `is_response_header==False` either. This test class is specifically for broad filter checks, so the more coverage the better, imo.

nateprewitt: Sorry, I responded somewhat hastily here without reading @Lukasa's comment clearly. @Lukasa this has the same issue that I hit in #305, were testing on specifically request headers we end up with a lot of combinations that don't work. Part of me kind of feels like this test doesn't belong in the filter class, but I didn't know where to move it while working on #305.

We'll have to continually keep tweaking the hdr_validation_no_trailers list specifically for `test_matching_authority_host_headers` as we add more flags, which become fairly tedious as complexity grows. This specific test will never be a trailer, or a response. I'm also thinking it can probably only be a client too, but I'm not sure on that. Does it make sense for a server to send a request in HTTP/2?

Lukasa: We cannot safely ignore response headers: we still need to confirm that they don't explode. =)

sunu: I'm not sure I understand or whether we're talking about the same thing. So let me just push my changes. May be it'll be easier then.

h2/utilities.py
```diff
@@ -330,11 +330,9 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
         )
 
     # If ':status' pseudo-header is not there in a response header, reject it
-    if (
-        hdr_validation_flags.is_response_header and
-        b':status' not in seen_pseudo_header_fields and
-        u':status' not in seen_pseudo_header_fields
-    ):
+    seen_status_field = (b':status' in seen_pseudo_header_fields or
```
nateprewitt: @sunu I believe that the parens on their own lines is actually a style thing for this project that @Lukasa prefers.

h2/utilities.py
```diff
@@ -330,11 +330,9 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
...
+                         u':status' in seen_pseudo_header_fields)
```
nateprewitt: @sunu I believe that the parens on their own lines is actually a style thing for this project that @Lukasa prefers. You can see is working similarly on the conditional above.

h2/events.py
```diff
@@ -146,6 +146,17 @@ class _HeadersSent(object):
...
 
+class _ResponseHeadersSent(_HeadersSent):
+    """
+    The _ResponseHeadersSent event is fired whenever response headers are sent
+    on a stream.
+
+    This is an internal event, used to determine validation steps on
+    outgoing header blocks.
+    """
+    pass
```
Lukasa: I'd like at this point to also add a `_RequestHeadersSent` field. We don't have to do it as part of this PR, but can you open a new issue to track adding the new internal event? The lack of symmetry is making me sad. ;)

h2/utilities.py
sunu: Oops. Yeah, sorry about that. Will fix.

test/test_basic_logic.py
```diff
@@ -605,7 +605,7 @@ def test_reject_trailers_not_ending_stream(self, frame_factory):
         c = h2.connection.H2Connection()
         c.initiate_connection()
         c.send_headers(1, self.example_request_headers)
-        f = frame_factory.build_headers_frame(self.example_request_headers)
+        f = frame_factory.build_headers_frame(self.example_response_headers)
```
Lukasa: Heh, nice catch! That's a good validation that the new code is working as expected!

sunu:
@Lukasa I have pushed some commits. But the test `test_matching_authority_host_headers` will fail for all cases where `is_response_header` is true, because [`headers`](https://github.com/sunu/hyper-h2/blob/1074566adb07cf7a9055b8406e5a7beafc8d985c/test/test_invalid_headers.py#L294) doesn't contain `:status`.

So does it make sense to add `:status` to `headers` there? Does `:authority` header ever appear in a response?

Lukasa:
Response headers may never contain `:authority`.

test/test_invalid_headers.py
nateprewitt: So the [latest build](https://travis-ci.org/python-hyper/hyper-h2/jobs/160480020#L419) shows the issue a bit clearer I think. @Lukasa I know we want to be checking for the broadest set of options here in `TestFilter` and I think that's why this test may not belong in this specific class. It is hard coded to a set of headers that by definition cannot be a response according to the HTTP/2 spec (It contains request pseudo-headers). 

Since we're already testing that the failure happens in `test_response_header_without_status`, adding a `pytest.raises()` block for every combination that fundamentally cannot work for the given header set is a bit redundant.

I completely agree that we should be testing for all cases, I'm just afraid I may have accidentally started a trend of building unnecessary infrastructure (see: `hdr_validation_no_trailers`) around making a test pass, rather than fixing the underlying problem.

sunu:
> Awesome, one small comment change and then I think we're ready to go!

@Lukasa which comment change are you refering to? (I have already added a comment about ResponseHeadersSent event if that's the one you're refering to) 

Also, what about the failing test? :P

test/test_invalid_headers.py
Lukasa: Hrm. Are we able to filter out inconsistent combinations effectively? That is, can we run a filter that requires that we never have `is_response_header` and `is_client`?

nateprewitt: We can modify `hdr_validation_no_trailers` further since it was created **only** for use in `test_matching_authority_host_headers`. However, the only combination out of the whole matrix that will (or maybe should?) work for `test_matching_authority_host_headers` is: 
```python
(
  (is_client, is_trailer, is_response_header'), 
  (True, False, False)
)
```
So we lose any point of having this in the TestFilter class to begin with.

Lukasa: Ok, in that case let's just simplify that down. =)

sunu:
@Lukasa @nateprewitt Off topic but I get emails whenever you guys comment, but I can't find the last couple of comments on the Github ui anywhere (I tried for at least 5 minutes :P) Any idea what's going on?
Sorry for being a noob lol :P

Lukasa:
Heh, sorry, they're in an outdated diff. If you scroll up a little bit you'll find a dropdown saying "x commented on an outdated diff": you want the one with lots of comments in it. ;)

sunu:
@Lukasa PR updated

test/test_invalid_headers.py
sunu: In the latest change I'm considering both the cases for `is_client` true and false. Because we check for authority-host match on the server side too in https://github.com/python-hyper/hyper-h2/blob/94960a5324dac853ba256ffcc05ca2198df4b633/h2/utilities.py#L383 for example.

test/test_invalid_headers.py
```diff
@@ -282,11 +285,11 @@ def test_invalid_pseudo_headers(self, hdr_validation_flags):
...
     @pytest.mark.parametrize(
-        'hdr_validation_flags', hdr_validation_no_trailers
+        'hdr_validation_flags', hdr_validation_request_headers_no_trailer
     )
-    def test_matching_authority_host_headers(self,
-                                             validation_function,
-                                             hdr_validation_flags):
+    def test_matching_authority_host_headers(
+        self, validation_function, hdr_validation_flags
+    ):
```
Lukasa: Heh, sorry, I know that this isn't what you were told earlier, but the original style of this function was better. Basically, for function definitions I like having the argument names lined up, but for any other multiline sections I like having the parentheses by themselves.

sunu: Ok. fixed :)

h2/stream.py
```diff
@@ -1040,18 +1040,26 @@ def _build_hdr_validation_flags(self, events):
...
                 events[0], (_TrailersSent, TrailersReceived)
             )
+            is_response_header = isinstance(
+                events[0], (_ResponseHeadersSent, ResponseReceived)
+            )
         except IndexError:
             # Some state changes don't emit an internal event (for example,
             # sending a push promise).  We *always* emit an event for trailers,
             # so the absence of an event means this definitely isn't a trailer.
+            # Similarly, we also emit an event whenever reponse headers are
```
Lukasa: This should be `response`. =)

sunu: Fixed :)

h2/utilities.py
```diff
@@ -329,6 +329,16 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
             seen_pseudo_header_fields
         )
 
+    # If ':status' pseudo-header is not there in a response header, reject it
+    seen_status_field = (
```
alexwlchan: I wonder if it’s worth moving the evaluation of `seen_status_field` into a conditional block that only runs if this is a response header, i.e.,

```python
if hdr_validation_flags.is_response_header:
    seen_status_header = (...)
    if not seen_status_field:
        raise ProtocolError(...)
```

which saves us two unnecessary set lookups on request headers.

test/test_invalid_headers.py
```diff
@@ -239,22 +241,24 @@ class TestFilter(object):
...
-    hdr_validation_no_trailers = [
-        h2.utilities.HeaderValidationFlags(is_client, is_trailer)
-        for is_client, is_trailer in [
-            (True, False),
-            (False, False)
-        ]
-    ]
+    hdr_validation_response_headers = list(
+        filter(lambda flags: flags.is_response_header, hdr_validation_combos)
+    )
```
alexwlchan: (Opinion) I’d prefer this as a list comprehension, as I think that’s easier to see at a glance what the condition is, i.e.,

```python
hdr_validation_response_headers = [
    flags for flags in hdr_validation_combos
    if flags.is_response_header
]
```

and

```python
hdr_validation_response_headers = [
    flags for flags in hdr_validation_combos
    if not (flags.is_trailer or flags.is_response_header)
]
```

but if @Lukasa is happy with `list(filter(lambda(...)))`, I’ll shut up about it.

h2/utilities.py
```diff
@@ -329,6 +329,16 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
             seen_pseudo_header_fields
         )
 
+    # If ':status' pseudo-header is not there in a response header, reject it
```
alexwlchan: It would be nice if this comment had a reference to the exact RFC section – I know we‘ve not been super consistent on that point, but I think it’s helpful.

To save you looking it up again: `RFC 7540 § 8.1.2.4`.

https://tools.ietf.org/html/rfc7540#section-8.1.2.4

h2/stream.py
```diff
@@ -1040,18 +1040,26 @@ def _build_hdr_validation_flags(self, events):
             is_trailer = isinstance(
                 events[0], (_TrailersSent, TrailersReceived)
             )
+            is_response_header = isinstance(
+                events[0], (_ResponseHeadersSent, ResponseReceived)
```
alexwlchan: (Trivial) Seeing these side-by-side, I wonder if we should drop `Headers` from the name of this event, i.e., `_ResponseSent`. Better symmetry, and matches what I did for `_TrailerSent`.

test/test_invalid_headers.py
Lukasa: I have no problem with using a list comprehension.

h2/stream.py
Lukasa: Rather than deal with this in this ticket, let's open a new one related to this issue and #314 that refactors the event names.

Lukasa:
@sunu I'm aware that you're probably just taking a break, but as an FYI I'd like you to address @alexwlchan's reviews before I do another review. =)

h2/stream.py
alexwlchan: Added as a comment on #314.

sunu:
@Lukasa @alexwlchan PR updated!

Lukasa:
Can you please merge in the latest master branch? I'd like to confirm that the tests still pass with it in place.

sunu:
@Lukasa Done! :)

PR #315, opened 2016-09-19
Receiving a PUSH_PROMISE from a client is an error

alexwlchan:
This is explicitly forbidden in the spec (RFC 7540 § 8.2).

> A client cannot push.  Thus, servers MUST treat the receipt of a
>    PUSH_PROMISE frame as a connection error (Section 5.4.1) of type
>    PROTOCOL_ERROR.

In practice hyper-h2 was already raising a ProtocolError, because
local_settings.enable_push is False for servers.  This commit:
- Tweaks the error message to make it clear that receiving it from
  the client is important.
- Add an explicit test case.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1487,7 +1487,7 @@ class H2Connection(object):
         """
         Receive a push-promise frame on the connection.
         """
-        # A client cannot push. - RFC 7540 § 8.2
+        # A client cannot push - RFC 7540 § 8.2
         if not self.config.client_side:
             raise ProtocolError("Received pushed stream from a client")
 
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1487,6 +1487,10 @@ class H2Connection(object):
         """
         Receive a push-promise frame on the connection.
         """
+        # A client cannot push. - RFC 7540 § 8.2
+        if not self.config.client_side:
+            raise ProtocolError("Received pushed stream from a client")
+
         if not self.local_settings.enable_push:
             raise ProtocolError("Received pushed stream")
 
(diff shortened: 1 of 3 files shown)


test/test_basic_logic.py
```diff
@@ -1523,3 +1523,32 @@ def test_receiving_goaway_frame_with_unknown_error(self, frame_factory):
...
+        c.receive_data(f1.serialize())
+        c.clear_outbound_data_buffer()
+
+        with pytest.raises(h2.exceptions.ProtocolError):
+            c.receive_data(f2.serialize())
+
+        expected_frame = frame_factory.build_goaway_frame(
+            1, h2.errors.ErrorCodes.PROTOCOL_ERROR
+        )
+        assert c.data_to_send() == expected_frame.serialize()
```
Lukasa: Running this on my machine shows that this test would have passed before the code change above. Can you check the exception message, which is the substantive part of this change?

alexwlchan: The test down checks the exception message.

h2/connection.py
```diff
@@ -1487,6 +1487,10 @@ def _receive_push_promise_frame(self, frame):
         """
         Receive a push-promise frame on the connection.
         """
+        # A client cannot push - RFC 7540 § 8.2
+        if not self.config.client_side:
+            raise ProtocolError("Received pushed stream from a client")
+
         if not self.local_settings.enable_push:
             raise ProtocolError("Received pushed stream")
```
Lukasa: So, here's my question: is this worth it? I think, ordinarily, an invalid push would hit this code block here. That means that we'd still raise a `ProtocolError`. It also means that now, on push promise receipt, we have an extra `if` test that always fires in order to check for an error that we'll then immediately look for.

Should we consider whether it'd be better to put the new check into this `if` block? Or, even more aggressively, should we just implement #316? If we implement #316 then, definitionally, a server will be forbidden from setting `ENABLE_PUSH` to `1` (because we'll tell it not to), which means that all servers will automatically fall into the `if` block here.

alexwlchan:
Closing due to inactivity (from me).

PR #349, opened 2016-10-05
[requires.io] dependency update on 2.2.X branch

Lukasa:


alexwlchan:
Weird. This is the error we’re getting from Hypothesis:

> FailedHealthCheck: Data generation is extremely slow: Only produced 6 valid examples in 1.22 seconds (0 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.average_size or max_leaves parameters).

I can’t reproduce this failure locally. It worked for 3.5.2, and I don’t think there was a 3.5.3 change that would affect any of the bits of Hypothesis this test use.

This is the only test in that file that uses Hypothesis, so quite plausible we just got unlucky here. I’ll prod Travis to do another test run.

nateprewitt:
@alexwlchan I've hit this when committing before as well. This seems to happen when the travis instance gets backed up a bit, but a rerun has always fixed the issue.

alexwlchan:
Seems to have passed on #352. I’ll close this and somebody else can merge it via that PR.

nateprewitt:
Ok great, and just FYI it looks like you should be able to trigger a new build from Travis with the "Restart Job" button. This PR is running clean now.

alexwlchan:
Huh, I didn’t think I could trigger Travis runs on other people’s projects. Neat. =)

Lukasa:
@alexwlchan This isn't another person's project. You have write access to the repo, which grants you the privilege of re-running travis builds. ;)

PR #360, opened 2016-10-17
Documentation for negotiating HTTP/2.

### LINKED ISSUE #358: Add documentation with example code for plaintext upgrade.
For servers and clients.

Lukasa:
This change adds sample code for negotiating HTTP/2, both using ALPN and with plaintext. This resolves #358.

### DIFF
--- a/docs/source/negotiating-http2.rst
+++ b/docs/source/negotiating-http2.rst
@@ -14,6 +14,28 @@ This method is the simplest to use once the TLS connection is established. To us
 
 At this point, you're free to use all the HTTP/2 functionality provided by Hyper-h2.
 
+Server Setup Example
+~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


SethMichaelLarson:
@Lukasa Regarding the comment about Python 3.5+, Python 2.7.10+ has ALPN and NPN SSL/TLS support. Perhaps providing a check for `hasattr(ssl, "HAS_ALPN") and ssl.HAS_ALPN` (and it's NPN equivalent) would be better?

Lukasa:
So while that's reasonable, I'm reluctant to overly complicate the examples. Few people will use this code as written. It's mostly intended to explain the flow of logic. Does that seem sensible?

SethMichaelLarson:
Oh this is only for docs? Nevermind then, disregard me, I thought this was modifying functionality for plaintext upgrades within the actual code rather than just an example.

Lukasa:
Heh @SethMichaelLarson, nah, hyper-h2 doesn't know anything about the ssl module. =D

SethMichaelLarson:
@Lukasa Thankful for that haha

PR #363, opened 2016-10-19
Add support for automated window management.

### LINKED ISSUE #272: Add integrated window management.
Having had discussions with Tatsuhiro last week, I think it may be worthwhile to investigate hyper-h2's current policy on HTTP/2 window management.

Right now we expect users to manage their flow control window themselves, by calling `increment_flow_control_window`. This causes frames to be emitted directly. That allows users to carefully tailor their flow control management to their particular use-case, which is nice, but for 99% of users this isn't important. Instead, they'll fall back to the 'naive' flow control algorithm, which incurs quite a lot of overhead.

It would be nice, then, to add the ability to have a default flow control behaviour. To get this we need an API that allows users to report when they have consumed data, but not necessarily emit a frame. This could either piggyback on the current API (`increment_flow_control_window`), or it could use a new API: either could work.

We'll need a flag that allows enabling this behaviour. It'll default to `False` for the moment, for backward compatibility purposes, but for 3.0 we'll swap it to `True`: only users like mitmproxy actually require this to be turned off, so `True` would be the ideal default.

I suspect we can check the best algorithm by looking at nghttp2, but at a simple level we should avoid emitting per-stream WINDOW_UPDATE frames until the flow control window has been more than 50% consumed, at which point we should emit WINDOW_UPDATE frames for the entire chunk.

We also need to think about connection-level flow control. I don't have good thoughts here yet, I'll ponder more.

Lukasa:
Resolves #272.

This PR would add functionality for allowing hyper-h2 to make window management decisions for users. Rather than before, when users had to choose when to increment flow control windows and by how much, this change would allow users to simply inform hyper-h2 when they are done with data that has been received, and hyper-h2 will choose when to actually increment the flow control window.

This helps users who are implementing naive flow control algorithms, which may cause a barrage of WINDOW_UPDATE frames to be emitted, clogging up the network for no good reason.

Unfortunately, this turned into a fairly sizeable change. It adds quite a complex and subtle bit of function.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,8 @@ API Changes (Backward-Compatible)
 - Added a new ``H2Configuration`` object that allows rich configuration of
   a ``H2Connection``. This object supersedes the prior keyword arguments to the
   ``H2Connection`` object, which are now deprecated and will be removed in 3.0.
+- Added support for automated window management via the
+  ``acknowledge_received_data`` method. See the documentation for more details.
 - Added a ``DenialOfServiceError`` that is raised whenever a behaviour that
   looks like a DoS attempt is encountered: for example, an overly large
   decompressed header list. This is a subclass of ``ProtocolError``.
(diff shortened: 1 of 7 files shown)


Lukasa:
For those keeping track at home, I consider this enough of a marquee feature to want it to be the last thing in before we drop 2.5.0.

PR #366, opened 2016-10-21
Test Twisted as well as hyper-h2.

Lukasa:
Let's start taking our backward-compat guarantees seriously and try to avoid breaking our dependees.


Lukasa:
I should note that to achieve this goal I've centralised all our testing around tox. This is probably for the best.

.travis.yml
```diff
@@ -1,21 +1,36 @@
...
 
-python:
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-  - "pypy-5.3.1"
+cache:
+  directories:
+    - $HOME/.cache/pip
```
alexwlchan: While we’re caching stuff, let’s cache `.hypothesis` as well. That ensures the examples database is preserved over consecutive runs.

Lukasa: Ok, let's do that.

.travis/run.sh
```diff
@@ -1,11 +0,0 @@
-#!/bin/bash
-
-set -e
-set -x
-
-if [[ $TRAVIS_PYTHON_VERSION == pypy ]]; then
-    py.test --hypothesis-profile travis test/
```
alexwlchan: Have we lost the Hypothesis profile on Travis? That seems like it would be a regression &ndash; or is it sneaking in somewhere I can't see?

Lukasa: Arg, we have, though I did half of the work to try to plumb it through.

Lukasa: Fixed.

.travis.yml
```diff
@@ -1,21 +1,37 @@
...
+    # Main tests on supported Python versions.
+    - python: "2.7"
+      env: TOXENV=py27
+    - python: "3.3"
+      env: TOXENV=py33
+    - python: "3.4"
+      env: TOXENV=py34
+    - python: "3.5"
+      env: TOXENV=py35
+    - python: "pypy-5.3.1"
```
alexwlchan: Also, Travis appears to have PyPy 5.4 available, just not 5.4.1 (yet). Is that preferable to 5.3.1?

https://github.com/travis-ci/travis-ci/issues/6553

tox.ini
```diff
@@ -20,5 +35,5 @@ deps = sphinx==1.4.8
 whitelist_externals = make
 changedir = {toxinidir}/docs
 commands =
-    make clean
-    make html
+    rm -rf build
```
alexwlchan: Final question (I think): what was wrong with the make commands?

Lukasa: The `make` commands work great locally, but they add an implicit dependency on `make` which doesn't fly on Travis: that's why Travis previously had the `sphinx-build` command instead of `make`.

alexwlchan: Got it. Given that's the case, we should remove `whitelist_externals = make`.

Lukasa: Done.

PR #370, opened 2016-10-27
Add a no-op base class for events

alexwlchan:
Fixes #369

PR #372, opened 2016-10-28
Run the tests the way Twisted wants them run.

Lukasa:
Looks like one of Twisted's tests fails if run directly via `trial`.

I actually think this is a Twisted bug, but for the moment let's work around it here. I'll follow up with Twisted directly.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,7 +22,7 @@ deps =
     idna
     priority
     sphinx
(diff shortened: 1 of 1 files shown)


Lukasa:
This should unblock #370.

alexwlchan:
Third time lucky! Merging.

PR #381, opened 2016-11-04
Validate push promise header fields.

### LINKED ISSUE #318: hyper should validate the header fields in a PUSH_PROMISE frame
[RFC 7540 section 8.2.1](https://tools.ietf.org/html/rfc7540#section-8.2.1):

> The header fields in `PUSH_PROMISE` and any subsequent `CONTINUATION` frames MUST be a valid and complete set of request header fields (Section&nbsp;8.1.2.3).  The server MUST include a method in the `:method` pseudo-header field that is safe and cacheable.  If a client receives a `PUSH_PROMISE` that does not include a complete and valid set of header fields or the `:method` pseudo-header field identifies a method that is not safe, it MUST respond with a stream error (Section&nbsp;5.4.2) of type `PROTOCOL_ERROR`.

Breaking this down somewhat, I think this means:
- "valid and complete set of request header fields" – your header needs to have a :method, :scheme, :authority/Host and :path header.
- "safe and cacheable :method pseudo-header" – combining [RFC&nbsp;7231 §&nbsp;4.2.1](https://tools.ietf.org/html/rfc7231#section-4.2.1) and [RFC&nbsp;7231 §&nbsp;4.2.3](https://tools.ietf.org/html/rfc7231#section-4.2.3), that only allows GET and HEAD.

The first can be checked with an extra check at the end of the `_reject_pseudo_header_fields()` validation step; the second should probably be spun off into its own like `_reject_te()`.

Not sure (off the top of my head) whether we can distinguish a PUSH_PROMISE frame using the current set of `hdr_validation_flags`. If not, we’ll need to add the ability to do that as well.

Lukasa:
Resolves #318.

This seems to have been a smooth enough patch, which is a good thing! Suggests that our validation pipeline is pretty functional.

### DIFF
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -19,7 +19,7 @@ from .events import (
     RequestReceived, ResponseReceived, DataReceived, WindowUpdated,
     StreamEnded, PushedStreamReceived, StreamReset, TrailersReceived,
     InformationalResponseReceived, AlternativeServiceAvailable,
-    _ResponseSent, _RequestSent, _TrailersSent
+    _ResponseSent, _RequestSent, _TrailersSent, _PushedRequestSent
 )
 from .exceptions import (
     ProtocolError, StreamClosedError, InvalidBodyLengthError
@@ -248,7 +248,8 @@ class H2StreamStateMachine(object):
         if self.client is True:
             raise ProtocolError("Cannot push streams from client peers.")
 
-        return []
+        event = _PushedRequestSent()
+        return [event]
 
     def recv_push_promise(self, previous_state):
         """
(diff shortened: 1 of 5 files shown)


h2/stream.py
```diff
@@ -1071,30 +1074,26 @@ def _build_hdr_validation_flags(self, events):
...
-                events[0], (_ResponseSent, ResponseReceived)
+        is_trailer = isinstance(
+            events[0], (_TrailersSent, TrailersReceived)
+        )
+        is_response_header = isinstance(
+            events[0],
+            (
+                _ResponseSent,
+                ResponseReceived,
+                InformationalResponseReceived
```
alexwlchan: Seeing this makes me wonder: should `InformationalResponseReceived` be a subclass of `ResponseReceived`?

Lukasa: I *strongly* recommend we don't do that. The default logic for 1XX responses that user agents can't tolerate is to ignore them, and having them caught in an `isinstance()` check for `ResponseReceived` will cause them in most cases to accidentally be treated as "proper" responses.

alexwlchan: Sounds good, I thought it might be something like that. Otherwise LGTM.

Lukasa: If only there were some button you could hit where GitHub would turn your LGTM into a nice tick mark.... :wink:

Lukasa:
Thanks @alexwlchan! :sparkles:

PR #382, opened 2016-11-06
Remove connection specific headers

nateprewitt:
This should address the issue discussed in lukasa/hyper#291. I expanded the scope of this PR a bit after reading RFC7540 because it has pretty clear wording on what to do in this situation.

   HTTP/2 will need to remove any header fields nominated by the
   Connection header field, along with the Connection header field
   itself.  Such intermediaries SHOULD also remove other connection-
   specific header fields, such as Keep-Alive, Proxy-Connection,
   Transfer-Encoding, and Upgrade, even if they are not nominated by the
   Connection header field.

There may have been another discussion on why this portion wasn't previously implemented but I couldn't find it from a cursory search.

With this patch, `h2` will now strip all connection related fields from headers. While the RFC specifies this specifically for messages transformed from HTTP/1.x to HTTP/2, I didn't initially implement a check because I wanted to check first. Does this distinction matter in practice, or will a universal strip work?

### DIFF
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -441,6 +441,15 @@ def _strip_surrounding_whitespace(headers, hdr_validation_flags):
             yield (header[0].strip(), header[1].strip())
 
 
+def _strip_connection_headers(headers, hdr_validation_flags):
+    """
+    Strip any connection headers as per RFC7540 § 8.1.2.2.
+    """
+    for header in headers:
+        if header[0] not in CONNECTION_HEADERS:
+            yield header
+
+
 def _check_sent_host_authority_header(headers, hdr_validation_flags):
     """
     Raises an InvalidHeaderBlockError if we try to send a header block
@@ -469,6 +478,7 @@ def normalize_outbound_headers(headers, hdr_validation_flags):
     """
     headers = _lowercase_header_names(headers, hdr_validation_flags)
     headers = _strip_surrounding_whitespace(headers, hdr_validation_flags)
+    headers = _strip_connection_headers(headers, hdr_validation_flags)
     headers = _secure_headers(headers, hdr_validation_flags)
 
     return headers
--- a/test/test_invalid_headers.py
+++ b/test/test_invalid_headers.py
@@ -232,6 +232,7 @@ class TestSendingInvalidFrameSequences(object):
         base_request_headers + [('transfer-encoding', 'gzip')],
         base_request_headers + [('upgrade', 'super-protocol/1.1')]
     ]
+    all_header_blocks = invalid_header_blocks + strippable_header_blocks
 
     @pytest.mark.parametrize('headers', invalid_header_blocks)
     def test_headers_event(self, frame_factory, headers):
@@ -267,7 +268,7 @@ class TestSendingInvalidFrameSequences(object):
                 stream_id=1, promised_stream_id=2, request_headers=headers
             )
 
-    @pytest.mark.parametrize('headers', invalid_header_blocks)
+    @pytest.mark.parametrize('headers', all_header_blocks)
     def test_headers_event_skipping_validation(self, frame_factory, headers):
         """
         If we have ``validate_outbound_headers`` disabled, then all of these
@@ -284,7 +285,12 @@ class TestSendingInvalidFrameSequences(object):
         c.clear_outbound_data_buffer()
         c.send_headers(1, headers)
 
-    @pytest.mark.parametrize('headers', invalid_header_blocks)
+        # Ensure headers are still normalized.
+        norm_headers = h2.utilities.normalize_outbound_headers(headers, None)
+        f = frame_factory.build_headers_frame(norm_headers)
+        assert c.data_to_send() == f.serialize()
+
+    @pytest.mark.parametrize('headers', all_header_blocks)
     def test_push_promise_skipping_validation(self, frame_factory, headers):
         """
         If we have ``validate_outbound_headers`` disabled, then all of these
@@ -304,13 +310,21 @@ class TestSendingInvalidFrameSequences(object):
         )
         c.receive_data(header_frame.serialize())
 
+        # Create push promise frame with normalized headers.
+        frame_factory.refresh_encoder()
+        norm_headers = h2.utilities.normalize_outbound_headers(headers, None)
+        pp_frame = frame_factory.build_push_promise_frame(
+            stream_id=1, promised_stream_id=2, headers=norm_headers
+        )
+
         # Clear the data, then send a push promise.
         c.clear_outbound_data_buffer()
         c.push_stream(
             stream_id=1, promised_stream_id=2, request_headers=headers
         )
+        assert c.data_to_send() == pp_frame.serialize()
 
-    @pytest.mark.parametrize('headers', invalid_header_blocks)
+    @pytest.mark.parametrize('headers', all_header_blocks)
     def test_headers_event_skip_normalization(self, frame_factory, headers):
         """
         If we have ``normalize_outbound_headers`` disabled, then all of these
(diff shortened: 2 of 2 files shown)


h2/utilities.py
```diff
@@ -484,7 +480,7 @@ def validate_outbound_headers(headers, hdr_validation_flags):
     headers = _reject_te(
         headers, hdr_validation_flags
     )
-    headers = _reject_connection_header(
+    headers = _strip_connection_header(
```
alexwlchan: Because this is modifying the outbound headers rather than checking for correctness, I think it should be a step in `normalize_outbound_headers` instead.

Implicit in `validate_outbound_headers` is that you either get a `ProtocolError` or the headers returned unmodified.

h2/utilities.py
```diff
@@ -197,10 +197,10 @@ def validate_headers(headers, hdr_validation_flags):
     headers = _reject_surrounding_whitespace(
         headers, hdr_validation_flags
     )
-    headers = _reject_te(
+    headers = _strip_connection_header(
```
Lukasa: I suspect the answer should be that we should consider this a normalisation step on outbound headers only. We should continue to hold our peers to the same standard as RFC7540 requires, but we can make our users lives easier by stripping these headers for them. That means this should be a new normalisation step in `normalize_outbound_headers`, rather than an adjustment of this current filtering logic.

alexwlchan: I wonder if we should break this out into a separate `normalize_headers` pipeline, enabled by default, but which can then be disabled separately if necessary.

RFC § 8.1.2.2 also says:

> HTTP/2 does not use the Connection header field to indicate
   connection-specific header fields; in this protocol, connection-
   specific metadata is conveyed by other means.  An endpoint MUST NOT
   generate an HTTP/2 message containing connection-specific header
   fields; any message containing connection-specific header fields MUST
   be treated as malformed (Section 8.1.2.6).

and you only encounter them and remove them if you’re interworking from HTTP/1.x to HTTP/2. If I’m communicating with an endpoint that speaks pure HTTP/2, I might want to raise a ProtocolError if I get these bad headers.

Or maybe we should add a different configuration option "I expect to get messages that came from HTTP/1.x"? Less sure about that.

h2/utilities.py
nateprewitt: Yep, completely agree. I missed the normalization step when skimming.

h2/utilities.py
nateprewitt: I moved the stripping logic into `normalize_outbound_headers`, and this now works as anticipated. As for the separate pipeline, it looks like both the questions posed involve the receipt of invalid headers rather than sending. With this normalization only running for outbound headers, we will properly raise a ProtocolError if we receive a malformed message. 

> An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields;

I feel like stripping the headers satisfies this requirement but I can understand a situation where someone may want to catch the exception and possibly change the request workflow. If we do decide to go this route, how does a flag like `normalize_transformed_connection_headers` and a separate normalization function sound?

nateprewitt:
@alexwlchan @Lukasa, it looks like force pushing removed my last response and I can no longer respond to your comment in h2/utilities.py.

In short, I've moved everything over to `normalize_outbound_headers` which now allows h2 to raise a ProtocolError when bad headers are received. Both scenarios you've listed involve receiving rather than sending, so I believe this covers those cases.

> An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields;

In regards to a flag, I think stripping the headers satisfies the above requirement, but I can see a scenario where the user would rather catch the exception and modify the request flow. If this feature is needed, how does something like a `normalize_transformed_connection` flag with an additional function run prior to `normalize_outbound_headers` sound?

Lukasa:
>  If this feature is needed, how does something like a normalize_transformed_connection flag with an additional function run prior to normalize_outbound_headers sound?

It's not needed really: at this point I don't see any use-case for having extremely granular normalization flags. The normalization flag primarily exists for mitmproxy which wants to reproduce h2 requests/responses as directly as possible, replicating any errors the endpoints may have made. So they aren't going to need to turn only *part* of the normalization on or off.

h2/utilities.py
```diff
@@ -441,6 +441,16 @@ def _strip_surrounding_whitespace(headers, hdr_validation_flags):
             yield (header[0].strip(), header[1].strip())
 
 
+def _strip_connection_headers(headers, hdr_validation_flags):
+    """
+    Strip any connection headers as per RFC7540 § 8.1.2.2.
+    We ignore the 'te' header because it's handled in `_reject_te`.
+    """
+    for header in headers:
+        if header[0] not in CONNECTION_HEADERS or header[0] in (b'te', u'te'):
```
Lukasa: `te` isn't in `CONNECTION_HEADERS`, so I don't think we need this extra check.

test/test_invalid_headers.py
```diff
@@ -365,6 +367,27 @@ def test_push_promise_skip_normalization(self, frame_factory, headers):
...
+
+        # Clear the data, then try to send headers.
+        c.clear_outbound_data_buffer()
+        c.send_headers(1, headers)
+
+        conn_headers = (
+            'connection', 'proxy-connection', 'keep-alive',
+            'transfer-encoding', 'upgrade'
+        )
+        stripped_headers = [h for h in headers if h[0] not in conn_headers]
```
Lukasa: This replicates too much of the logic of the underlying code. In this case, we can say without loss of generality that the end result should be `stripped_headers == base_request_headers``, so let's use that as our assertion instead.

test/test_invalid_headers.py
```diff
@@ -365,6 +367,27 @@ def test_push_promise_skip_normalization(self, frame_factory, headers):
...
+
+        f = frame_factory.build_headers_frame(stripped_headers)
+        assert c.data_to_send() == f.serialize()
+
```
Lukasa: So we need more testing here.

In particular, we need to test that these headers are left in place (and therefore rejected) if `normalize_header_blocks` is not turned on. We also need to test that these header blocks fail on receipt unless `validate_inbound_headers` is turned off. Essentially, you need to go back to the following tests and have them also test `stripped_headers` to ensure that `stripped_headers` does the right thing:

- `TestSendingInvalidFrameSequences.test_headers_event_skipping_validation` (should emit the header block without the stripped field but otherwise left intact)
- `TestSendingInvalidFrameSequences.test_push_promise_skipping_validation` (should emit the header block without the stripped field but otherwise left intact)
- `TestSendingInvalidFrameSequences.test_headers_event_skip_normalization`
- `TestSendingInvalidFrameSequences.test_push_promise_skip_normalization`

h2/utilities.py
nateprewitt: Yep, someone needs to take away my keyboard 😬

test/test_invalid_headers.py
nateprewitt: I've combined `strippable_header_blocks` and `invalid_header_blocks` into an `all_header_blocks` variable which I'm now using for these tests. Since this normalization only runs in `normalize_outbound_headers` are we concerned about the `validate_inbound_headers` flag? `TestInvalidFrameSequences` should still cover the general ProxyErrors raised by receipt of these headers.

test/test_invalid_headers.py
```diff
@@ -282,7 +285,12 @@ def test_headers_event_skipping_validation(self, frame_factory, headers):
         c.clear_outbound_data_buffer()
         c.send_headers(1, headers)
 
-    @pytest.mark.parametrize('headers', invalid_header_blocks)
+        # Ensure headers are still normalized.
+        norm_headers = h2.utilities.normalize_outbound_headers(headers, None)
```
nateprewitt: None of the `normalize_outbound_headers` methods currently use `HeaderValidationFlags` but I also feel like passing `None` is asking for trouble later. I have a commit adding this ready to push, but it feels messy for what it's providing. Is there a better way to retrieve the `HeaderValidationFlags` for this given event, or do we need to build it from scratch?
```python
        hvf = h2.utilities.HeaderValidationFlags(
            is_client=True,
            is_trailer=False,
            is_response_header=False,
            is_push_promise=False
        )
        norm_headers = h2.utilities.normalize_outbound_headers(headers, hvf)
```

nateprewitt: Ok, that sounds good. I think I've got the tests addressed then.

Lukasa:
Naturally, I'd like @alexwlchan to confirm that his concerns are also addressed. =)

alexwlchan:
Nope, all looks good to me! 😄

PR #420, opened 2016-12-07
Add low-level documentation.

Lukasa:
This PR begins to add some low-level documentation, discussing implementation details. It also pulls some code in from @glyph's `automat` to render out the FSMs, and puts them into the documentation.


glyph:
This is cool!

But, is there anything I could do in order to get you to depend on (or vendor in) automat, rather than pulling out bits of its code?

Lukasa:
At this point, very little of automat's code is being used here. automat has a fundamentally very different view of its state machines than hyper-h2 does, which necessitated a fairly substantial rewrite of almost all of the code in automat's visualizer.

glyph:
Fair enough! Thanks for the shout-out and license acknowledgement, at any rate :)

Lukasa:
That's genuinely my pleasure: y'all did the hard bit of actually learning how to graphviz.

Issue #449, opened 2017-01-26
HTTP2-Settings of h2c upgrade should only contains the Settings payload


chhsiao90:
I found that when I'm doing h2c a client, and trying doing the upgrade by used```initiate_upgrade_connection``` to generate the base64 encoded HTTP2-Settings frame, it also contains unexpected frame information.

related code: [connection.py#initiate_upgrade_connection](https://github.com/python-hyper/hyper-h2/blob/master/h2/connection.py#L592-L597)
related rfc: [rfc7549#section-3.2](https://tools.ietf.org/html/rfc7540#section-3.2)
> GET / HTTP/1.1
     Host: server.example.com
     Connection: Upgrade, HTTP2-Settings
     Upgrade: h2c
     HTTP2-Settings: <base64url encoding of HTTP/2 SETTINGS payload>

If you confirmed that it's an rfc violation and had no time to fix it, I would like to help on a PR for this.
Thanks!

Lukasa:
Hrm, interesting. I hadn't noticed the subtlety in the word "payload": you may well be right. I'm going to check with the HTTPbis working group to confirm that that's the intent.

Lukasa:
Ok, so this definitely looks like a RFC violation. Good spot!

Lukasa:
Thanks so much for this work @chhsiao90! I've backported this fix to the various maintenance branches it affects, and I'll be aiming to ship releases for those branches sometime soon. Per our [one of the team policy](http://python-hyper.org/en/latest/one-of-the-team.html), you'll find an invitation to the Python-Hyper team in your inbox. Thanks so much!

chhsiao90:
@Lukasa - I had joined the team and thanks for the invitation!

PR #450, opened 2017-01-26
Only include SettingsFrame payload in upgrade request for h2c

chhsiao90:
Resolved: #449 

Only update the code for client side.
The server side seems didn't parse the HTTP2-Settings value at the latest master version.


Lukasa:
So, this is a good fix and I think we need to merge it. As an extra note, the server-side functionality here doesn't really work: we don't do anything with the provided data on the server side. I'm going to consider that a separate issue, but if you'd like to tackle it you're certainly welcome to.

PR #452, opened 2017-01-27
Resolve issue with plaintext upgrade on the server side

### LINKED ISSUE #451: Servers don't do anything with the HTTP2-Settings field.
Despite adding plaintext upgrade support in `2.3.0`, I managed to miss the fact that we don't actually *do* anything with the SETTINGS frame data sent by the remote peer. That's clearly pretty silly, so we should do something with it.

Lukasa:
Resolves #451.

### DIFF
--- a/test/test_h2_upgrade.py
+++ b/test/test_h2_upgrade.py
@@ -272,6 +272,13 @@ class TestServerUpgrade(object):
         server = h2.connection.H2Connection(client_side=False)
         client = h2.connection.H2Connection(client_side=True)
 
+        # As a precaution, let's confirm that the server and client, at the
+        # start of the connection, do not agree on their initial settings
+        # state.
+        assert (
+            client.local_settings._settings != server.remote_settings._settings
+        )
+
         # Get the client header data and pass it to the server.
         header_data = client.initiate_upgrade_connection()
         server.initiate_upgrade_connection(header_data)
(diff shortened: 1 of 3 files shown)


test/test_h2_upgrade.py
```diff
@@ -263,3 +263,35 @@ def test_cannot_receive_data_stream_1(self, frame_factory):
...
+            server.local_settings
+        )
+        assert server.data_to_send() == expected_frame.serialize()
+
+        # We violate abstraction layers here, but I don't think defining __eq__
+        # for this is worth it. In this case, both the client and server should
+        # agree that these settings have been ACK'd, so their underlying
+        # dictionaries should be identical.
+        assert (
+            client.local_settings._settings == server.remote_settings._settings
```
alexwlchan: This tells us that after we’ve exchanged some frames, the server/client settings match – but do we know they weren’t the same to start with? Should there be a second assert at the start of the test:

```python
assert (
    client.local_settings._settings != server.remote_settings._settings
)
```

so we know sending the settings to the server really did have an effect on its settings?

Lukasa: Yeah, good call.

Lukasa: Done. =D

alexwlchan: I’m slightly annoyed that if the new precondition fails, we’ll get a failure (bug in product code) rather than an error (bug in test setup).

But AFAIK, the only way pytest can throw an error is inside a fixture, and it’s not worth breaking this into a separate fixture for what’s a pretty trivial distinction.

Lukasa: I think there is an alternative, which is to define `__eq__` for the settings class. There is probably some value in that, but I shied away from it because I was concerned about adding all those extra tests in this PR.

PR #455, opened 2017-02-02
Reorder CONNECTION_HEADERS entries

sam-bristow:
When I first looked at this section of code I was a bit confused about why some of the entries were bytes while others were unicode strings.

By keeping the byte string and unicode versions of the headers together the relationship between the two is more obvious to a casual reader. This also matches the layout of the other `frozenset`s in this file.


Lukasa:
Thanks for the patch @sam-bristow! Per our [one of the team policy](http://python-hyper.org/en/latest/one-of-the-team.html), you'll find an invite to the python-hyper organisation in your inbox. Welcome to the team!

Thanks so much for the review + merge @alexwlchan. :sparkles:

PR #463, opened 2017-02-09
Update asyncio-server example to let it could be test with h2spec

chhsiao90:
I had updated ```examples/asyncio/asyncio-server.py```
To let it handle ProtocolError and ConnectionTerminated event,
so that it could be test with [h2spec](https://github.com/summerwind/h2spec).

Currently the execution result of h2spec is
```145 tests, 121 passed, 0 skipped, 24 failed```

### DIFF
--- a/examples/asyncio/asyncio-server.py
+++ b/examples/asyncio/asyncio-server.py
@@ -21,8 +21,11 @@ import collections
 from typing import List, Tuple
 
 from h2.connection import H2Connection
-from h2.events import DataReceived, RequestReceived, StreamEnded
+from h2.events import (
+    ConnectionTerminated, DataReceived, RequestReceived, StreamEnded
+)
 from h2.errors import ErrorCodes
+from h2.exceptions import ProtocolError
 
 
 RequestData = collections.namedtuple('RequestData', ['headers', 'data'])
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: examples/asyncio/asyncio-server.py [hunk 1]
```diff
@@ -40,18 +43,26 @@ def connection_made(self, transport: asyncio.Transport):
...
-                self.request_received(event.headers, event.stream_id)
-            elif isinstance(event, DataReceived):
-                self.receive_data(event.data, event.stream_id)
-            elif isinstance(event, StreamEnded):
-                self.stream_complete(event.stream_id)
-
+        try:
+            events = self.conn.receive_data(data)
+        except ProtocolError as e:
+            self.conn.close_connection(error_code=e.error_code)
```
Lukasa: This should not be necessary. A `ProtocolError` will already have written the GOAWAY frame, so you can just do the write/close below.

examples/asyncio/asyncio-server.py
```diff
@@ -40,18 +43,26 @@ def connection_made(self, transport: asyncio.Transport):
...
+            self.transport.write(self.conn.data_to_send())
+            self.transport.close()
+            raise
```
Lukasa: If you're going to re-raise, there's no need for the `else` underneath it. I recommend removing this `raise`.

chhsiao90: OK, I will remove it! Just want to see stack trace here :)

examples/asyncio/asyncio-server.py
chhsiao90: OK, thanks for the advise!

chhsiao90:
Had updated the code according to the review!

examples/asyncio/asyncio-server.py
Lukasa: That's fair enough. =)

PR #464, opened 2017-02-09
Implementation of equality for Settings class

bhavishyagopesh:
I have defined `__eq__` and `__ne__` for Settings class, and wrote tests for the same.

### DIFF
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -254,30 +254,37 @@ class Settings(collections.MutableMapping):
         return len(self._settings)
 
 
-    def _validate_setting(setting, value):
-        
-
-        if setting == ENABLE_PUSH:
-            if value not in (0, 1):
-                return ErrorCodes.PROTOCOL_ERROR
-        elif setting == INITIAL_WINDOW_SIZE:
-            if not 0 <= value <= 2147483647:  # 2^31 - 1
-                return ErrorCodes.FLOW_CONTROL_ERROR
-        elif setting == MAX_FRAME_SIZE:
-            if not 16384 <= value <= 16777215:  # 2^14 and 2^24 - 1
-                return ErrorCodes.PROTOCOL_ERROR
-        elif  setting == MAX_HEADER_LIST_SIZE:
-            if value < 0:
-                return ErrorCodes.PROTOCOL_ERROR
-    
-        return 0
-
-
     def __eq__(self,other):
         return self._settings==other._settings
 
     def __ne__(self,other):
         return not self._settings==other._settings
+
+
+def _validate_setting(setting, value):
+    """
+    Confirms that a specific setting has a well-formed value. If the setting is
+    invalid, returns an error code. Otherwise, returns 0 (NO_ERROR).
+    """
+    if setting == ENABLE_PUSH:
+        if value not in (0, 1):
+            return ErrorCodes.PROTOCOL_ERROR
+    elif setting == INITIAL_WINDOW_SIZE:
+        if not 0 <= value <= 2147483647:  # 2^31 - 1
+            return ErrorCodes.FLOW_CONTROL_ERROR
+    elif setting == MAX_FRAME_SIZE:
+        if not 16384 <= value <= 16777215:  # 2^14 and 2^24 - 1
+            return ErrorCodes.PROTOCOL_ERROR
+    elif setting == MAX_HEADER_LIST_SIZE:
+        if value < 0:
+            return ErrorCodes.PROTOCOL_ERROR
+
+    return 0
+
+
+
+
+    
 
 
 
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -253,18 +253,14 @@ class Settings(collections.MutableMapping):
     def __len__(self):
         return len(self._settings)
 
-
     def __eq__(self, other):
         if isinstance(other, Settings):
             return self._settings == other._settings
         else:
-            return NotImplemented
+            return False
             
     def __ne__(self, other):
-        if isinstance(other, Settings):
-            return self._settings != other._settings
-        else:
-            return NotImplemented
+        return not self == other
                 
 
 def _validate_setting(setting, value):
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -263,7 +263,8 @@ class Settings(collections.MutableMapping):
         if isinstance(other, Settings):
             return not self == other
         else:
-            return NotImplemented 
+            return NotImplemented
+
 
 def _validate_setting(setting, value):
     """
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -260,8 +260,10 @@ class Settings(collections.MutableMapping):
             return NotImplemented
 
     def __ne__(self, other):
-        return not self == other
-
+        if isinstance(other, Settings):
+            return not self == other
+        else:
+            return NotImplemented 
 
 def _validate_setting(setting, value):
     """
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -262,7 +262,6 @@ class Settings(collections.MutableMapping):
     def __ne__(self, other):
         return not self == other
                 
-
 def _validate_setting(setting, value):
     """
     Confirms that a specific setting has a well-formed value. If the setting is
@@ -282,10 +281,3 @@ def _validate_setting(setting, value):
             return ErrorCodes.PROTOCOL_ERROR
 
     return 0
-
-
-
-
-
-
-    
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -254,22 +254,33 @@ class Settings(collections.MutableMapping):
         return len(self._settings)
 
 
-def _validate_setting(setting, value):
-    """
-    Confirms that a specific setting has a well-formed value. If the setting is
-    invalid, returns an error code. Otherwise, returns 0 (NO_ERROR).
-    """
-    if setting == ENABLE_PUSH:
-        if value not in (0, 1):
-            return ErrorCodes.PROTOCOL_ERROR
-    elif setting == INITIAL_WINDOW_SIZE:
-        if not 0 <= value <= 2147483647:  # 2^31 - 1
-            return ErrorCodes.FLOW_CONTROL_ERROR
-    elif setting == MAX_FRAME_SIZE:
-        if not 16384 <= value <= 16777215:  # 2^14 and 2^24 - 1
-            return ErrorCodes.PROTOCOL_ERROR
-    elif setting == MAX_HEADER_LIST_SIZE:
-        if value < 0:
-            return ErrorCodes.PROTOCOL_ERROR
+    def _validate_setting(setting, value):
+        
 
-    return 0
+        if setting == ENABLE_PUSH:
+            if value not in (0, 1):
+                return ErrorCodes.PROTOCOL_ERROR
+        elif setting == INITIAL_WINDOW_SIZE:
+            if not 0 <= value <= 2147483647:  # 2^31 - 1
+                return ErrorCodes.FLOW_CONTROL_ERROR
+        elif setting == MAX_FRAME_SIZE:
+            if not 16384 <= value <= 16777215:  # 2^14 and 2^24 - 1
+                return ErrorCodes.PROTOCOL_ERROR
+        elif  setting == MAX_HEADER_LIST_SIZE:
+            if value < 0:
+                return ErrorCodes.PROTOCOL_ERROR
+    
+        return 0
+
+
+    def __eq__(self,other):
+        return self._settings==other._settings
+
+    def __ne__(self,other):
+        return not self._settings==other._settings
+
+
+
+
+
+    
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -279,5 +279,5 @@ def _validate_setting(setting, value):
     elif setting == MAX_HEADER_LIST_SIZE:
         if value < 0:
             return ErrorCodes.PROTOCOL_ERROR
-        
+
     return 0
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -254,12 +254,18 @@ class Settings(collections.MutableMapping):
         return len(self._settings)
 
 
-    def __eq__(self,other):
-        return self._settings==other._settings
-
-    def __ne__(self,other):
-        return not self._settings==other._settings
-
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self._settings == other._settings
+        else:
+            return NotImplemented
+            
+    def __ne__(self, other):
+        if isinstance(other, Settings):
+            return self._settings != other._settings
+        else:
+            return NotImplemented
+                
 
 def _validate_setting(setting, value):
     """
@@ -281,10 +287,6 @@ def _validate_setting(setting, value):
 
     return 0
 
-
-
-
-    
 
 
 
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -257,7 +257,7 @@ class Settings(collections.MutableMapping):
         if isinstance(other, Settings):
             return self._settings == other._settings
         else:
-            return False
+            return NotImplemented
             
     def __ne__(self, other):
         return not self == other
@@ -279,5 +279,4 @@ def _validate_setting(setting, value):
     elif setting == MAX_HEADER_LIST_SIZE:
         if value < 0:
             return ErrorCodes.PROTOCOL_ERROR
-
     return 0
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -279,4 +279,5 @@ def _validate_setting(setting, value):
     elif setting == MAX_HEADER_LIST_SIZE:
         if value < 0:
             return ErrorCodes.PROTOCOL_ERROR
+        
     return 0
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -258,10 +258,11 @@ class Settings(collections.MutableMapping):
             return self._settings == other._settings
         else:
             return NotImplemented
-            
+
     def __ne__(self, other):
         return not self == other
-                
+
+
 def _validate_setting(setting, value):
     """
     Confirms that a specific setting has a well-formed value. If the setting is
--- a/test/test_h2_upgrade.py
+++ b/test/test_h2_upgrade.py
@@ -138,7 +138,6 @@ class TestClientUpgrade(object):
         with pytest.raises(h2.exceptions.ProtocolError):
             c.send_data(stream_id=1, data=b'some data')
 
-
 class TestServerUpgrade(object):
     """
     Tests of the server-side of the HTTP/2 upgrade dance.
@@ -301,3 +300,4 @@ class TestServerUpgrade(object):
         assert (
             client.local_settings._settings == server.remote_settings._settings
         )
+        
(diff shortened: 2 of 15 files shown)


h2/settings.py
```diff
@@ -254,6 +254,13 @@ def __len__(self):
         return len(self._settings)
 
 
+    def __eq__(self,other):
+        return self._settings==other._settings
```
Lukasa: Style nit: we need a space either side of the operator.

I should also note that this does not behave correctly when compared with objects that are not `Settings` objects: specifically, it'll probably throw `AttributeError`s.

To deal with this, you need to confirm that the other object is an instance of `Settings` before doing the comparison. If it's not, you want to indicate that we don't know how to compare them by returning `NotImplemented`.

h2/settings.py
```diff
@@ -254,6 +254,13 @@ def __len__(self):
         return len(self._settings)
 
 
+    def __eq__(self,other):
```
Lukasa: Style nit: space after comma.

h2/settings.py
```diff
@@ -254,6 +254,13 @@ def __len__(self):
...
+        return self._settings==other._settings
+
+    def __ne__(self,other):
```
Lukasa: Style nit: space after comma.

h2/settings.py
```diff
@@ -254,6 +254,13 @@ def __len__(self):
...
+        return not self._settings==other._settings
```
Lukasa: So it's usually best to implement this in terms of equality unless there's a good reason not to, so I'd write this as `return not self == other`.

h2/settings.py
```diff
@@ -273,3 +280,14 @@ def _validate_setting(setting, value):
...
+
+
+
+    
+
+
+
+
+
+
```
Lukasa: We don't want to add all this whitespace.

test/test_h2_upgrade.py
```diff
@@ -301,4 +300,4 @@ def test_client_settings_are_applied(self, frame_factory):
         # dictionaries should be identical.
         assert (
             client.local_settings._settings == server.remote_settings._settings
-        )
+        )
```
Lukasa: You've made some changes in here that are only whitespace: please remove them.

test/test_settings.py
```diff
@@ -342,11 +342,22 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
             assert e.value.error_code == h2.errors.ErrorCodes.PROTOCOL_ERROR
             assert s.max_header_list_size is None
 
-            with pytest.raises(h2.exceptions.InvalidSettingsValueError) as e:
+            with pytest.raises(h2.exceptions.InvalidSettingsValueError) as e:
```
Lukasa: Please remove the trailing whitespace from here.

test/test_settings.py
```diff
@@ -342,11 +342,22 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
...
+
+    def equal_objects_are_equal(s1,s2):
+        s1.acknowledge()
+        s2.acknowledge()
+        assert s1._settings==s2._settings
+   
+    def notequal_objects_are_notequal(s1,s2):
+        s1.acknowledge()
+        s2.acknowledge()
+        assert s1._settings!=s2._settings
```
Lukasa: Ok, so these tests cannot possibly work right now for a number of reasons. Let's discuss why.

Firstly, and most importantly, where are `s1` and `s2` coming from? Nowhere else in the code knows how to provide them. If you look at tests higher up the file you'll see they create their own `Settings` objects: your tests need to do that as well.

Secondly, you need to think a bit harder about what these tests are actually testing. What you want to do is to confirm that the code you just wrote *actually works*. That means you want to test that two Settings entities that are the same are equal, and two that are different are not equal.

As an example, the simplest test that meaningfully tests something looks like this:

```python
def test_newly_created_settings_are_equal(self):
    s1 = h2.settings.Settings()
    s2 = h2.settings.Settings()

    assert s1 == s2
```

Note as well some basic rules of Python code: specifically, note that these test methods are *methods*, and so must take a `self` parameter, and note as well that to have them count as a test they need to have names that begin with`test_`.

h2/settings.py
SethMichaelLarson: Maybe even skip the middle-man and call `return not self.__eq__(other)` :)

Lukasa: Actually, you don't want to do that. ;)

The reason is that the rich equality tools in Python have some complex logic, and we don't want to privilege our own execution order. It's better to let Python work it out then to say "you must look at this specific method first".

h2/settings.py
```diff
@@ -254,12 +254,18 @@ def __len__(self):
...
-        return self._settings==other._settings
-
-    def __ne__(self,other):
-        return not self._settings==other._settings
-
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self._settings == other._settings
+        else:
+            return NotImplemented
```
SethMichaelLarson: If `other` isn't a `Settings` object we can pretty easily say we're not equal. Should return `False` in this case.

h2/settings.py
```diff
@@ -254,12 +254,18 @@ def __len__(self):
...
+            
+    def __ne__(self, other):
+        if isinstance(other, Settings):
+            return self._settings != other._settings
+        else:
+            return NotImplemented
```
SethMichaelLarson: Same as above, except `True`.

h2/settings.py
```diff
@@ -254,12 +254,18 @@ def __len__(self):
         return len(self._settings)
```
SethMichaelLarson: This whitespace should probably get removed. Only supposed to be 1 line between class methods.

SethMichaelLarson:
@bhavishyagopesh Okay, my suggestions are fulfilled, now go through @Lukasa's change requests.

h2/settings.py
```diff
@@ -253,7 +253,18 @@ def __iter__(self):
...
 
-
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self._settings == other._settings
+        else:
+            return False
+            
+    def __ne__(self, other):
+        if isinstance(other, Settings):
```
alexwlchan: I’d push for

```python
return not self == other
```

unless there’s a good reason not to, as otherwise we’re duplicating logic from the `__eq__` method.

test/test_h2_upgrade.py
```diff
@@ -302,3 +300,4 @@ def test_client_settings_are_applied(self, frame_factory):
         assert (
             client.local_settings._settings == server.remote_settings._settings
         )
+
```
alexwlchan: Please revert these unrelated whitespace changes.

test/test_settings.py
bhavishyagopesh: I have done most of the changes as you said but , tox is showing AssertionError with

"assert not s1 == s2"
or

"assert s1 != s2"
tox shows this error
assert <h2.settings.Settings object at 0x10a749410> != <h2.settings.Settings object at 0x10a749490>

h2/settings.py
alexwlchan: I’d go one step further and drop the `isinstance()` call – we’re still duplicating that. Simply:

```python
def __ne__(self, other):
    return not self == other
```

test/test_h2_upgrade.py
```diff
@@ -301,4 +299,4 @@ def test_client_settings_are_applied(self, frame_factory):
         # dictionaries should be identical.
         assert (
             client.local_settings._settings == server.remote_settings._settings
-        )
+        )
```
alexwlchan: We still have whitespace changes creeping into this file.

h2/settings.py
Lukasa: Sorry @SethMichaelLarson, but again I disagree. =) Per the documentation, [`NotImplemented`](https://docs.python.org/2/library/constants.html#NotImplemented) should be returned when we have no method of determining equality with respect to the other type. This is to allow *other* rich comparison methods to potentially be invoked that may know how to do an equality comparison.

The correct thing to do here is to return `NotImplemented` when we don't have matching types.

h2/settings.py
```diff
@@ -253,7 +253,18 @@ def __iter__(self):
     def __len__(self):
         return len(self._settings)
 
-
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self == other
```
Lukasa: So, this doesn't work because it leads to a recursive definition. Specifically, it says `self` is equal to `other` if `self` equals `other`. That's tautologically accurate, but doesn't help Python to actually work out what equality looks like.

This should be changed to specifically compare the `_settings` properties of the two objects.

h2/settings.py
Lukasa: Yup, completely agreed with @alexwlchan here.

bhavishyagopesh:
Thanks for your patience till now, I have made the changes requested....

h2/settings.py
```diff
@@ -253,7 +253,15 @@ def __iter__(self):
     def __len__(self):
         return len(self._settings)
 
-
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self._settings == other._settings
+        else:
+            return False
```
Lukasa: So this needs to return `NotImplemented`.

h2/settings.py
```diff
@@ -272,4 +280,4 @@ def _validate_setting(setting, value):
         if value < 0:
             return ErrorCodes.PROTOCOL_ERROR
 
-    return 0
+    return 0
```
Lukasa: There's still a whitespace change here.

test/test_h2_upgrade.py
Lukasa: There are still whitespace changes here that need removing.

test/test_settings.py
```diff
@@ -350,3 +350,120 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
...
+        
+
+    def anotherInstance(self):
+        """
+        Return an instance of the class under test.  Each call to this method
+        must return a different object.  The objects must not be equal to the
+        objects returned by C{anInstance}.  They may or may not be equal to
+        each other (they will not be compared against each other).
+        """
+        return h2.settings.Settings()
```
Lukasa: So this method must return a meaningfully different object than the one above it. The easiest way to do that is to use the `client` and `initial_values` arguments to the [`Settings` constructor](https://github.com/python-hyper/hyper-h2/blob/master/h2/settings.py#L100) to create a meaningfully different object.

test/test_settings.py
```diff
@@ -350,3 +350,120 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
 
             with pytest.raises(KeyError):
                 s[h2.settings.MAX_HEADER_LIST_SIZE]
+
+class EqualityTestsMixin(object):
```
Lukasa: Please call this class `TestSettingsEquality`. =)

test/test_settings.py
```diff
@@ -350,3 +350,120 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
...
+    """
+    A mixin defining tests for the standard implementation of C{==} and C{!=}.
```
Lukasa: Please replace the word `mixin` with the word `class`.

Lukasa: Please also replace all instances of `C{}` surrounding text: we don't use that style in this project.

test/test_settings.py
```diff
@@ -350,3 +350,120 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
...
+        The result of comparison using C{!=} is delegated to the right-hand
+        operand if it is of an unrelated type.
+        """
+        class Delegate(object):
+            def __ne__(self, other):
+                return [self]
+
+        a = self.anInstance()
+        b = Delegate()
+        assert (a != b) == [b]
```
Lukasa: This entire class has its method names in camel case (that is, with the bumpy bits). It shouldn't: it should be snake case (with underscores). That means you need to change each method name appropriately. For example, `anInstance` should become `an_instance`. `test_delegateNe` should become `test_delegate_ne`.

test/test_settings.py
bhavishyagopesh: What arguments should I pass, for client(it is true by default) and for initial_values?

Lukasa: For now, they only have to be different to what is passed in `anInstance`. Later on we'll discuss how we can programmatically test a wide range of options.

bhavishyagopesh:
So what's next?

SethMichaelLarson:
@Lukasa I sure hope you have the "Cancel Build if new Commit" option enabled for the `hyper` organizations Travis. Otherwise RIP build queues.

bhavishyagopesh:
ARE there still any whitespace errors?

Lukasa:
@bhavishyagopesh If you open the "files changed" tab you'll see them quite clearly. =)

bhavishyagopesh:
I have deleted that last line several times and then committed it too but it pops again....

h2/settings.py
```diff
@@ -271,5 +279,4 @@ def _validate_setting(setting, value):
     elif setting == MAX_HEADER_LIST_SIZE:
         if value < 0:
             return ErrorCodes.PROTOCOL_ERROR
-
```
Lukasa: This line needs to be put back.

test/test_h2_upgrade.py
```diff
@@ -2,7 +2,6 @@
 """
 test_h2_upgrade.py
 ~~~~~~~~~~~~~~~~~~
-
```
Lukasa: This line needs to be put back.

test/test_h2_upgrade.py
```diff
@@ -16,7 +15,6 @@
 import h2.events
 import h2.exceptions
 
-
```
Lukasa: This line needs to be put back.

test/test_h2_upgrade.py
```diff
@@ -139,7 +137,6 @@ def test_cannot_send_data_stream_1(self, frame_factory):
         with pytest.raises(h2.exceptions.ProtocolError):
             c.send_data(stream_id=1, data=b'some data')
 
-
```
Lukasa: This line needs to be put back.

test/test_h2_upgrade.py
```diff
@@ -301,4 +298,4 @@ def test_client_settings_are_applied(self, frame_factory):
         # dictionaries should be identical.
         assert (
             client.local_settings._settings == server.remote_settings._settings
-        )
+        )
```
Lukasa: Please delete this space.

bhavishyagopesh:
Now...

h2/settings.py
SethMichaelLarson: @Lukasa Huh! TIL That `NotImplemented` exists. I never knew of this construct. I will keep this in mind! Thanks! :)

h2/settings.py
Lukasa: This line should not have any spaces in it: it should be present but empty.

test/test_h2_upgrade.py
Lukasa: This line should not have any spaces in it: it should be present but empty.

test/test_h2_upgrade.py
Lukasa: You still need to delete this space.

SethMichaelLarson:
Gonna recommend a Squash and Merge to not destroy the poor Branch Network. Most of the commits are empty.

bhavishyagopesh:
@SethMichaelLarson Are things alright now?

SethMichaelLarson:
@bhavishyagopesh No. All tests need to pass on Travis, currently they are not passing. Missing test coverage on [lines 257-260 and 263](https://travis-ci.org/python-hyper/hyper-h2/jobs/200483807). You also have [linting errors](https://travis-ci.org/python-hyper/hyper-h2/jobs/200483813).

Also you should probably add a test where the settings object is actually used for it's purpose and then compared for equality.

bhavishyagopesh:
@SethMichaelLarson You mean to say I should give some value to `initial_values` argument....

SethMichaelLarson:
That would certainly work.

bhavishyagopesh:
@SethMichaelLarson I was asking for values of `initial_values` parameter.

SethMichaelLarson:
@bhavishyagopesh Look around the other tests within `test_settings.py` for some good ideas on values to use.

bhavishyagopesh:
@SethMichaelLarson I gave some values to `initial_values` but still it doesn't cover line 257-260 and 263.

bhavishyagopesh:
@Lukasa I guess, I have corrected linting errors, now what changes(additional tests to be added...) need to be made to cover line 257-260 and 263.

test/test_settings.py
```diff
@@ -350,3 +350,132 @@ def test_cannot_set_invalid_values_for_max_header_list_size(self, val):
 
             with pytest.raises(KeyError):
                 s[h2.settings.MAX_HEADER_LIST_SIZE]
+
+
+class test_settings_equality(object):
```
Lukasa: This class name should be `TestSettingsEquality`.

bhavishyagopesh:
Comparing two lists return `assertionerror` with __ne__ but not with __eq__ ?

bhavishyagopesh:
@SethMichaelLarson  Comparing two lists return assertionerror with ne but not with eq ?

bhavishyagopesh:
@Lukasa If you could cite some resource for what's really happening?Why it isn't passing?

bhavishyagopesh:
Is anyone there who would be kind enough to tell the error. Please......

h2/settings.py
```diff
@@ -253,6 +253,15 @@ def __iter__(self):
...
         return len(self._settings)
 
+    def __eq__(self, other):
+        if isinstance(other, Settings):
+            return self._settings == other._settings
+        else:
+            return NotImplemented
+
+    def __ne__(self, other):
+        return not self == other
```
Lukasa: Ok, this needs to be expanded a little bit. In particular, the `if isinstance` check needs to be repeated here so that we return `NotImplemented` if `other` is not a `Settings` object.

bhavishyagopesh:
@Lukasa Sorry for being impatient(but you are certainly AWESOME =).) So what's need's to be done next...

Lukasa:
Ok, so I'd call this "good enough" to get going with. I think I'd also like us to add some more complex testing based on Hypothesis, which is really perfect for testing this kind of thing. Would you like to tackle that too, or would you like a break?

bhavishyagopesh:
I think, I still need to learn a lot about python so,can this be merged first and I would love to continue doing whatever you suggest next....

Lukasa:
Ok cool, so we just need to wait until @alexwlchan is happy. =) @alexwlchan whenever you're happy, go ahead and merge.

alexwlchan:
Personally, I’d prefer to see the commits squashed/rebased before merging. 66 is a few too many for my tastes.

(I’m happy to do the rebase and/or be overruled on this one.)

Lukasa:
@alexwlchan Can you do a "rebase and merge" when you go ahead? We don't need the 66 commits.

alexwlchan:
Superseded by #465.

Lukasa:
Thanks so much @bhavishyagopesh! Per our [one of the team](http://python-hyper.org/en/latest/one-of-the-team.html) policy, you'll find a GitHub invitation to the Hyper organisation in your inbox.

PR #468, opened 2017-02-16
Allow sending data frames with padding

apolcyn:
This allows sending padding along with a data frame.

This defaults `pad_length` to `None`, indicating that padding flag should be unset - wanted to allow setting zero length padding with a zero-valued `pad_length`.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -833,7 +833,7 @@ class H2Connection(object):
 
         self._prepare_for_sending(frames)
 
-    def send_data(self, stream_id, data, end_stream=False):
+    def send_data(self, stream_id, data, end_stream=False, pad_length=None):
         """
         Send data on a given stream.
 
@@ -857,24 +857,44 @@ class H2Connection(object):
         :param end_stream: (optional) Whether this is the last data to be sent
             on the stream. Defaults to ``False``.
         :type end_stream: ``bool``
+        :param pad_length: (optional) Length of the padding to apply to the
+            data frame. Defaults to ``None`` for no use of padding. Note that
+            a value of ``0`` results in padding of length ``0``
+            (with the "padding" flag set on the frame).
+
+            .. versionadded:: 2.6.0
+
+        :type pad_length: ``int``
         :returns: Nothing
         """
-        if len(data) > self.local_flow_control_window(stream_id):
+        frame_size = len(data)
+        if pad_length is not None:
+            if not isinstance(pad_length, int):
+                raise TypeError("pad_length must be an int")
+            if pad_length < 0 or pad_length > 255:
+                raise ValueError("pad_length must be within range: [0, 255]")
+            # Account for padding bytes plus the 1-byte padding length field.
+            frame_size += pad_length + 1
+
+        if frame_size > self.local_flow_control_window(stream_id):
             raise FlowControlError(
                 "Cannot send %d bytes, flow control window is %d." %
-                (len(data), self.local_flow_control_window(stream_id))
+                (frame_size, self.local_flow_control_window(stream_id))
             )
-        elif len(data) > self.max_outbound_frame_size:
+        elif frame_size > self.max_outbound_frame_size:
             raise FrameTooLargeError(
                 "Cannot send frame size %d, max frame size is %d" %
-                (len(data), self.max_outbound_frame_size)
+                (frame_size, self.max_outbound_frame_size)
             )
 
         self.state_machine.process_input(ConnectionInputs.SEND_DATA)
-        frames = self.streams[stream_id].send_data(data, end_stream)
+        frames = self.streams[stream_id].send_data(
+            data, end_stream, pad_length=pad_length
+        )
+
         self._prepare_for_sending(frames)
 
-        self.outbound_flow_control_window -= len(data)
+        self.outbound_flow_control_window -= frame_size
         assert self.outbound_flow_control_window >= 0
 
     def end_stream(self, stream_id):
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -841,7 +841,7 @@ class H2Stream(object):
         assert not events
         return []
 
-    def send_data(self, data, end_stream=False):
+    def send_data(self, data, end_stream=False, pad_length=None):
         """
         Prepare some data frames. Optionally end the stream.
 
@@ -854,8 +854,12 @@ class H2Stream(object):
         if end_stream:
             self.state_machine.process_input(StreamInputs.SEND_END_STREAM)
             df.flags.add('END_STREAM')
+        if pad_length is not None:
+            df.flags.add('PADDED')
+            df.pad_length = pad_length
 
-        self.outbound_flow_control_window -= len(data)
+        # Subtract flow_controlled_length to account for possible padding
+        self.outbound_flow_control_window -= df.flow_controlled_length
         assert self.outbound_flow_control_window >= 0
 
         return [df]
--- a/test/test_basic_logic.py
+++ b/test/test_basic_logic.py
@@ -109,9 +109,116 @@ class TestBasicClient(object):
         c.clear_outbound_data_buffer()
         events = c.send_data(1, b'some data')
         assert not events
+        data_to_send = c.data_to_send()
         assert (
-            c.data_to_send() == b'\x00\x00\t\x00\x00\x00\x00\x00\x01some data'
+            data_to_send == b'\x00\x00\t\x00\x00\x00\x00\x00\x01some data'
         )
+
+        buffer = h2.frame_buffer.FrameBuffer(server=False)
+        buffer.max_frame_size = 65535
+        buffer.add_data(data_to_send)
+        data_frame = list(buffer)[0]
+        sanity_check_data_frame(
+            data_frame=data_frame,
+            expected_flow_controlled_length=len(b'some data'),
+            expect_padded_flag=False,
+            expected_data_frame_pad_length=0
+        )
+
+    def test_sending_data_with_padding(self):
+        """
+        Single data frames with padding are encoded correctly.
+        """
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        # Clear the data, then send some data.
+        c.clear_outbound_data_buffer()
+        events = c.send_data(1, b'some data', pad_length=5)
+        assert not events
+        data_to_send = c.data_to_send()
+        assert data_to_send == (
+            b'\x00\x00\x0f\x00\x08\x00\x00\x00\x01'
+            b'\x05some data\x00\x00\x00\x00\x00'
+        )
+
+        buffer = h2.frame_buffer.FrameBuffer(server=False)
+        buffer.max_frame_size = 65535
+        buffer.add_data(data_to_send)
+        data_frame = list(buffer)[0]
+        sanity_check_data_frame(
+            data_frame=data_frame,
+            expected_flow_controlled_length=len(b'some data') + 1 + 5,
+            expect_padded_flag=True,
+            expected_data_frame_pad_length=5
+        )
+
+    def test_sending_data_with_zero_length_padding(self):
+        """
+        Single data frames with zero-length padding are encoded
+        correctly.
+        """
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        # Clear the data, then send some data.
+        c.clear_outbound_data_buffer()
+        events = c.send_data(1, b'some data', pad_length=0)
+        assert not events
+        data_to_send = c.data_to_send()
+        assert data_to_send == (
+            b'\x00\x00\x0a\x00\x08\x00\x00\x00\x01'
+            b'\x00some data'
+        )
+
+        buffer = h2.frame_buffer.FrameBuffer(server=False)
+        buffer.max_frame_size = 65535
+        buffer.add_data(data_to_send)
+        data_frame = list(buffer)[0]
+        sanity_check_data_frame(
+            data_frame=data_frame,
+            expected_flow_controlled_length=len(b'some data') + 1,
+            expect_padded_flag=True,
+            expected_data_frame_pad_length=0
+        )
+
+    @pytest.mark.parametrize("expected_error,pad_length", [
+        (None,  0),
+        (None, 255),
+        (None, None),
+        (ValueError, -1),
+        (ValueError, 256),
+        (TypeError, 'invalid'),
+        (TypeError, ''),
+        (TypeError, '10'),
+        (TypeError, {}),
+        (TypeError, ['1', '2', '3']),
+        (TypeError, []),
+        (TypeError, 1.5),
+        (TypeError, 1.0),
+        (TypeError, -1.0),
+    ])
+    def test_sending_data_with_invalid_padding_length(self,
+                                                      expected_error,
+                                                      pad_length):
+        """
+        ``send_data`` with a ``pad_length`` parameter that is an integer
+        outside the range of [0, 255] throws a ``ValueError``, and a
+        ``pad_length`` parameter which is not an ``integer`` type
+        throws a ``TypeError``.
+        """
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.clear_outbound_data_buffer()
+        if expected_error is not None:
+            with pytest.raises(expected_error):
+                c.send_data(1, b'some data', pad_length=pad_length)
+        else:
+            c.send_data(1, b'some data', pad_length=pad_length)
 
     def test_closing_stream_sending_data(self, frame_factory):
         """
@@ -1576,3 +1683,27 @@ class TestBasicServer(object):
         assert c.state_machine.state == h2.connection.ConnectionState.CLOSED
 
         assert not c.data_to_send()
+
+
+def sanity_check_data_frame(data_frame,
+                            expected_flow_controlled_length,
+                            expect_padded_flag,
+                            expected_data_frame_pad_length):
+        """
+        ``data_frame`` is a frame of type ``hyperframe.frame.DataFrame``,
+        and the ``flags`` and ``flow_controlled_length`` of ``data_frame``
+        match expectations.
+        """
+
+        assert isinstance(data_frame, hyperframe.frame.DataFrame)
+
+        assert(
+            (data_frame.flow_controlled_length ==
+             expected_flow_controlled_length)
+        )
+        if expect_padded_flag:
+            assert 'PADDED' in data_frame.flags
+        else:
+            assert 'PADDED' not in data_frame.flags
+
+        assert data_frame.pad_length == expected_data_frame_pad_length
--- a/test/test_flow_control_window.py
+++ b/test/test_flow_control_window.py
@@ -52,6 +52,23 @@ class TestFlowControl(object):
         remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data')
         assert (c.local_flow_control_window(1) == remaining_length)
 
+    @pytest.mark.parametrize("pad_length", [5, 0])
+    def test_flow_control_decreases_with_sent_data_with_padding(self,
+                                                                pad_length):
+        """
+        When padded data is sent on a stream, the flow control window drops
+        by the length of the padding plus 1 for the 1-byte padding length
+        field.
+        """
+        c = h2.connection.H2Connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=pad_length)
+        remaining_length = (
+            self.DEFAULT_FLOW_WINDOW - len(b'some data') - pad_length - 1
+        )
+        assert c.local_flow_control_window(1) == remaining_length
+
     def test_flow_control_decreases_with_received_data(self, frame_factory):
         """
         When data is received on a stream, the remote flow control window
@@ -70,7 +87,7 @@ class TestFlowControl(object):
     def test_flow_control_decreases_with_padded_data(self, frame_factory):
         """
         When padded data is received on a stream, the remote flow control
-        window should drop by an amount that includes the padding.
+        window drops by an amount that includes the padding.
         """
         c = h2.connection.H2Connection(client_side=False)
         c.receive_data(frame_factory.preamble())
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -22,6 +22,9 @@ API Changes (Backward-Compatible)
   a subclass of ``int``, this is non-breaking.
 - Deprecated the other fields in ``h2.settings``. These will be removed in
   3.0.0.
+- Added an optional ``pad_length`` parameter to ``H2Connection.send_data``
+  to allow the user to include padding on a data frame.
+
 
 Bugfixes
 ~~~~~~~~


h2/connection.py
```diff
@@ -858,25 +858,35 @@ def send_data(self, stream_id, data, end_stream=False):
         :type data: ``bytes``
         :param end_stream: (optional) Whether this is the last data to be sent
             on the stream. Defaults to ``False``.
+        :param pad_length: (optional) Length of the padding to apply to the
+            data frame. Defaults to ``None`` for no use of padding. Note that
+            a value of ``0`` results in padding of length ``0``
+            (with the "padding" flag set on the frame).
```
alexwlchan: Please add a `versionadded: 2.6.0` directive to this parameter ([example](https://github.com/apolcyn/hyper-h2/blob/4322904ac68cce2f5271034199cf400a82e289de/h2/connection.py#L294)), so the documentation will reflect when it became available.

h2/connection.py
```diff
@@ -858,25 +858,35 @@ def send_data(self, stream_id, data, end_stream=False):
...
         :type end_stream: ``bool``
         :returns: Nothing
         """
-        if len(data) > self.local_flow_control_window(stream_id):
+        frame_size = len(data)
+        if pad_length is not None:
+            frame_size += pad_length + 1
```
alexwlchan: Maybe it’s too early in the morning – why do we need the `+1`? (if it’s not obvious, worth a comment?)

h2/stream.py
```diff
@@ -850,12 +850,17 @@ def send_data(self, data, end_stream=False):
...
         df.data = data
         if end_stream:
             self.state_machine.process_input(StreamInputs.SEND_END_STREAM)
             df.flags.add('END_STREAM')
+        if pad_length is not None:
+            df.flags.add('PADDED')
+            df.pad_length = pad_length
 
-        self.outbound_flow_control_window -= len(data)
+        # Subtract flow_controll_length to account for possible padding
```
alexwlchan: /s/flow_controll_length/flow_controlled_length

test/test_flow_control_window.py
```diff
@@ -52,6 +52,28 @@ def test_flow_control_decreases_with_sent_data(self):
...
+    def test_flow_control_decreases_with_sent_data_with_padding(self):
+        """
+        When data is sent on a stream, the flow control window should drop.
+        """
+        c = h2.connection.H2Connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=5)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 6
+        assert (c.local_flow_control_window(1) == remaining_length)
```
alexwlchan: Parentheses are unnecessary.

test/test_flow_control_window.py
```diff
@@ -52,6 +52,28 @@ def test_flow_control_decreases_with_sent_data(self):
...
+    def test_flow_control_decreases_with_sent_data_zero_length_padding(self):
+        """
+        When data is sent on a stream, the flow control window should drop.
+        """
+        c = h2.connection.H2Connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=0)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 1
+        assert (c.local_flow_control_window(1) == remaining_length)
```
alexwlchan: Ditto above.

test/test_basic_logic.py
```diff
@@ -113,6 +113,40 @@ def test_sending_data(self):
...
+        c.send_headers(1, self.example_request_headers)
+
+        # Clear the data, then send some data.
+        c.clear_outbound_data_buffer()
+        events = c.send_data(1, b'some data', pad_length=0)
+        assert not events
+        assert c.data_to_send() == (
+            b'\x00\x00\x0a\x00\x08\x00\x00\x00\x01'
+            b'\x00some data'
+        )
```
alexwlchan: I’d be keen on adding an `assert` to these tests that the "PADDED" flag is present in both cases. That might already be encoded in `data_to_send()`, but if a human could see that without unparsing that data. :-)

alexwlchan: Also, I think we should have a test for the fallback behaviour: if you pass `pad_length`, your frame _doesn’t_ have the "PADDED" flag set.

h2/connection.py
Lukasa: So, I know why we need the `+1`, but it's still worth a comment expressly because it's unclear unless you've read the RFC fairly closely.

(@alexwlchan an FYI, the reason is that when you add the PADDING flag you don't just add the padding bytes, you also add an extra byte that indicates how many padding bytes there are. That byte is also part of the flow controlled length.)

h2/connection.py
```diff
@@ -858,25 +858,35 @@ def send_data(self, stream_id, data, end_stream=False):
...
             raise FrameTooLargeError(
                 "Cannot send frame size %d, max frame size is %d" %
-                (len(data), self.max_outbound_frame_size)
+                (frame_size, self.max_outbound_frame_size)
             )
 
         self.state_machine.process_input(ConnectionInputs.SEND_DATA)
-        frames = self.streams[stream_id].send_data(data, end_stream)
+        frames = self.streams[stream_id].send_data(
+            data, end_stream, pad_length=pad_length)
```
Lukasa: This is a style thing for the project, but can the closing parenthesis be moved to a line of its own? See what is done for the `FrameTooLargeError` above.

h2/stream.py
```diff
@@ -850,12 +850,17 @@ def send_data(self, data, end_stream=False):
         self.state_machine.process_input(StreamInputs.SEND_DATA)
 
         df = DataFrame(self.stream_id)
+
```
Lukasa: We don't really need this whitespace line.

h2/connection.py
Lukasa: Can we please validate the input? That is, we allow `pad_length` to be `None` or an integer that is `>= 0`. I'd like us to actually check that and to raise a `ValueError` if it's wrong.

Note that you can go ahead and raise the `ValueError` inside this `if` block if you want, which saves checking twice.

test/test_flow_control_window.py
```diff
@@ -52,6 +52,28 @@ def test_flow_control_decreases_with_sent_data(self):
         remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data')
         assert (c.local_flow_control_window(1) == remaining_length)
 
+    def test_flow_control_decreases_with_sent_data_with_padding(self):
+        """
+        When data is sent on a stream, the flow control window should drop.
```
Lukasa: Do you mind updating this docstring to be unique and to explain specifically what it tests? That is, that the flow control window drops by an amount that includes the padding?

test/test_flow_control_window.py
```diff
@@ -52,6 +52,28 @@ def test_flow_control_decreases_with_sent_data(self):
...
+        c = h2.connection.H2Connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=5)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 6
+        assert (c.local_flow_control_window(1) == remaining_length)
+
+    def test_flow_control_decreases_with_sent_data_zero_length_padding(self):
+        """
+        When data is sent on a stream, the flow control window should drop.
```
Lukasa: Same note here: this docstring needs to be updated to note that we're testing that with zero-length padding the flow control window should drop by one byte more than the data we sent.

h2/connection.py
SethMichaelLarson: Also can't be longer than 255 if I remember?

Lukasa: Ah, good spot, also true.

apolcyn:
tried to address comments in latest 3 commits. 

@alexwlchan  wanted to reach out on one though :)
> ... adding an assert to these tests that the "PADDED" flag is present ...
A bit stuck, is there a convenience helper for grabbing the `DataFrame` that's to send, or parsing one from the `data_to_send()`?

h2/connection.py
```diff
@@ -859,24 +859,40 @@ def send_data(self, stream_id, data, end_stream=False):
         :param end_stream: (optional) Whether this is the last data to be sent
             on the stream. Defaults to ``False``.
         :type end_stream: ``bool``
+        :param pad_length: (optional) Length of the padding to apply to the
+            data frame. Defaults to ``None`` for no use of padding. Note that
+            a value of ``0`` results in padding of length ``0``
+            (with the "padding" flag set on the frame).
+        :type pad_length: ``int``
+            .. versionadded:: 2.6.0
```
alexwlchan: If you build this the docs locally, the directive doesn’t show up correctly. I confess I don’t fully understand how Sphinx directives work, but this block of RST seems to do the trick:

```rst
        :param pad_length: (optional) Length of the padding to apply to the
            data frame. Defaults to ``None`` for no use of padding. Note that
            a value of ``0`` results in padding of length ``0``
            (with the "padding" flag set on the frame).

            .. versionadded:: 2.6.0

        :type pad_length: ``int``
```

h2/connection.py
```diff
@@ -859,24 +859,40 @@ def send_data(self, stream_id, data, end_stream=False):
...
         :returns: Nothing
         """
-        if len(data) > self.local_flow_control_window(stream_id):
+        frame_size = len(data)
+        if pad_length is not None:
+            # Account for padding bytes and the 1-byte padding length field.
+            if pad_length < 0 or pad_length > 255:
+                raise ValueError("pad length must be within range: [0, 255]")
```
alexwlchan: What if I pass in a non-integer padding length?

Also, tweak the error message to use the exact name of the parameter (with the underscore).

h2/connection.py
```diff
@@ -859,24 +859,40 @@ def send_data(self, stream_id, data, end_stream=False):
...
+            frame_size += pad_length + 1
```
alexwlchan: Still missing a comment about why we have the extra `+1`.

test/test_basic_logic.py
```diff
@@ -113,6 +113,57 @@ def test_sending_data(self):
...
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.clear_outbound_data_buffer()
+        # 255 is valid
+        c.send_data(1, b'some data', pad_length=255)
+        with pytest.raises(ValueError):
+            c.send_data(1, b'some data', pad_length=-1)
+        with pytest.raises(ValueError):
+            c.send_data(1, b'some data', pad_length=256)
```
alexwlchan: We’ll want some extra tests here that we throw a `TypeError` if passing a non-integer padding length.

h2/connection.py
alexwlchan: (No, putting it into the tests doesn’t count. 😉 I might look at the tests if I’ve broken something, but it doesn’t help somebody who’s just casually reading the code.)

Lukasa:
> is there a convenience helper for grabbing the DataFrame that's to send, or parsing one from the data_to_send()?

Yes, there is. Just import the frame buffer from the codebase. You can see this done in `test_basic_logic` `test_oversize_headers`.

h2/connection.py
Lukasa: I'm inclined to want to leave that off: on Python 3 they'll mostly raise type errors, and on Python 2 it'll fail lower down the stack.

apolcyn:
latest two commits try to address recent comments again.

added in some simple type checks on the pad_length param too.

thanks for comments :) I think it's ready for another look

test/test_basic_logic.py
```diff
@@ -105,14 +105,142 @@ def test_sending_data(self):
...
+        assert len(frames) == 3
+        settings_frame = frames[0]
+        header_frame = frames[1]
+        data_frame = frames[2]
+        assert isinstance(settings_frame, hyperframe.frame.SettingsFrame)
+        assert isinstance(header_frame, hyperframe.frame.HeadersFrame)
+        assert isinstance(data_frame, hyperframe.frame.DataFrame)
+
+        assert 'PADDED' in data_frame.flags
+
```
Lukasa: So there is a bunch of duplicate code around checking the output of the stream and whether there is a padding flag, and if so how much. It might be nice to factor that out to a helper function.

test/test_basic_logic.py
```diff
@@ -105,14 +105,142 @@ def test_sending_data(self):
...
+        frames = list(buffer)
+        assert len(frames) == 3
+        settings_frame = frames[0]
+        header_frame = frames[1]
+        data_frame = frames[2]
+        assert isinstance(settings_frame, hyperframe.frame.SettingsFrame)
+        assert isinstance(header_frame, hyperframe.frame.HeadersFrame)
+        assert isinstance(data_frame, hyperframe.frame.DataFrame)
+
+        assert 'PADDED' in data_frame.flags
```
Lukasa: Let's also check the amount of padding and flow controlled length, just to cover all our bases?

test/test_basic_logic.py
```diff
@@ -105,14 +105,142 @@ def test_sending_data(self):
...
+
+    def test_sending_data_with_invalid_padding_length(self):
+        """
+        Pad_lengths < 0 or > 255 are be checked for and rejected.
```
Lukasa: You can remove `be` from this sentence.

test/test_basic_logic.py
```diff
@@ -105,14 +105,142 @@ def test_sending_data(self):
...
+        with pytest.raises(TypeError):
+            c.send_data(1, b'some data', pad_length='')
+        with pytest.raises(TypeError):
+            c.send_data(1, b'some data', pad_length='10')
+        with pytest.raises(TypeError):
+            c.send_data(1, b'some data', pad_length={})
+        with pytest.raises(TypeError):
+            c.send_data(1, b'some data', pad_length=[1, 2, 3])
+        with pytest.raises(TypeError):
+            c.send_data(1, b'some data', pad_length=[])
```
Lukasa: This test would be nicer if rewritten using`pytest.mark.parametrize`.

test/test_basic_logic.py
```diff
@@ -105,14 +105,142 @@ def test_sending_data(self):
...
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.clear_outbound_data_buffer()
+        # 255 is valid
+        c.send_data(1, b'some data', pad_length=255)
+        with pytest.raises(ValueError):
+            c.send_data(1, b'some data', pad_length=-1)
+        with pytest.raises(ValueError):
+            c.send_data(1, b'some data', pad_length=256)
```
Lukasa: This test would be nicer if written with `pytest.mark.parametrize`.

test/test_flow_control_window.py
```diff
@@ -52,6 +52,32 @@ def test_flow_control_decreases_with_sent_data(self):
         remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data')
         assert (c.local_flow_control_window(1) == remaining_length)
 
+    def test_flow_control_decreases_with_sent_data_with_padding(self):
+        """
+        When padded data is sent on a stream, the flow control window should
+        drop by the length of the padding plus 1 for the 1-byte padding length
+        field.
```
Lukasa: This is a minor nit, but we try to write these docstrings as imperative: that is, we don't write that X *should* happen, we write that it *does* happen. For example, this would be:

> When padded data is sent on a stream, the flow control window drops by the length of the padding plus 1 for the 1-byte padding length field.

test/test_flow_control_window.py
```diff
@@ -52,6 +52,32 @@ def test_flow_control_decreases_with_sent_data(self):
...
+
+        c.send_data(1, b'some data', pad_length=5)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 6
+        assert c.local_flow_control_window(1) == remaining_length
+
+    def test_flow_control_decreases_with_sent_data_zero_length_padding(self):
+        """
+        When zero-length-padded data is sent on a stream, the flow control
+        window should drop by the 1 for the 1-byte padding length field, which
+        is still present though zero-valued.
```
Lukasa: Again, this should be imperative.

test/test_basic_logic.py
apolcyn: err, while adding extra checks here, I think I encountered problem!

Currently it looks to me like the `hyperframe.DataFrame` class has an issue in the way it computes the flow controlled length. I think the fix is small, created https://github.com/python-hyper/hyperframe/pull/80 for it

apolcyn:
thanks for comments again :) latest updates tried to address.

extra checks seem to have pointed out an issue though, not sure what we should do about this one in the meantime since the "zero valued pad length" test will fail for the time being (fix in https://github.com/python-hyper/hyperframe/pull/80)

Lukasa:
I propose we shelve this patch until a new version of hyperframe is ready.

apolcyn:
Shelving this until the hyperframe update is published sounds good to me, thanks for the review on that!

Lukasa:
Ok, hyperframe v4.0.2 has been shipped and is ready to go. This can be unblocked now. =D

apolcyn:
great, much thanks! 

So I think the last updates should address the comments again so far. (Added a new helper function in `test/helpers.py` - but not completely sure where it should go..)

For me, tests passed locally after clearing out my `.tox` and forcing the update to the new `hyperframe` package. I'm not sure what needs to be done here, but if doable I believe that a similar sort of test re-run would work here on travis.

test/helpers.py
```diff
@@ -167,3 +168,29 @@ def build_alt_svc_frame(self, stream_id, origin, field):
...
+        assert(
+            (data_frame.flow_controlled_length ==
+             expected_flow_controlled_length)
+        )
+        if expect_padded_flag:
+            assert 'PADDED' in data_frame.flags
+        else:
+            assert 'PADDED' not in data_frame.flags
+
+        assert data_frame.pad_length == expected_data_frame_pad_length
```
Lukasa: Given that this is only used in one test location, I think it'd be better to make it a helper function in that file rather than place it in a central location.

test/test_basic_logic.py
```diff
@@ -105,14 +105,118 @@ def test_sending_data(self):
...
-            c.data_to_send() == b'\x00\x00\t\x00\x00\x00\x00\x00\x01some data'
+            data_to_send == b'\x00\x00\t\x00\x00\x00\x00\x00\x01some data'
+        )
+
+        buffer.add_data(data_to_send)
+        helpers.sanity_check_settings_header_data(
+            frames=list(buffer),
+            expected_flow_controlled_length=len(b'some data'),
+            expect_padded_flag=False,
+            expected_data_frame_pad_length=0)
```
Lukasa: Can you drop this closing parenthesis to the next line please?

test/test_basic_logic.py
```diff
@@ -105,14 +105,118 @@ def test_sending_data(self):
...
+            b'\x00\x00\x0f\x00\x08\x00\x00\x00\x01'
+            b'\x05some data\x00\x00\x00\x00\x00'
         )
 
+        buffer.add_data(data_to_send)
+        helpers.sanity_check_settings_header_data(
+            frames=list(buffer),
+            expected_flow_controlled_length=len(b'some data') + 1 + 5,
+            expect_padded_flag=True,
+            expected_data_frame_pad_length=5)
```
Lukasa: Same closing parenthesis note here.

test/test_basic_logic.py
```diff
@@ -105,14 +105,118 @@ def test_sending_data(self):
...
+            b'\x00\x00\x0a\x00\x08\x00\x00\x00\x01'
+            b'\x00some data'
+        )
+
+        buffer.add_data(data_to_send)
+        helpers.sanity_check_settings_header_data(
+            frames=list(buffer),
+            expected_flow_controlled_length=len(b'some data') + 1,
+            expect_padded_flag=True,
+            expected_data_frame_pad_length=0)
```
Lukasa: Same closing parenthesis note here.

test/test_basic_logic.py
```diff
@@ -105,14 +105,118 @@ def test_sending_data(self):
...
+        ([], TypeError),
+    ])
+    def test_sending_data_with_invalid_padding_length(self,
+                                                      pad_length,
+                                                      expected_error):
+        """
+        Integer valued pad_length values < 0 or > 255 are checked for
+        and rejected.
+        Also ``send_data`` with a ``pad_length`` parameter that is neither
+        ``None`` or an ``integer`` raises a ``TypeError``
```
Lukasa: I'd rather have this docstring be one sentence if we can do it.

test/test_flow_control_window.py
Lukasa: This still needs to be made imperative.

test/test_flow_control_window.py
```diff
@@ -52,6 +52,32 @@ def test_flow_control_decreases_with_sent_data(self):
...
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=5)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 6
+        assert c.local_flow_control_window(1) == remaining_length
+
+    def test_flow_control_decreases_with_sent_data_zero_length_padding(self):
+        """
+        When zero-length-padded data is sent on a stream, the flow control
+        window is drop by the 1 for the 1-byte padding length field (which
```
Lukasa: Nit: we don't want "is drop", we want "drops".

apolcyn:
Latest update should address comments again.

I think the introduced `non-imperative` should be fixed..

test/test_basic_logic.py
```diff
@@ -1576,3 +1683,29 @@ def test_receiving_goaway_frame_with_unknown_error(self, frame_factory):
         assert c.state_machine.state == h2.connection.ConnectionState.CLOSED
 
         assert not c.data_to_send()
+
+
+def sanity_check_settings_header_data(frames,
+                                      expected_flow_controlled_length,
+                                      expect_padded_flag,
+                                      expected_data_frame_pad_length):
```
alexwlchan: Let’s get a docstring on this function to explain what it’s checking.

test/test_basic_logic.py
```diff
@@ -1576,3 +1683,29 @@ def test_receiving_goaway_frame_with_unknown_error(self, frame_factory):
...
+        assert len(frames) == 3
+        settings_frame = frames[0]
+        header_frame = frames[1]
+        data_frame = frames[2]
```
alexwlchan: Alternative:

```python
settings_frame, header_frame, data_frame = frames
```

test/test_flow_control_window.py
```diff
@@ -52,6 +52,32 @@ def test_flow_control_decreases_with_sent_data(self):
...
+        When zero-length-padded data is sent on a stream, the flow control
+        window drops by the 1 for the 1-byte padding length field (which
+        is still present though zero-valued).
+        """
+        c = h2.connection.H2Connection()
+        c.send_headers(1, self.example_request_headers)
+
+        c.send_data(1, b'some data', pad_length=0)
+        remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - 1
+        assert c.local_flow_control_window(1) == remaining_length
```
alexwlchan: These two tests are very similar. Could we combine them into two tests which take a `padding_length` parameter, and then define:

```python
remaining_length = self.DEFAULT_FLOW_WINDOW - len(b'some data') - padding_length - 1
```

?

test/test_basic_logic.py
```diff
@@ -105,14 +105,121 @@ def test_sending_data(self):
...
+        (255, None),
+        (None, None),
+        (-1, ValueError),
+        (256, ValueError),
+        ('invalid', TypeError),
+        ('', TypeError),
+        ('10', TypeError),
+        ({}, TypeError),
+        (['1', '2', '3'], TypeError),
+        ([], TypeError),
```
alexwlchan: Style nit: I’d consider rewriting this to put the error first. That way cases line up a bit more neatly:

```
(None, 0)
(None, 255)
(None, None)
(ValueError, -1)
(ValueError, 256)
(TypeError, 'invalid')
...
```

It’s easy to skim and see where the `ValueError`'s stop and the `TypeError`s start.

alexwlchan: Also, let’s throw in a floating point number, which could be integral but is still disallowed.

apolcyn:
latest update tries to address comments from @alexwlchan (in commit b0ca7fc) 

not super confident in the docstring for `sanity_check_settings_header_data` but I think it's ready for another look.

test/test_basic_logic.py
```diff
@@ -1576,3 +1686,34 @@ def test_receiving_goaway_frame_with_unknown_error(self, frame_factory):
...
+
+def sanity_check_settings_header_data(frames,
+                                      expected_flow_controlled_length,
+                                      expect_padded_flag,
+                                      expected_data_frame_pad_length):
+        """
+        ```frames``` is a list of http2 frames in order of types
+        "settings", "header", and "data", and the ``flags``
+        and ```flow_controlled_length``` of the "data frame" match
+        expectations.
```
Lukasa: So, while I was writing a docstring for this function I realised that it is just doing some unnecessary work. I think we should change it so that it doesn't bother to receive the first two frames, as it doesn't do any meaningful validation of them, and it limits the scope of the function to only test straight after connection setup.

We can just ignore the first two frames and assume that they're not relevant in the tests.

apolcyn:
@Lukasa  sounds good, latest update just tried to simplify the helper function to only check the data frame.

test/test_basic_logic.py
```diff
@@ -105,14 +105,127 @@ def test_sending_data(self):
...
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        buffer = h2.frame_buffer.FrameBuffer(server=True)
+        buffer.max_frame_size = 65535
+        buffer.add_data(c.data_to_send())
+
+        # Clear the data, then send some data.
+        c.clear_outbound_data_buffer()
```
Lukasa: This shouldn't be needed if you've called `data_to_send()`.

test/test_basic_logic.py
```diff
@@ -105,14 +105,127 @@ def test_sending_data(self):
         c.initiate_connection()
         c.send_headers(1, self.example_request_headers)
 
+        buffer = h2.frame_buffer.FrameBuffer(server=True)
+        buffer.max_frame_size = 65535
+        buffer.add_data(c.data_to_send())
+
         # Clear the data, then send some data.
         c.clear_outbound_data_buffer()
```
Lukasa: This is no longer needed.

test/test_basic_logic.py
```diff
@@ -105,14 +105,127 @@ def test_sending_data(self):
...
+        c = h2.connection.H2Connection()
+        c.initiate_connection()
+        c.send_headers(1, self.example_request_headers)
+
+        buffer = h2.frame_buffer.FrameBuffer(server=True)
+        buffer.max_frame_size = 65535
+        buffer.add_data(c.data_to_send())
+
+        # Clear the data, then send some data.
+        c.clear_outbound_data_buffer()
```
Lukasa: Also not needed.

apolcyn:
removed the extra frames added to buffer, I think should be ready again

test/test_basic_logic.py
```diff
@@ -109,10 +109,117 @@ def test_sending_data(self):
...
+        (ValueError, 256),
+        (TypeError, 'invalid'),
+        (TypeError, ''),
+        (TypeError, '10'),
+        (TypeError, {}),
+        (TypeError, ['1', '2', '3']),
+        (TypeError, []),
+        (TypeError, 1.5),
+        (TypeError, 1.0),
+        (TypeError, -1.0),
```
alexwlchan: 👍

test/test_basic_logic.py
```diff
@@ -1576,3 +1683,27 @@ def test_receiving_goaway_frame_with_unknown_error(self, frame_factory):
...
 
         assert not c.data_to_send()
+
+
+def sanity_check_data_frame(data_frame,
+                            expected_flow_controlled_length,
+                            expect_padded_flag,
+                            expected_data_frame_pad_length):
+        """
+        ```data_frame``` is a frame of type ```hyperframe.frame.DataFrame```,
```
alexwlchan: … triple quotes?

Lukasa:
Yeah, squash and changelog would be good.

alexwlchan:
> squash and changelog would be good

Done.

test/test_basic_logic.py
apolcyn: ah woops I can still add to the branch, unless we should freeze it here?

Lukasa:
Cool, looks great!

Lukasa:
Thanks so much @apolcyn! Per our [one of the team policy](http://python-hyper.org/en/latest/one-of-the-team.html), you'll find an invite to the Python-hyper organisation in your inbox. Welcome on board!

SethMichaelLarson:
:tada:

apolcyn:
great thanks for the time in the review, this was really helpful!

PR #303, opened 2016-08-24
Add better representation for error codes.

Lukasa:
Resolves #295. /cc @Kriechi

PR #469, opened 2017-02-17
Define SettingCodes and improve ChangedSetting representation.

### LINKED ISSUE #359: Provide SettingsEnum to dynamically list and print
Similar to #303

It would be nice to have a list/dict/enum of all valid settings (key + names).
This can be useful for printing debug information or just plain logging of changed settings.

Kriechi:
fixes #359
based on #303

### DIFF
--- a/h2/events.py
+++ b/h2/events.py
@@ -11,7 +11,7 @@ H2 state machine it processes the data and returns a list of Event objects.
 """
 import binascii
 
-from .settings import ChangedSetting
+from .settings import ChangedSetting, _setting_code_from_int
 
 
 class Event(object):
@@ -343,6 +343,7 @@ class RemoteSettingsChanged(Event):
         """
         e = cls()
         for setting, new_value in new_settings.items():
+            setting = _setting_code_from_int(setting)
             original_value = old_settings.get(setting)
             change = ChangedSetting(setting, original_value, new_value)
             e.changed_settings[setting] = change
@@ -350,8 +351,8 @@ class RemoteSettingsChanged(Event):
         return e
 
     def __repr__(self):
-        return "<RemoteSettingsChanged changed_settings:%s>" % (
-            self.changed_settings,
+        return "<RemoteSettingsChanged changed_settings:{%s}>" % (
+            ", ".join(repr(cs) for cs in self.changed_settings.values()),
         )
 
 
@@ -453,8 +454,8 @@ class SettingsAcknowledged(Event):
         self.changed_settings = {}
 
     def __repr__(self):
-        return "<SettingsAcknowledged changed_settings:%s>" % (
-            self.changed_settings,
+        return "<SettingsAcknowledged changed_settings:{%s}>" % (
+            ", ".join(repr(cs) for cs in self.changed_settings.values()),
         )
 
 
--- a/h2/settings.py
+++ b/h2/settings.py
@@ -8,6 +8,7 @@ API for manipulating HTTP/2 settings, keeping track of both the current active
 state of the settings and the unacknowledged future values of the settings.
 """
 import collections
+import enum
 
 from hyperframe.frame import SettingsFrame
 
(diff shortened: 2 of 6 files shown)


h2/settings.py
```diff
@@ -8,58 +8,139 @@
...
 
 from h2.errors import ErrorCodes
 from h2.exceptions import InvalidSettingsValueError
 
 
+class SettingCodes(enum.IntEnum):
+    """
+        All known HTTP/2 setting codes.
+
+        .. versionadded:: 2.5.3
```
Lukasa: This docstring can be dedented.

As to the versionadded comment, this won't be able to make it into the 2.5 release because this is a public part of the API, and so will need to go into 2.6. So the version added is 2.6.0.

h2/settings.py
```diff
@@ -8,58 +8,139 @@
...
+
+
 # Aliases for all the settings values.
 
 #: Allows the sender to inform the remote endpoint of the maximum size of the
 #: header compression table used to decode header blocks, in octets.
-HEADER_TABLE_SIZE = SettingsFrame.HEADER_TABLE_SIZE
+#:
+#: .. deprecated:: 2.6.0
+#:    Deprecated in favour of :class:`SettingCodes.HEADER_TABLE_SIZE
```
Lukasa: These are all `class` but should be `data`, I think.

h2/settings.py
```diff
@@ -8,58 +8,139 @@
...
+
+class ChangedSetting:
 
-try:  # Platform-specific: Hyperframe < 4.0.0
-    MAX_FRAME_SIZE = SettingsFrame.SETTINGS_MAX_FRAME_SIZE
-except AttributeError:  # Platform-specific: Hyperframe >= 4.0.0
-    MAX_FRAME_SIZE = SettingsFrame.MAX_FRAME_SIZE
+    def __init__(self, setting, original_value, new_value):
+        #: The setting code given. Either one of :class:`SettingCodes
+        #: <h2.settings.SettingCodes>` or ``int``
```
Lukasa: This should get a `.. versionchanged:: 2.6.0` note as well.

h2/settings.py
```diff
@@ -8,58 +8,139 @@
...
+        self.new_value = new_value
 
-#: A value structure for storing changed settings.
-ChangedSetting = collections.namedtuple(
-    'ChangedSetting', ['setting', 'original_value', 'new_value']
-)
+    def __repr__(self):
+        return ("ChangedSetting(setting=%s, original_value=%s, "
+        "new_value=%s)") % (
+            self.setting.name,
```
Lukasa: Does `.name` work here in all cases?

h2/settings.py
Kriechi: Done.

h2/settings.py
Kriechi: Done.

h2/settings.py
Kriechi: Done.

h2/settings.py
Kriechi: It should... Do you want me to add an `if isinstance(self.setting, SettingCodes)` check?

Lukasa: Well, I mean: it definitely won't work for integers, right? :grin:

Rather than isinstance, are you happy to just tolerate the default repr of an intenum member?

Kriechi: How about we make sure that the init of `ChangedSetting` always converts it to and  `SettingCodes` object?

Lukasa: You can't convert to an enum if you don't have a member defined for that int.

h2/settings.py
```diff
@@ -8,58 +8,143 @@
...
 
 from h2.errors import ErrorCodes
 from h2.exceptions import InvalidSettingsValueError
 
 
+class SettingCodes(enum.IntEnum):
+    """
+    All known HTTP/2 setting codes.
+
+    .. versionadded:: 2.5.3
```
alexwlchan: 2.6.0, surely?

h2/settings.py
Kriechi: Ok, I'll changed it to use the default repr.

h2/settings.py
Kriechi: Done.

h2/events.py
```diff
@@ -343,15 +343,16 @@ def from_settings(cls, old_settings, new_settings):
...
             change = ChangedSetting(setting, original_value, new_value)
             e.changed_settings[setting] = change
 
         return e
 
     def __repr__(self):
-        return "<RemoteSettingsChanged changed_settings:%s>" % (
-            self.changed_settings,
+        return "<RemoteSettingsChanged changed_settings:{%s}>" % (
+            ", ".join([repr(cs) for cs in self.changed_settings.values()]),
```
Lukasa: Small note here, but you don't need the list. `join` works just fine on generators, so we can save ourselves the allocation of an intermediate list here and below.

h2/settings.py
```diff
@@ -8,58 +8,147 @@
...
-)
+    def __repr__(self):
+        return (
+            "ChangedSetting(setting=%s, original_value=%s, "
+            "new_value=%s)"
+        ) % (
+            self.setting,
+            self.original_value,
+            self.new_value
+        )
```
Lukasa: Wow. This looks weird. It fits in with the code style, so...I guess we'll leave it.

Kriechi: Either this, or you allow lines with more than 79 characters! 😄

h2/events.py
Kriechi: Done.

h2/settings.py
Lukasa: NEVER!

PR #471, opened 2017-02-19
Update H2Configuration test to use init params

### LINKED ISSUE #454: Extend config tests in `test_config.py` to use the constructor
Picking up one of @Lukasa’s comments on #436: all the tests in `test_config.py` try to set attributes on `H2Configuration` directly, e.g.

```python
def test_boolean_config_option_is_reflected(self, option_name, value):
     """
     The value of the boolean config options, when set, is reflected
     in the value.
     """
     config = h2.config.H2Configuration()
     setattr(config, option_name, value)
     assert getattr(config, option_name) == value
```

There’s another way to use `H2Configuration`: by passing arguments to the constructor, e.g.

```python
config = h2.config.H2Configuration(client_side=False)
```

It would be good to have some tests that exercise the constructor as well as direct attribute access.

fredthomsen:
Update all tests in TestH2Config to use the H2Configuration initializer
parameters and verify that attributes of that object indeed to get set
properly via those parameters.

This resolves #454.  Not sure if splitting these tests into two, one containing the constructor, and another containing the attribute modifications would be more appropriate here.

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -36,8 +36,10 @@ class TestH2Config(object):
         The boolean config options raise an error if you try to set a value
         that isn't a boolean.
         """
-        config = h2.config.H2Configuration()
+        with pytest.raises(ValueError):
+            config = h2.config.H2Configuration(**{option_name: value})
 
+        config = h2.config.H2Configuration()
         with pytest.raises(ValueError):
             setattr(config, option_name, value)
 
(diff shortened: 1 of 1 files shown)


Lukasa:
Per our [one of the team](http://python-hyper.org/en/latest/one-of-the-team.html) policy, you'll find an invite to the python-hyper organisation in your inbox @fredthomsen. Welcome on board!

fredthomsen:
Thanks @Lukasa and @alexwlchan!  Looking forward to more learning and doing.

Issue #470, opened 2017-02-19
Recv frame on closed state should be connection error or stream error


chhsiao90:
Currently, hyper-h2 always send stream error when received unexpected frame on stream in closed state.

But in [rfc7540#section-5.1](https://tools.ietf.org/html/rfc7540#section-5.1)
> An endpoint MUST NOT send frames other than PRIORITY on a closed
      stream.  An endpoint that receives any frame other than PRIORITY
      after receiving a RST_STREAM MUST treat that as a stream error
      (Section 5.4.2) of type STREAM_CLOSED.  Similarly, an endpoint
      that receives any frames after receiving a frame with the
      END_STREAM flag set MUST treat that as a connection error
      (Section 5.4.1) of type STREAM_CLOSED, unless the frame is
      permitted as described below.

Which action should be taken, connection error or stream error, depends on how the stream had been closed.
If stream is closed by RST_STREAM, it should return stream error when received unexpected frame.
If stream is closed by END_STREAM, it should return connection error when received unexpected frame.

Lukasa:
Yup, this sounds like a reasonable change.

chhsiao90:
I would like to work on this issue.
I think that the PR solved this may have some major change on the code.
So I will ask your advise that which to change will better.

One idea is added a field in the state machine of stream called ```stream_closed_by``` that had state how this stream is closed by.
Here is the code https://github.com/chhsiao90/hyper-h2/commit/9b5ad36ad7b4de833b362c163665cc12d64178b4

And will add the logic at ```H2StreamStateMachine#send_on_closed_stream``` and a new function ```send_on_closed_stream#recv_on_closed_stream```

WDYT

Lukasa:
I think that approach is reasonable, if you open a PR I'll provide more feedback on the specifics. 👍

chhsiao90:
No problem, I will keep working on that and will create a PR soon!

Lukasa:
Resolved by #472.

PR #472, opened 2017-02-20
ProtocolError or StreamClosedError depends on how the stream is closed by

chhsiao90:
Related to: #470 

I will add more tests in next commits.
Maybe I will try hypothesis :)

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -378,11 +378,12 @@ class H2Connection(object):
         # Data that needs to be sent.
         self._data_to_send = b''
 
-        # Keeps track of how streams is closed by.
+        # Keeps track of how streams are closed.
         # Used to ensure that we don't blow up in the face of frames that were
         # in flight when a RST_STREAM was sent.
-        # And used to determine stream error or connection error when more
-        # frames was received at closed state.
+        # Also used to determine whether we should consider a frame received
+        # while a stream is closed as either a stream error or a connection
+        # error.
         self._closed_streams = {}
 
         # The flow control window manager for the connection.
@@ -1469,8 +1470,9 @@ class H2Connection(object):
             # I don't love using __class__ here, maybe reconsider it.
             frames, events = self._frame_dispatch_table[frame.__class__](frame)
         except StreamClosedError as e:
-            # If stream is closed by RST_STREAM, just send RST_STREAM to remote
-            # end point. Otherwise, re-raise and closed the connection
+            # If the stream was closed by RST_STREAM, we just send a RST_STREAM
+            # to the remote peer. Otherwise, this is a connection error, and so
+            # we will re-raise to trigger one.
             if self._stream_is_closed_by_reset(e.stream_id):
                 f = RstStreamFrame(e.stream_id)
                 f.error_code = e.error_code
@@ -1480,18 +1482,24 @@ class H2Connection(object):
                 raise
         except StreamIDTooLowError as e:
             # The stream ID seems invalid. This may happen when the closed
-            # stream had bean cleanup, or when there is another stream opened
-            # with higher stream id before this stream open
-            # We will consider how the stream is closed by and treat it as
-            # stream error or connection error depends
+            # stream has been cleaned up, or when the remote peer has opened a
+            # new stream with a higher stream ID than this one, forcing it
+            # closed implicitly.
+            #
+            # Check how the stream was closed: depending on the mechanism, it
+            # is either a stream error or a connection error.
             if self._stream_is_closed_by_reset(e.stream_id):
+                # Closed by RST_STREAM is a stream error.
                 f = RstStreamFrame(e.stream_id)
                 f.error_code = ErrorCodes.STREAM_CLOSED
                 self._prepare_for_sending([f])
                 events = []
             elif self._stream_is_closed_by_end(e.stream_id):
+                # Closed by END_STREAM is a connection error.
                 raise StreamClosedError(e.stream_id)
             else:
+                # Closed implicitly, also a connection error, but of type
+                # PROTOCOL_ERROR.
                 raise
         except KeyError as e:  # pragma: no cover
             # We don't have a function for handling this frame. Let's call this
@@ -1602,7 +1610,7 @@ class H2Connection(object):
         except StreamClosedError:
             # The parent stream was reset by us, so we presume that
             # PUSH_PROMISE was in flight when we reset the parent stream.
-            # So we just reset the new stream
+            # So we just reset the new stream.
             f = RstStreamFrame(frame.promised_stream_id)
             f.error_code = ErrorCodes.REFUSED_STREAM
             return [f], events
@@ -1872,9 +1880,12 @@ class H2Connection(object):
 
     def _stream_closed_by(self, stream_id):
         """
-        Returns how the stream closed, the value will be
-        ``h2.stream.StreamClosedBy`` or ``None`` if the stream is closed by
-        another opened stream had higher stream id.
+        Returns how the stream was closed.
+
+        The return value will be either a member of
+        ``h2.stream.StreamClosedBy`` or ``None``. If ``None``, the stream was
+        closed implicitly by the peer opening a stream with a higher stream ID
+        before opening this one.
         """
         if stream_id in self.streams:
             return self.streams[stream_id].closed_by
@@ -1884,19 +1895,22 @@ class H2Connection(object):
 
     def _stream_is_closed_by_reset(self, stream_id):
         """
-        Returns ``True`` if the stream is closed by send or receive RST_STREAM
-        frame, returns ``False`` otherwise.
+        Returns ``True`` if the stream was closed by sending or receiving a
+        RST_STREAM frame. Returns ``False`` otherwise.
         """
         return self._stream_closed_by(stream_id) in (
-            StreamClosedBy.RECV_RST_STREAM, StreamClosedBy.SEND_RST_STREAM)
+            StreamClosedBy.RECV_RST_STREAM, StreamClosedBy.SEND_RST_STREAM
+        )
 
     def _stream_is_closed_by_end(self, stream_id):
         """
-        Returns ``True`` if the stream is closed by send or receive END_STREAM
-        flag in HEADER or DATA frame, returns ``False`` otherwise.
+        Returns ``True`` if the stream was closed by sending or receiving an
+        END_STREAM flag in a HEADERS or DATA frame. Returns ``False``
+        otherwise.
         """
         return self._stream_closed_by(stream_id) in (
-            StreamClosedBy.RECV_END_STREAM, StreamClosedBy.SEND_END_STREAM)
+            StreamClosedBy.RECV_END_STREAM, StreamClosedBy.SEND_END_STREAM
+        )
 
 
 def _add_frame_priority(frame, weight=None, depends_on=None, exclusive=None):
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -33,7 +33,7 @@ from .exceptions import (
 )
 from .frame_buffer import FrameBuffer
 from .settings import Settings, SettingCodes
-from .stream import H2Stream
+from .stream import H2Stream, StreamClosedBy
 from .utilities import guard_increment_window
 from .windows import WindowManager
 
@@ -378,10 +378,12 @@ class H2Connection(object):
         # Data that needs to be sent.
         self._data_to_send = b''
 
-        # Keeps track of streams that have been forcefully reset by this peer.
+        # Keeps track of how streams is closed by.
         # Used to ensure that we don't blow up in the face of frames that were
-        # in flight when a stream was reset.
-        self._reset_streams = set()
+        # in flight when a RST_STREAM was sent.
+        # And used to determine stream error or connection error when more
+        # frames was received at closed state.
+        self._closed_streams = {}
 
         # The flow control window manager for the connection.
         self._inbound_flow_control_window_manager = WindowManager(
@@ -425,7 +427,8 @@ class H2Connection(object):
                 to_delete.append(stream_id)
 
         for stream_id in to_delete:
-            del self.streams[stream_id]
+            stream = self.streams.pop(stream_id)
+            self._closed_streams[stream_id] = stream.closed_by
 
         return count
 
@@ -1015,8 +1018,6 @@ class H2Connection(object):
         stream = self._get_stream_by_id(stream_id)
         frames = stream.reset_stream(error_code)
         self._prepare_for_sending(frames)
-        self._reset_streams.add(stream_id)
-        del self.streams[stream_id]
 
     def close_connection(self, error_code=0, additional_data=None,
                          last_stream_id=None):
@@ -1468,27 +1469,30 @@ class H2Connection(object):
             # I don't love using __class__ here, maybe reconsider it.
             frames, events = self._frame_dispatch_table[frame.__class__](frame)
         except StreamClosedError as e:
-            # We need to send a RST_STREAM frame on behalf of the stream.
-            # The frame the stream wants to emit is already present in the
-            # exception.
-            # This does not require re-raising: it's an expected behaviour. The
-            # only time we don't do that is if this is a stream the user
-            # manually reset.
-            if frame.stream_id not in self._reset_streams:
+            # If stream is closed by RST_STREAM, just send RST_STREAM to remote
+            # end point. Otherwise, re-raise and closed the connection
+            if self._stream_is_closed_by_reset(e.stream_id):
                 f = RstStreamFrame(e.stream_id)
                 f.error_code = e.error_code
                 self._prepare_for_sending([f])
                 events = e._events
             else:
-                events = []
-        except StreamIDTooLowError as e:
-            # The stream ID seems invalid. This is unlikely, so it's probably
-            # the case that this frame is actually for a stream that we've
-            # already reset and removed the state for. If it is, just swallow
-            # the error. If we didn't do that, re-raise.
-            if frame.stream_id not in self._reset_streams:
                 raise
-            events = []
+        except StreamIDTooLowError as e:
+            # The stream ID seems invalid. This may happen when the closed
+            # stream had bean cleanup, or when there is another stream opened
+            # with higher stream id before this stream open
+            # We will consider how the stream is closed by and treat it as
+            # stream error or connection error depends
+            if self._stream_is_closed_by_reset(e.stream_id):
+                f = RstStreamFrame(e.stream_id)
+                f.error_code = ErrorCodes.STREAM_CLOSED
+                self._prepare_for_sending([f])
+                events = []
+            elif self._stream_is_closed_by_end(e.stream_id):
+                raise StreamClosedError(e.stream_id)
+            else:
+                raise
         except KeyError as e:  # pragma: no cover
             # We don't have a function for handling this frame. Let's call this
             # a PROTOCOL_ERROR and exit.
@@ -1573,7 +1577,8 @@ class H2Connection(object):
             # PROTOCOL_ERROR: pushing a stream on a naturally closed stream is
             # a real problem because it creates a brand new stream that the
             # remote peer now believes exists.
-            if frame.stream_id in self._reset_streams:
+            if (self._stream_closed_by(frame.stream_id) ==
+                    StreamClosedBy.SEND_RST_STREAM):
                 f = RstStreamFrame(frame.promised_stream_id)
                 f.error_code = ErrorCodes.REFUSED_STREAM
                 return [f], events
@@ -1588,11 +1593,19 @@ class H2Connection(object):
         if (frame.stream_id % 2) == 0:
             raise ProtocolError("Cannot recursively push streams.")
 
-        frames, stream_events = stream.receive_push_promise_in_band(
-            frame.promised_stream_id,
-            pushed_headers,
-            self.config.header_encoding,
-        )
+        try:
+            frames, stream_events = stream.receive_push_promise_in_band(
+                frame.promised_stream_id,
+                pushed_headers,
+                self.config.header_encoding,
+            )
+        except StreamClosedError:
+            # The parent stream was reset by us, so we presume that
+            # PUSH_PROMISE was in flight when we reset the parent stream.
+            # So we just reset the new stream
+            f = RstStreamFrame(frame.promised_stream_id)
+            f.error_code = ErrorCodes.REFUSED_STREAM
+            return [f], events
 
         new_stream = self._begin_new_stream(
             frame.promised_stream_id, AllowedStreamIDs.EVEN
@@ -1856,6 +1869,34 @@ class H2Connection(object):
         (one initiated by this peer), returns ``False`` otherwise.
         """
         return (stream_id % 2 == int(self.config.client_side))
+
+    def _stream_closed_by(self, stream_id):
+        """
+        Returns how the stream closed, the value will be
+        ``h2.stream.StreamClosedBy`` or ``None`` if the stream is closed by
+        another opened stream had higher stream id.
+        """
+        if stream_id in self.streams:
+            return self.streams[stream_id].closed_by
+        if stream_id in self._closed_streams:
+            return self._closed_streams[stream_id]
+        return None
+
+    def _stream_is_closed_by_reset(self, stream_id):
+        """
+        Returns ``True`` if the stream is closed by send or receive RST_STREAM
+        frame, returns ``False`` otherwise.
+        """
+        return self._stream_closed_by(stream_id) in (
+            StreamClosedBy.RECV_RST_STREAM, StreamClosedBy.SEND_RST_STREAM)
+
+    def _stream_is_closed_by_end(self, stream_id):
+        """
+        Returns ``True`` if the stream is closed by send or receive END_STREAM
+        flag in HEADER or DATA frame, returns ``False`` otherwise.
+        """
+        return self._stream_closed_by(stream_id) in (
+            StreamClosedBy.RECV_END_STREAM, StreamClosedBy.SEND_END_STREAM)
 
 
 def _add_frame_priority(frame, weight=None, depends_on=None, exclusive=None):
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -64,6 +64,13 @@ class StreamInputs(Enum):
     UPGRADE_SERVER = 18  # Added 2.3.0
 
 
+class StreamClosedBy(Enum):
+    SEND_END_STREAM = 0
+    RECV_END_STREAM = 1
+    SEND_RST_STREAM = 2
+    RECV_RST_STREAM = 3
+
+
 # This array is initialized once, and is indexed by the stream states above.
 # It indicates whether a stream in the given state is open. The reason we do
 # this is that we potentially check whether a stream in a given state is open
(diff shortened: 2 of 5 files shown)


h2/stream.py
```diff
@@ -64,6 +64,13 @@ class StreamInputs(Enum):
     UPGRADE_SERVER = 18  # Added 2.3.0
 
 
+class StreamClosedBy(IntEnum):
```
Lukasa: We don't need this to be an `IntEnum`. This doesn't ever need to be coerced to an integer for any reason, so we can just use an `Enum`. If you do that, the natural values of that enum are printable strings you can use in `recv_on_closed_stream`.

h2/stream.py
```diff
@@ -206,6 +216,8 @@ def stream_ended(self, previous_state):
         """
         Fires when a stream is cleanly ended.
         """
+        if previous_state is StreamState.HALF_CLOSED_LOCAL:
+            self.stream_closed_by = StreamClosedBy.RECV_END_STREAM
```
Lukasa: Rather than check what the initial state is, it would be better to split this function up so that the right one is called based on the predicate state.

That is, rather than have `stream_ended` be called from two or more different `previous_state`, have two functions: one the regular `stream_ended` function as it was, and one that is only called from `HALF_CLOSED_LOCAL` and that does the `stream_closed_by` call *and then* calls the original `stream_ended` function.

h2/stream.py
```diff
@@ -269,6 +282,10 @@ def recv_push_promise(self, previous_state):
         event.parent_stream_id = self.stream_id
         return [event]
 
+    def send_end_stream(self, previous_state):
+        if previous_state is StreamState.HALF_CLOSED_REMOTE:
```
Lukasa: Same note here: it's better to have this function be called only in the appropriate state than to do this checking.

h2/stream.py
```diff
@@ -98,6 +105,9 @@ def __init__(self, stream_id):
         self.headers_received = None
         self.trailers_received = None
 
+        # Whether how the stream is closed by send/recv END_STREAM/RST_STREAM
```
Lukasa: Nit: Remove `how`.

h2/stream.py
```diff
@@ -318,6 +357,34 @@ def push_on_closed_stream(self, previous_state):
         assert previous_state == StreamState.CLOSED
         raise ProtocolError("Attempted to push on closed stream.")
 
+    def window_on_closed_stream(self, previous_state):
+        """
+        Called when a WINDOW_UPDATE frame received on an already-closed stream
+
+        If a END_STREAM had just sent for a short period, then just ignore the
```
Lukasa: This docstring talks about `END_STREAM`, but the code looks for `RST_STREAM`.

h2/stream.py
```diff
@@ -633,19 +700,33 @@ def send_alt_svc(self, previous_state):
     (StreamState.CLOSED, StreamInputs.RECV_ALTERNATIVE_SERVICE):
         (None, StreamState.CLOSED),
 
-    # While closed, all other received frames should cause RST_STREAM
-    # frames to be emitted. END_STREAM is always carried *by* a frame,
-    # so it should do nothing.
+    # An endpoint that receives any frame other than PRIORIT after receiving
```
Lukasa: `PRIORITY`.

h2/stream.py
```diff
@@ -633,19 +700,33 @@ def send_alt_svc(self, previous_state):
...
-        (H2StreamStateMachine.send_reset, StreamState.CLOSED),
+        (H2StreamStateMachine.recv_on_closed_stream, StreamState.CLOSED),
     (StreamState.CLOSED, StreamInputs.RECV_DATA):
-        (H2StreamStateMachine.send_reset, StreamState.CLOSED),
-    (StreamState.CLOSED, StreamInputs.RECV_PUSH_PROMISE):
-        (H2StreamStateMachine.push_on_closed_stream, StreamState.CLOSED),
+        (H2StreamStateMachine.recv_on_closed_stream, StreamState.CLOSED),
     (StreamState.CLOSED, StreamInputs.RECV_END_STREAM):
-        (None, StreamState.CLOSED),
+        (H2StreamStateMachine.recv_on_closed_stream, StreamState.CLOSED),
```
Lukasa: This should probably stay as `None`. Note the previous comment about `END_STREAM`.

h2/stream.py
```diff
@@ -318,6 +357,34 @@ def push_on_closed_stream(self, previous_state):
...
+
+        if self.stream_closed_by is StreamClosedBy.SEND_RST_STREAM:
+            return []
+        return self.recv_on_closed_stream(previous_state)
+
+    def reset_on_closed_stream(self, previous_state):
+        """
+        Called when a RST_STREAM frame received on an already-closed stream
+
+        If a END_STREAM had just sent for a short period, then just ignore the
```
Lukasa: Again, this one comments about `END_STREAM` but has code for `RST_STREAM`.

test/test_closed_streams.py
```diff
@@ -48,24 +48,24 @@ def test_can_receive_multiple_rst_stream_frames(self, frame_factory):
 
         assert isinstance(event, h2.events.StreamReset)
 
-    def test_closed_stream_resets_further_frames(self, frame_factory):
+    def test_resett_closed_stream_resets_further_frames(self, frame_factory):
```
Lukasa: Nit: `resett` should be `reset`

test/test_closed_streams.py
```diff
@@ -48,24 +48,24 @@ def test_can_receive_multiple_rst_stream_frames(self, frame_factory):
...
 
-        f = frame_factory.build_headers_frame(
+        header_frame = frame_factory.build_headers_frame(
             self.example_request_headers, flags=['END_STREAM']
         )
-        c.receive_data(f.serialize())
-        c.send_headers(
-            stream_id=1,
-            headers=self.example_response_headers,
-            end_stream=True
```
Lukasa: Was there any need to remove this?

test/test_closed_streams.py
```diff
@@ -89,6 +89,31 @@ def test_closed_stream_resets_further_frames(self, frame_factory):
...
+            headers=self.example_response_headers,
+            end_stream=True
+        )
+        c.clear_outbound_data_buffer()
+
+        data_frame = frame_factory.build_data_frame(b'hi there')
+
+        with pytest.raises(h2.exceptions.ProtocolError):
+            c.receive_data(data_frame.serialize())
+
```
Lukasa: This test doesn't actually indicate that a `RST_STREAM` frame is sent: it should look in the outbound data buffer.

Lukasa: Of course, that's because you don't expect a `RST_STREAM` frame, you expect a `GOAWAY` frame. You'll need to change the docstring too. =)

chhsiao90:
@Lukasa - Thanks for your review, I will fix that stuffs and sorry for the so many typos :(

But I just find another critical problem related to the stream state machine garbage collection.
The h2.Connection will clean up the closed stream, so it still had problem when the stream state machine had been set to None.
Is there any good idea for this?

test/test_closed_streams.py
chhsiao90: No, had it or not affect the test result

Lukasa:
> Thanks for your review, I will fix that stuffs and sorry for the so many typos :(

That's ok. =)

> The h2.Connection will clean up the closed stream, so it still had problem when the stream state machine had been set to None.
> Is there any good idea for this?

Yeah, it will do that.

The simplest idea might be to keep a pair of `set` objects that indicate stream IDs that have been closed, and how they were closed. That way quick checks can be done that tell us whether a) the stream never existed at all, b) if it did exist, how it was closed.

That way some higher level code can be responsible for determining how to error out in future.

test/test_closed_streams.py
Lukasa: Ok, so we should put it back. =)

h2/stream.py
chhsiao90: Got it!

h2/stream.py
chhsiao90: Yeah, that was mistake I made, thanks!

chhsiao90:
Almost understand what I should do.
And there is a lot of code need improved here.
I will ping you once I have done. Thanks!

h2/stream.py
```diff
@@ -630,22 +699,36 @@ def send_alt_svc(self, previous_state):
...
         (None, StreamState.CLOSED),
 
-    # While closed, all other received frames should cause RST_STREAM
-    # frames to be emitted. END_STREAM is always carried *by* a frame,
-    # so it should do nothing.
+    # An endpoint that receives any frame other than PRIORITY after receiving
+    # a RST_STREAM MUST treat that as a stream error of type STREAM_CLOSED.
+    # An endpoint that receives any frames after receiving a frame with the
+    # END_STREAM flag set MUST treat that as a connection error of type
+    # STREAM_CLOSED.
```
alexwlchan: Can we get a link/reference to the relevant section of the RFC here?

h2/stream.py
```diff
@@ -206,6 +216,9 @@ def stream_ended(self, previous_state):
         """
         Fires when a stream is cleanly ended.
         """
+        assert previous_state is StreamState.HALF_CLOSED_REMOTE
```
alexwlchan: Better to use == than is, as we only care about equality of these states, not identity.

h2/stream.py
```diff
@@ -269,6 +282,10 @@ def recv_push_promise(self, previous_state):
         event.parent_stream_id = self.stream_id
         return [event]
 
+    def send_end_stream(self, previous_state):
```
chhsiao90: OK!

h2/stream.py
chhsiao90: OK!

h2/stream.py
chhsiao90: No problem, is docstring like ``` In rfc7540 section 5.1 ``` OK?
Or should be more detailed ?

alexwlchan: “RFC” should be capitalised, but otherwise that would be fine.

Example here: https://github.com/chhsiao90/hyper-h2/blob/9146feeae5443bcd5b4526e59bc8585c62c7adf4/h2/stream.py#L803

chhsiao90:
@Lukasa - Another problem related to the RFC

In RFC7540, section 5.1
> An endpoint MUST NOT send frames other than PRIORITY on a closed
      stream.  An endpoint that receives any frame other than PRIORITY
      after receiving a RST_STREAM MUST treat that as a stream error
      (Section 5.4.2) of type STREAM_CLOSED.  Similarly, an endpoint
      that receives any frames after receiving a frame with the
      END_STREAM flag set MUST treat that as a connection error
      (Section 5.4.1) of type STREAM_CLOSED, unless the frame is
      permitted as described below.

It didn't mention how about if the stream is closed by ```a higher stream id is sent/recv before```  from section 5.1.1
> The first use of a new stream identifier implicitly closes all
   streams in the "idle" state that might have been initiated by that
   peer with a lower-valued stream identifier.

Is that need to be asked on the mailing list for clarify?

Lukasa:
No, I don't think so: I think the `END_STREAM` case applies here.

chhsiao90:
OK, I think I just make some little progress on this!
Basic functionality and test coverage.
I think there might be still a lot need improved.

I will keep write test and update doc string.
And if any comment please kindly let me know. Thank you!

h2/stream.py
```diff
@@ -202,10 +212,19 @@ def window_updated(self, previous_state):
...
+        assert previous_state == StreamState.OPEN
+        event = StreamEnded()
+        event.stream_id = self.stream_id
+        return [event]
+
     def stream_ended(self, previous_state):
         """
         Fires when a stream is cleanly ended.
         """
+        assert previous_state == StreamState.HALF_CLOSED_LOCAL
```
Lukasa: It should not be necessary to make these assertions about previous states.

h2/stream.py
```diff
@@ -269,7 +289,23 @@ def recv_push_promise(self, previous_state):
...
+        """
+        assert previous_state == StreamState.HALF_CLOSED_REMOTE
+        self.stream_closed_by = StreamClosedBy.SEND_END_STREAM
+
+    def send_reset_stream(self, previous_state):
+        """
+        Called when an attempt is made to send RST_STREAM at non-closed stream
+        state.
+        """
+        assert previous_state != StreamState.CLOSED
```
Lukasa: No need to make the assertions about previous state.

h2/stream.py
```diff
@@ -269,7 +289,23 @@ def recv_push_promise(self, previous_state):
         event.parent_stream_id = self.stream_id
         return [event]
 
-    def send_reset(self, previous_state):
+    def send_end_stream(self, previous_state):
+        """
+        Called when an attempt is made to send END_STREAM at HALF_CLOSED_REMOTE
+        state.
+        """
+        assert previous_state == StreamState.HALF_CLOSED_REMOTE
```
Lukasa: No need to make the assertions about previous state.

h2/stream.py
```diff
@@ -279,19 +315,40 @@ def send_reset(self, previous_state):
         it's the first time we've done this if the stream is currently in a
         state other than CLOSED.
         """
-        events = []
+        assert previous_state != StreamState.CLOSED
```
Lukasa: We shouldn't need this assertion either.

h2/stream.py
```diff
@@ -279,19 +315,40 @@ def send_reset(self, previous_state):
...
+    def recv_on_closed_stream(self, previous_state):
+        """
+        Called when an unexpected frame is received on an already-closed
+        stream.
+
+        If the stream is closed by RST_STREAM, it will raise StreamClosedError
+        to reset stream. If the stream is closed by END_STREAM, it will raise
+        ProtocolError and close the connection.
+        """
+        assert previous_state == StreamState.CLOSED
```
Lukasa: No need for this assertion either.

h2/stream.py
```diff
@@ -279,19 +315,40 @@ def send_reset(self, previous_state):
...
+        self.stream_closed_by = StreamClosedBy.SEND_RST_STREAM
 
         error = StreamClosedError(self.stream_id)
-        error._events = events
+
+        event = StreamReset()
+        event.stream_id = self.stream_id
+        event.error_code = ErrorCodes.STREAM_CLOSED
+        event.remote_reset = False
+        error._events = [event]
```
Lukasa: I don't think we need to provide this event every time. Is there any reason for you to change that?

h2/stream.py
```diff
@@ -305,7 +362,26 @@ def send_on_closed_stream(self, previous_state):
...
+        Called when a PUSH_PROMISE frame received on an already-closed stream
+
+        If the stream is closed by us with sending RST_STREAM, then we presume
+        that the PUSH_PROMISE was in flight when we reset the parent stream.
+        Rathen than accept the new stream, just reset it.
+        Otherwise, we should call this a PROTOCOL_ERROR: pushing a stream on a
+        naturally closed stream is a real problem because it creates a brand
+        new stream that the remote peer now believes exists
+        """
+        assert previous_state == StreamState.CLOSED
```
Lukasa: No need for state assertions.

h2/stream.py
```diff
@@ -318,6 +394,34 @@ def push_on_closed_stream(self, previous_state):
...
         raise ProtocolError("Attempted to push on closed stream.")
 
+    def window_on_closed_stream(self, previous_state):
+        """
+        Called when a WINDOW_UPDATE frame received on an already-closed stream
+
+        If a END_STREAM had just sent for a short period, then just ignore the
+        frame. Otherwise, follow same rule of recv_on_closed_stream.
+        """
+        assert previous_state == StreamState.CLOSED
```
Lukasa: No need for state assertions.

h2/stream.py
```diff
@@ -269,7 +289,23 @@ def recv_push_promise(self, previous_state):
...
+
+    def send_reset_stream(self, previous_state):
+        """
+        Called when an attempt is made to send RST_STREAM at non-closed stream
+        state.
+        """
+        assert previous_state != StreamState.CLOSED
+        self.stream_closed_by = StreamClosedBy.SEND_RST_STREAM
+
+    def reset_stream(self, previous_state):
```
Lukasa: I think this needs a better name, as it's now misleadingly close to `send_reset_stream`. Something that is close to what it's actually for, which is maybe more like "invalid_frame_received".

test/test_closed_streams.py
```diff
@@ -145,3 +104,199 @@ def test_closed_stream_not_present_in_streams_dict(self, frame_factory):
...
+        c.send_headers(
+            stream_id=1,
+            headers=self.example_response_headers,
+            end_stream=True
+        )
+
+        if clear_streams:
+            # Call open_outbound_streams to force connection clean
+            # closed stream
+            c.open_outbound_streams
```
Lukasa: We should call `open_inbound_streams` here just to be safe: even though both share the same code path at the moment, they may diverge in future and we should try to force the correct behaviour.

h2/stream.py
chhsiao90: OK, all the unnecessary assert will be removed!

test/test_closed_streams.py
chhsiao90: No problem, that was more clear!

h2/stream.py
chhsiao90: Because in current version, this method never be called at CLOSE state.
So the "if condition" never be false and I remove the "if condition" for the coverage.
Should I add a comment on this?

chhsiao90:
@Lukasa - Thanks for the review!
And all comments addressed except rename ```reset_stream``` for a proper name
I will take some times to think about that!
Yes, I think there might be some work variable name and docstrings because of my poor english

h2/stream.py
chhsiao90: Is that ```must_reset_error_stream``` or something like that better ?

chhsiao90:
@Lukasa - Another change I plan to add that I think need your feedback before I change it.

I think ```StreamIDTooLowError``` should had error code ```STREAM_CLOSED```
and it should had same operation at [connection.py#L1470-L1491](https://github.com/python-hyper/hyper-h2/blob/master/h2/connection.py#L1470-L1491) as ```StreamClosedError```

Because ```StreamIDTooLowError``` always be raised because some actions happened at a closed stream like HEADER frame or PUSH_PROMISE frame.
And all error related to stream closed seems to be defined with error code ```STREAM_CLOSED```

So I had 3 options
- Add ```self.error_code = h2.errors.ErrorCodes.STREAM_CLOSED``` at ```StreamIDTooLowError```
- ```class StreamIDTooLowError(StreamClosedError)```
- Removed ```StreamIDTooLowError``` and used ```StreamClosedError``` instead of

WDYT

h2/stream.py
Lukasa: Hrm. `reset_stream_on_error` maybe?

Lukasa:
So, I disagree. Specifically, I think this paragraph of section 5.1 is the governing one:

> In the absence of more specific guidance elsewhere in this document implementations SHOULD treat the receipt of a frame that is not expressly permitted in the description of a state as a connection error of type PROTOCOL_ERROR.

The stream in question was not closed via RST_STREAM or END_STREAM, and therefore is not covered by the more specific guidance elsewhere, so I think it counts as a connection error.

chhsiao90:
OK, so for the case it should be connection error with type ```PROTOCOL_ERROR```,
not connection error with type ```STREAM_CLOSED```

Then, there is another problem.
For current code base, we only record the reset streams when the stream state machine was deleted by garbage collection.
Then there will some problem to distinguish closed stream with
- stream was closed by receive END_STREAM flag
- stream was closed because of a higher stream was opened

Then these two type of closed stream will all treat illegal frame as connection error with type ```PROTOCOL_ERROR```
And that will be another RFC violation for
> An endpoint MUST NOT send frames other than PRIORITY on a closed
      stream.  An endpoint that receives any frame other than PRIORITY
      after receiving a RST_STREAM MUST treat that as a stream error
      (Section 5.4.2) of type STREAM_CLOSED.  Similarly, an endpoint
      that receives any frames after receiving a frame with the
      END_STREAM flag set MUST treat that as a connection error
      (Section 5.4.1) of type STREAM_CLOSED, unless the frame is
      permitted as described below.

Should I also record the end_stream before closed stream cleanup ?
Thanks!

Lukasa:
So the thing to do is probably to change the `reset_streams` set to become a dictionary of `closed_streams`, with the values being the reason they were closed. That way we can keep track of what to do there.

chhsiao90:
OK, thanks!
Just one more confirm! is there any concern about the memory consumption about keep all of state how stream closed ?
Or maybe we can plan improved that in the future, not at this moment.

Lukasa:
> is there any concern about the memory consumption about keep all of state how stream closed ?

Yeah, I am moderately concerned. But short of dropping down a level and writing C code there is no space-efficient method of storing this data. Ideally we'd do it as a bitfield or something similar, which would allow us to cap out at 256 MB, but we can't really do that.

We can attempt to have some heuristics about this, but basically the short version is that for now this is good enough.

chhsiao90:
All comments are done. And thanks for the hard working on code review!

h2/connection.py
```diff
@@ -425,6 +427,7 @@ def _open_streams(self, remainder):
                 to_delete.append(stream_id)
 
         for stream_id in to_delete:
+            self._closed_streams[stream_id] = self.streams[stream_id].closed_by
             del self.streams[stream_id]
```
Lukasa: We can avoid doing two dict lookups here by changing the first line to `stream = self.streams.pop(stream_id)` and then do `self._closed_streams[stream_id] = stream.closed_by`. Moderately more efficient.

h2/connection.py
```diff
@@ -1015,8 +1018,6 @@ def reset_stream(self, stream_id, error_code=0):
         stream = self._get_stream_by_id(stream_id)
         frames = stream.reset_stream(error_code)
         self._prepare_for_sending(frames)
-        self._reset_streams.add(stream_id)
-        del self.streams[stream_id]
```
Lukasa: Why did this get removed? I'm not opposed to removing it, I just want to understand why.

h2/stream.py
```diff
@@ -202,10 +212,16 @@ def window_updated(self, previous_state):
         event.stream_id = self.stream_id
         return [event]
 
+    def stream_half_closed(self, previous_state):
```
Lukasa: This needs a docstring.

h2/stream.py
Lukasa: No, I think that explanation is enough, thank you. =)

h2/stream.py
```diff
@@ -731,6 +824,13 @@ def closed(self):
         """
         return self.state_machine.state == StreamState.CLOSED
 
+    @property
+    def closed_by(self):
+        """
+        Whether the stream is closed by RST_STREAM
```
Lukasa: This docstring isn't right. =)

test/test_closed_streams.py
Lukasa: This still does `outbound`.

h2/connection.py
chhsiao90: Before this PR, H2Connection need record that if it had sent RST_STREAM frame, so that H2Connection can handle receiving frames that just after we reset the stream.
But after this PR, this responsibility we will let H2StreamStateMachine and H2Stream take care for us, so it is no need to record reset_stream when we reset.

chhsiao90:
@Lukasa , sorry, I think I might edit your comment by accident...

Problem at https://github.com/python-hyper/hyper-h2/pull/472#discussion_r101964822
I think that might be fixed at https://github.com/python-hyper/hyper-h2/pull/472/commits/6423616fab1fafe4329479b66d5db550de00525f

I will check for that!
Is that we should call c.open_outbound_streams when H2Connection(client_side=True)
and call c.open_inbound_streams otherwise ?

h2/connection.py
Lukasa: Ok, that makes sense, thanks!

chhsiao90:
@Lukasa - Thanks for the review and cleanup!
And thanks @alexwlchan for the review!

test/test_closed_streams.py
alexwlchan: Are we still putting this test back, or are we happy it’s been superseded by other tests?

(I’ve read this code several times now, and I think I’m too tired to be sure either way.)

chhsiao90: This test is moved to class ```TestStreamClosedByRstStream``` at https://github.com/chhsiao90/hyper-h2/blob/handle-frames-on-closed-stream/test/test_closed_streams.py#L276-L323

Because currently we had closed stream that was closed by RST_STREAM or END_STREAM. So I separate them into different class.

And sorry for this massive PR that bring inconvenience to you :-(

alexwlchan: > This test is moved to class TestStreamClosedByRstStream at https://github.com/chhsiao90/hyper-h2/blob/handle-frames-on-closed-stream/test/test_closed_streams.py#L276-L323

Ah yes, I see it now. 👍 

> And sorry for this massive PR that bring inconvenience to you

Please don’t apologise – it’s great to have other people contributing, and have code to review. :-)

Sorry if I gave the impression of snappiness or annoyance – it’s been a long week, but that’s no excuse for bluntness.

chhsiao90: Haha, I am just feel embarrassed at my PR.
Think that I will do better next time!

test/test_closed_streams.py
```diff
@@ -145,3 +104,281 @@ def test_closed_stream_not_present_in_streams_dict(self, frame_factory):
...
+        ]
+    )
+    @pytest.mark.parametrize("clear_streams", [True, False])
+    def test_frames_after_recv_end_will_error(self,
+                                              frame_factory,
+                                              frame,
+                                              clear_streams):
+        """
+        A stream that is closed by receive END_STREAM will raise
+        ProtocolError when received unexpected frame.
```
chhsiao90: Thanks for the grammar correction!

h2/stream.py
```diff
@@ -292,6 +321,27 @@ def send_reset(self, previous_state):
         error._events = events
         raise error
 
+    def recv_on_closed_stream(self, previous_state):
+        """
+        Called when an unexpected frame is received on an already-closed
+        stream.
+
+        If the stream is closed by RST_STREAM, it will raise StreamClosedError
+        to reset stream. If the stream is closed by END_STREAM, it will raise
```
chhsiao90: Cannot find this in latest code, might be fixed by @Lukasa

chhsiao90:
@alexwlchan - Thanks for the review and comments were addressed by https://github.com/python-hyper/hyper-h2/pull/472/commits/08ec35488247c25fe5f0200f1028b6cbf6bd00ae .

h2/stream.py
```diff
@@ -279,19 +315,40 @@ def send_reset(self, previous_state):
...
+        ProtocolError and close the connection.
+        """
+        assert previous_state == StreamState.CLOSED
+        assert self.stream_closed_by is not None
+
+        if self.stream_closed_by in (StreamClosedBy.RECV_RST_STREAM,
+                                     StreamClosedBy.SEND_RST_STREAM):
+            raise StreamClosedError(self.stream_id)
+        else:
+            raise ProtocolError(
```
chhsiao90: I'm not sure that it's worth or not to add assert here.
I'm fine for with assert here or not!

Lukasa:
Great job @chhsiao90, this was a very big change to make and you did it very well. I think you have a lot to be proud of: there are not many contributors who would dive in so thoroughly, and you did it well and persevered through a bunch of code review, so I'm very impressed.

Also thanks so much for reviewing @alexwlchan, you're a star. I don't know what I'd do without you. :grin: Congratulations to you both, this is a fantastic improvement to land!

chhsiao90:
I must say, english and grammar is my biggest enemy LOL

Lukasa:
:wink: That's quite alright, so long as you're willing for us to go in and fix it then it doesn't matter that much: your English is more than good enough for us to understand what you're trying to say, which is what matters most.

PR #475, opened 2017-03-07
Update flow control diagrams.

Lukasa:
We changed our state machines in #472. This PR updates the images accordingly.


alexwlchan:
Is the source code for these diagrams checked in? Would make it much easier to see the diffs. 😄

Lukasa:
Yes: it's the two state machines. ;) The code is available in the `visualize` directory.

PR #476, opened 2017-03-07
Add extension frame support.

### LINKED ISSUE #435: Ignore unrecognised frames.
RFC 7540 says:

> Implementations MUST ignore unknown or unsupported values in all extensible protocol elements.  Implementations MUST discard frames that have unknown or unsupported types.

Currently hyper-h2 contravenes this policy: it throws exceptions when it encounters unsupported frame types. This is not a good plan, especially as extension frames are in active development and could be encountered on the open web.

A good question would be: how do we handle these frames? One option is to define a new event (`UnexpectedFrameReceived`) that simply hands the `hyperframe` frame object up the stack to allow the calling code to do whatever it needs. Another option is to simply swallow all unexpected frames and not emit them to user code.

I lean towards doing the first, but I'd like to know what @Kriechi or @mhils think, as they're the most likely to be interested in actually accessing the unexpected frame data.

Lukasa:
This PR adds support for emitting events when we receive extension frames.

Resolves #435.

The first few commits in here should be backported to older versions of hyper-h2 to allow them to use the 5.0 release series of hyperframe if needed.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -23,7 +23,7 @@ from .errors import ErrorCodes, _error_code_from_int
 from .events import (
     WindowUpdated, RemoteSettingsChanged, PingAcknowledged,
     SettingsAcknowledged, ConnectionTerminated, PriorityUpdated,
-    AlternativeServiceAvailable,
+    AlternativeServiceAvailable, UnknownFrameReceived
 )
 from .exceptions import (
     ProtocolError, NoSuchStreamError, FlowControlError, FrameTooLargeError,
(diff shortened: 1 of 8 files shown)


Lukasa:
Specifically, the first two commits are the ones that need backporting.

h2/connection.py
```diff
@@ -1922,6 +1927,23 @@ def _receive_alt_svc_frame(self, frame):
...
+        We have received a frame that we do not understand. This is almost
+        certainly an extension frame, though it's impossible to be entirely
+        sure.
+
+        RFC 7540 § 5.5 says that we MUST ignore unknown frame types: so we
+        do. We do notify the user that we received one, however.
+        """
+        # All we do here is log.
+        self.config.logger.debug(
+            "Received unknown extension frame (ID 0x%x)", frame.stream_id
```
alexwlchan: Elsewhere we log the stream ID as a `%d` integer. Should we be consistent in the way we log?

Lukasa: Hrm, probably, yeah.

alexwlchan: FWIW, I would probably lean towards doing logging in hex as you did it originally, but let’s make that a separate patch.

alexwlchan:
Okay, LGTM.

@Lukasa I’ll let you merge, so you can sort out backports and the like.

Lukasa:
Cool, thanks @alexwlchan! I'm going to try to write this in as a shell script that we can commit to the repo so that backporting gets a bit easier.

Issue #497, opened 2017-03-22
Errors with cookie handling during decompression.


Lukasa:
Spotted by @hawkowl.

RFC 7540 Section 8.1.2.5 reads as follows:

> To allow for better compression efficiency, the Cookie header field MAY be split into separate header fields, each with one or more cookie-pairs.  If there are multiple Cookie header fields after decompression, these MUST be concatenated into a single octet string using the two-octet delimiter of 0x3B, 0x20 (the ASCII string "; ") before being passed into a non-HTTP/2 context, such as an HTTP/1.1 connection, or a generic HTTP server application.

We are very much in violation of that MUST right now. This is breaking some Twisted applications that work with multiple cookies, whereby cookies are getting lost because Twisted (rightly) expects only a single cookie header.

For the moment, this issue tracks only the MUST. Later on we may want to tackle the other side of this (the MAY).

My belief is that the last sentence of that paragraph makes this a hyper-h2 issue, not a hpack issue: HTTP/2 contexts may want to see those headers individually, but in general we shouldn't show them to users that way. This means we need to add another step to the inbound header normalization pipeline that grabs all cookie fields and concatenates them together.

Lukasa:
This bug also affects a number of previous versions, so will want to be backported to the current backport releases.

Lukasa:
Actually, ok, I don't think we can backport this. We previously haven't normalized inbound header fields in this way, so starting to do so is absolutely a breaking change unless it defaults to "off". That default is extremely not helpful, so it would probably be better just to call this a 3.0 type change and bump the major version.

hawkowl:
I don't feel it's really a breaking change if it's a straight violation of
the spec. As a user, I'd count this as a bug fix. Plus, the separated and
joined cookie values are semantically the same, and some HTTP/2 clients can
send them ;-joined anyway, so your server would need to understand it
anyway.

Lukasa:
@hawkowl So it depends on the user. In particular, mitmproxy are heavy users of hyper-h2 and need the ability to get hold of the unedited headers as they were received from the remote peer. That means that this needs to be controlled by a switch, and if I just shove out a release that changes this behaviour then mitmproxy will break.

[...]

PR #498, opened 2017-03-22
Dedent over-indented test.

Lukasa:
Spotted this while I was working on #497.

### DIFF
--- a/test/test_basic_logic.py
+++ b/test/test_basic_logic.py
@@ -1699,21 +1699,21 @@ def sanity_check_data_frame(data_frame,
                             expected_flow_controlled_length,
                             expect_padded_flag,
                             expected_data_frame_pad_length):
-        """
-        ``data_frame`` is a frame of type ``hyperframe.frame.DataFrame``,
-        and the ``flags`` and ``flow_controlled_length`` of ``data_frame``
-        match expectations.
-        """
+    """
+    ``data_frame`` is a frame of type ``hyperframe.frame.DataFrame``,
+    and the ``flags`` and ``flow_controlled_length`` of ``data_frame``
+    match expectations.
+    """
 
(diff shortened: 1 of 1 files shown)


test/test_basic_logic.py
```diff
@@ -1699,21 +1699,21 @@ def sanity_check_data_frame(data_frame,
...
-             expected_flow_controlled_length)
-        )
-        if expect_padded_flag:
-            assert 'PADDED' in data_frame.flags
-        else:
-            assert 'PADDED' not in data_frame.flags
+    assert(
+        (data_frame.flow_controlled_length ==
+            expected_flow_controlled_length)
+    )
```
alexwlchan: With the extra dedenting, this assertion fits on a single line.

Lukasa:
Thanks @alexwlchan, I've addressed that.

test/test_basic_logic.py
```diff
@@ -1699,21 +1699,20 @@ def sanity_check_data_frame(data_frame,
...
-            (data_frame.flow_controlled_length ==
-             expected_flow_controlled_length)
-        )
-        if expect_padded_flag:
-            assert 'PADDED' in data_frame.flags
-        else:
-            assert 'PADDED' not in data_frame.flags
+    assert (
+        data_frame.flow_controlled_length == expected_flow_controlled_length
+    )
```
alexwlchan: Sorry, I should have been clearer. You can pull it onto a single line without the parentheses, i.e.,

```python
    assert data_frame.flow_controlled_length == expected_flow_controlled_length
```

Lukasa: Ah...yes. I thought I tried that and it didn't work, but clearly I'm an idiot. That's what happens when you're sick for a week. :(

PR #499, opened 2017-03-22
Fix up cookie handling.

### LINKED ISSUE #497: Errors with cookie handling during decompression.

Lukasa:
Resolves #497.

This is a breaking change, and I'd describe it as a moderately urgent one, so I'm hereby pronouncing that our next release will be 3.0. I'll be opening issues shortly to handle removing all deprecated functionality that needs to be stripped.

### DIFF
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -1064,8 +1064,9 @@ class H2Stream(object):
             headers = validate_headers(headers, hdr_validation_flags)
 
         if header_encoding:
-            headers = list(_decode_headers(headers, header_encoding))
-        events[0].headers = headers
+            headers = _decode_headers(headers, header_encoding)
+
+        events[0].headers = list(headers)
         return [], events
 
     def remotely_pushed(self, pushed_headers):
@@ -1118,9 +1119,11 @@ class H2Stream(object):
             headers = validate_headers(headers, hdr_validation_flags)
 
         if header_encoding:
-            headers = list(_decode_headers(headers, header_encoding))
+            headers = _decode_headers(headers, header_encoding)
 
-        events[0].headers = headers
+        # The above steps are all generators, so we need to concretize the
+        # headers now.
+        events[0].headers = list(headers)
         return [], events
 
     def receive_data(self, data, end_stream, flow_control_len):
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -1056,17 +1056,9 @@ class H2Stream(object):
         events[0].pushed_stream_id = promised_stream_id
 
         hdr_validation_flags = self._build_hdr_validation_flags(events)
-
-        if self.config.normalize_inbound_headers:
-            headers = normalize_inbound_headers(headers, hdr_validation_flags)
-
-        if self.config.validate_inbound_headers:
-            headers = validate_headers(headers, hdr_validation_flags)
-
-        if header_encoding:
-            headers = _decode_headers(headers, header_encoding)
-
-        events[0].headers = list(headers)
+        events[0].headers = self._process_received_headers(
+            headers, hdr_validation_flags, header_encoding
+        )
         return [], events
 
     def remotely_pushed(self, pushed_headers):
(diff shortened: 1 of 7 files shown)


h2/stream.py
```diff
@@ -1316,6 +1309,30 @@ def _build_headers_frames(self,
         frames[-1].flags.add('END_HEADERS')
         return frames
 
+    def _process_received_headers(self,
+                                  headers,
+                                  header_validation_flags,
+                                  header_encoding):
+        """
+        When headers have been received from the remote peer, runs a processing
```
alexwlchan: nit: "run a processing pipeline"?

test/test_basic_logic.py
```diff
@@ -850,6 +850,85 @@ def test_changing_max_frame_size(self, frame_factory, frame_size):
...
+
+        cookie_field_count = sum(1 for n, _ in e.headers if n == 'cookie')
+        assert cookie_field_count == 1
+
+        for n, v in e.headers:
+            if n == 'cookie':
+                assert v == expected
+                break
+        else:
+            pytest.mark.fail("Unable to check cookie header")
```
alexwlchan: When would we ever come down the `else:` branch of this loop?

If we pass the assert above, we know that there’s exactly one header whose name is `cookie`. So unless `e.headers` changes in the second iteration, we’ll always run into the `break` statement.

How about something like this?

```python
cookie_fields = [(n, v) for (n, v) in e.headers if n == 'cookie']
assert len(cookie_fields) == 1

_, v = cookie_fields[0]
assert v == expected
```

test/test_basic_logic.py
```diff
@@ -850,6 +850,85 @@ def test_changing_max_frame_size(self, frame_factory, frame_size):
...
+        events = c.receive_data(f.serialize())
+
+        assert len(events) == 1
+        e = events[0]
+
+        cookie_field_count = sum(1 for n, _ in e.headers if n == 'cookie')
+        assert cookie_field_count == 3
+
+        received_cookies = [(n, v) for n, v in e.headers if n == 'cookie']
+        assert cookie_headers == received_cookies
```
alexwlchan: Could we tidy this up in a similar way?

```python
received_cookies = [(n, v) for n, v in e.headers if n == 'cookie']

assert len(received_cookies) == 3
assert cookie_headers == received_cookies
```

h2/stream.py
```diff
@@ -1105,14 +1102,10 @@ def receive_headers(self, headers, end_stream, header_encoding):
...
-            headers = validate_headers(headers, hdr_validation_flags)
-
-        if header_encoding:
-            headers = list(_decode_headers(headers, header_encoding))
-
-        events[0].headers = headers
+        hdr_validation_flags = self._build_hdr_validation_flags(events)
+        events[0].headers = self._process_received_headers(
+            headers, hdr_validation_flags, header_encoding
+        )
```
alexwlchan: 👍

Lukasa:
Thanks for that feedback @alexwlchan! I think I've addressed it in the last two commits.

Lukasa:
Thanks @alexwlchan! :heart:

PR #500, opened 2017-03-22
Remove deprecated stuff: round one.

Lukasa:
This is an initial removal of the deprecated code within hyper-h2. There is one more change we'd like to make: we'd like to make the default representation of headers bytes, not unicode. That will come along in a separate PR.


Lukasa:
And rightly so. ;)

Lukasa:
So, the twisted master failures are unfortunate. Naturally, twisted master fails because this is a breaking change. However, they pin out our new breaking version. Not sure what to do about that: should we temporarily move it to "allowed failures"?

alexwlchan:
> So, the twisted master failures are unfortunate. Naturally, twisted master fails because this is a breaking change. However, they pin out our new breaking version. Not sure what to do about that: should we temporarily move it to "allowed failures"?

That feels a bit icky – especially as the whole point of this change is to fix things for twisted, it would be good to have some guarantee of that before releasing 3.0 and claiming it’s fixed.

I notice that our tox.ini [installs twisted from a GitHub repo](https://github.com/python-hyper/hyper-h2/blob/master/tox.ini#L19), not PyPI. We could fork twisted, bump the h2 version requirement in the fork and point tox at the fork – that allows us to check everything works correctly with our changes. Then when the version gets bumped in twisted master, we revert the tox change.

I haven’t had enough sleep to know if that’s a brilliant or a terrible idea.

alexwlchan:
Alternatively, it looks like the vast majority of twisted errors come from one line, where it uses deprecated parameters that’s being removed by this patch: https://github.com/twisted/twisted/blob/trunk/src/twisted/web/_http2.py#L125

We could just fix this on the twisted side to use `H2Configuration`?

alexwlchan:
I’ve opened https://github.com/twisted/twisted/pull/750 to fix this behaviour on twisted master. Once that’s merged, we should be able to re-run the twisted tests and (hopefully!) get a pass.

Lukasa:
Cool, good idea. Handily, I'm a Twisted committer, so we can get this through pretty nicely. :grin:

Lukasa:
Ok let's rebuild these.

Lukasa:
Looks like there are some other problems in that Twisted code too. ;) @alexwlchan, are you interested in making a follow-up patch to Twisted? If not, I'll do it.

alexwlchan:
> Looks like there are some other problems in that Twisted code too. ;) @alexwlchan, are you interested in making a follow-up patch to Twisted?

Yep, I’ll do it later.

alexwlchan:
See https://twistedmatrix.com/trac/ticket/9092#ticket

alexwlchan:
See https://github.com/twisted/twisted/pull/751 for what I think is the remaining twisted deprecations.

alexwlchan:
Okay, I’ve kicked off another round of twisted tests.

Lukasa:
🍿

Lukasa:
Looks like some stuff got missed. :smile:

Lukasa:
Hurrah, looks good! Thanks for the work today @alexwlchan. :sparkles:

PR #502, opened 2017-03-24
Change default header representation to bytes.

### LINKED ISSUE #154: Consider having received headers be bytes, not unicode strings.

Lukasa:
Resolves #154.

After this, we're very close to being ready to go for 3.0.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -19,6 +19,10 @@ API Changes (Backward-Incompatible)
 - Removed deprecated ``client_side`` and ``header_encoding`` properties from
   ``H2Connection``.
 - ``dict`` objects are no longer allowed for user-supplied headers.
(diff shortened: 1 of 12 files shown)


Lukasa:
Yeah, handily when we had to solve this the first time (to allow bytes through) almost all the problems went away.

PR #505, opened 2017-03-24
[WIP] Add h2spec testing.

### LINKED ISSUE #503: Add h2spec to CI.
It would be really nice if one of the CI tests we ran involved running h2spec. This is a moderately tricky thing to do because it means we need to expand one of our example servers to be good enough to pass h2spec, but that's not the end of the world.

Lukasa:
This PR would add CI testing of hyper-h2 using h2spec.

This is a WIP right now because, surprise surprise, we don't pass h2spec. However, I want this code present in a PR so people can evaluate it and suggest enhancements. This also has a problem in that tox doesn't fail if h2spec fails: this is being tracked as summerwind/h2spec#77.

Resolves #503.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,4 +39,5 @@ matrix:
 install:
   - _travis/install.sh
 script:
+  - export PATH=$PATH:$PWD/bin/
   - tox -- --hypothesis-profile travis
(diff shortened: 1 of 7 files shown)


Lukasa:
To be clear, I want to ship 3.0 even though we don't pass h2spec. However, we should make bugfixes to start passing h2spec and then merge this, and backport them where appropriate.

pgjones:
Fixed with f3239182792434d8bca22c349107faa3b4e16c75 and merged.

PR #508, opened 2017-03-27
Naked CONTINUATION frames must cause GOAWAY in all cases.

Lukasa:
Spotted with h2spec.

Our logic for handling frames on closed streams was overriding the logic for naked CONTINUATION frames. RFC 7540 requires that naked CONTINUATION frames always be treated as connection errors, but we were treating them as stream errors that boiled down to "receiving on a closed stream". That's wrong, so we need to fix it.


Lukasa:
Ah, wrong branch. Damnit.

PR #509, opened 2017-03-27
Naked CONTINUATION frames must cause GOAWAY.

Lukasa:
Spotted with h2spec.

Our logic for handling frames on closed streams was overriding the logic for naked CONTINUATION frames. RFC 7540 requires that naked CONTINUATION frames always be treated as connection errors, but we were treating them as stream errors that boiled down to "receiving on a closed stream". That's wrong, so we need to fix it.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,11 @@ API Changes (Backward-Compatible)
 Bugfixes
 ~~~~~~~~
 
+- CONTINUATION frames sent on closed streams previously caused stream errors
+  of type STREAM_CLOSED. RFC 7540 § 6.10 requires that these be connection
+  errors of type PROTOCOL_ERROR, and so this release changes to match that
+  behaviour.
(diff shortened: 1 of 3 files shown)


HISTORY.rst
```diff
@@ -10,6 +10,10 @@ API Changes (Backward-Compatible)
 Bugfixes
 ~~~~~~~~
 
+- CONTINUATION frames sent on closed streams previously caused stream errors
+  of type STREAM_CLOSED. RFC 7540 requires that these be connection errors of
```
alexwlchan: Am I right in thinking this is RFC 7540 § 6.10? Took me a minute or two to find – worth making the reference more specific?

Lukasa: Yeah, sorry, I'll update that. :smile:

Lukasa: Addressed.

PR #512, opened 2017-03-27
Turn flow control errors on the stream into stream-level errors.

### LINKED ISSUE #507: We emit GOAWAY(FLOW_CONTROL_ERROR) when we should emit RST_STREAM(FLOW_CONTROL_ERROR)
Found using h2spec.

When the remote peer sends WINDOW_UPDATE frames that take a stream's flow control window above `2^31 - 1`, we should send a RST_STREAM frame with error code FLOW_CONTROL_ERROR on that stream. Right now, we don't do that: instead, we GOAWAY the entire connection.

It would be better to move to h2spec's view of the world and save the connection. This will change our behaviour from emitting a ProtocolError exception to emitting a StreamReset event.

Lukasa:
Resolves #507.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -10,6 +10,10 @@ API Changes (Backward-Compatible)
 Bugfixes
 ~~~~~~~~
 
+- Remote peers incrementing their inbound connection window beyond the maximum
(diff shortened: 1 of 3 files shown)


alexwlchan:
Rebase against master and this is good to go.

Issue #510, opened 2017-03-27
We need to tighten up handling of pseudo-header fields.


Lukasa:
We have a number of failures in h2spec caused by being overly lenient with pseudo-header fields. We need to resolve the following problems:

- [ ] When we receive a HEADERS frame that contains a pseudo-header field defined for response only (that is, `:status`), we MUST respond with a stream error of type PROTOCOL_ERROR. Same policing needs to be applied to sending. (#514)
- [x] When we receive a HEADERS frame with an empty `:path` pseudo-header field, we MUST respond with a stream error of type PROTOCOL_ERROR. Same policing needs to be applied to sending. (#513)
- [x] When we receive a HEADERS frame that omits the `:method` pseudo-header field we MUST respond with a stream error of type PROTOCOL_ERROR. The same policing needs to be applied to sending. (#513)
- [x] When we receive a HEADERS frame that omits the `:scheme` pseudo-header field we MUST respond with a stream error of type PROTOCOL_ERROR. The same policing needs to be applied to sending. (#513)
- [x] When we receive a HEADERS frame that omits the `:path` pseudo-header field we MUST respond with a stream error of type PROTOCOL_ERROR. The same policing needs to be applied to sending. (#513)

All of these say stream error, but hyper-h2 has generally treated errors like this as connection errors, and so we may well continue down that route here.

PR #513, opened 2017-03-27
Improve handling of request pseudo header fields.

Lukasa:
This resolves most of the points in #510 by validating that the request pseudo-header fields are present-and-correct in request headers.


Lukasa:
@alexwlchan I'm going to add the inverse checks (things that must be absent from header blocks) in a separate PR. :D

Lukasa:
Changelog added.

PR #514, opened 2017-03-27
Prevent pseudo-headers appearing in incorrect contexts.

### LINKED ISSUE #510: We need to tighten up handling of pseudo-header fields.

Lukasa:
Resolves #510.

This prevents `:status` appearing in request header blocks and the various request pseudo-headers appearing in response header blocks.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -14,6 +14,8 @@ Bugfixes
   any of the mandatory pseudo-header fields (:path, :scheme, and :method).
 - h2 now rejects receiving and sending request header blocks that have an empty
   :path pseudo-header value.
+- h2 now rejects receiving and sending request header blocks that contain
+  response-only pseudo-headers, and vice versa.
 
 
 3.0.0 (2017-03-24)
--- a/h2/utilities.py
+++ b/h2/utilities.py
@@ -42,6 +42,18 @@ _SECURE_HEADERS = frozenset([
     b'proxy-authorization', u'proxy-authorization',
 ])
 
+
+_REQUEST_ONLY_HEADERS = frozenset([
+    b':scheme', u':scheme',
+    b':path', u':path',
+    b':authority', u':authority',
+    b':method', u':method'
+])
+
+
+_RESPONSE_ONLY_HEADERS = frozenset([b':status', u':status'])
+
+
 if sys.version_info[0] == 2:  # Python 2.X
     _WHITESPACE = frozenset(whitespace)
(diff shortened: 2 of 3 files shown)


HISTORY.rst
```diff
@@ -14,6 +14,8 @@ Bugfixes
   any of the mandatory pseudo-header fields (:path, :scheme, and :method).
 - h2 now rejects receiving and sending request header blocks that have an empty
   :path pseudo-header value.
+- h2 now rejects receiving and sending request header blocks that contain
+  response-only pseudo-headers, and vice versa
```
alexwlchan: nit: missing full stop

Lukasa:
> A thought: we have a lot of repetitive tests in `test_validate_headers.py`. Would it be worth parametrising a bit further, maybe pulling out the validation method as a parameter to the test?

Heh, I tried. However, the two validation functions actually have different valid inputs: in particular, the *inbound* validation functions cannot handle unicode headers, while the *outbound* ones can. That means that you need to parametrize across both simultaneously, which py.test has very little in the way of nice APIs to do. 😞

HISTORY.rst
Lukasa: Addressed. :smile:

Lukasa:
Thanks for the review @alexwlchan! :sparkles:

PR #525, opened 2017-03-29
Backport #513 to the 2.5 release series.

Lukasa:
See #513 for more.


Lukasa:
Ok, this patch won't backport to 2.5.X. That seems to be because the 2.5 branch has some bugs around validating pushes that didn't have fixes backported, presumably to avoid breaking stuff. So I'm just going to accept that if you want this bug fix (and likely the next one I backport) you'll need to update to the 2.6 series at the very least.

PR #535, opened 2017-03-29
Better police dynamic table size updates.

### LINKED ISSUE #511: We must prevent a dynamic table size update exceeding the value of SETTINGS_HEADER_TABLE_SIZE.
Spotted using h2spec.

We may want to do this by adding this as a property in hpack and then plumbing support through, but this is really a bug in the combined service.

See RFC 7541 § 6.3 for more.

Lukasa:
Resolves #511.


alexwlchan:
🎉

Issue #250, opened 2016-07-07
Support sending memoryview objects in DATA frames.


Lukasa:
Encountered this as part of working on Twisted's HTTP/2 client.

When writing large amounts of data in a non-prioritised way (like a client would!), it's entirely possible that a stack will want to send > SETTINGS_MAX_FRAME_SIZE data in one go. This will require that the stack split the write up into multiple chunks. For large single writes or small values of SETTINGS_MAX_FRAME_SIZE this represents quite a lot of copying, which is substantially less than ideal.

It would be better, then, if we could support `memoryview` objects in DATA frames for as long as possible before concretizing to bytes. This may require some additional support in hyperframe, as ideally we'd only concretize the memoryview when we serialize the frame.

An alternative is to allow the `send_data` call to automatically chunkify your writes. That would let the `memoryview` use be entirely internal. I'm inclined not to want to do that, but it's worth noting that the option exists.

fredthomsen:
I am interested in this and looking at the code seems like this is still outstanding.  I'll start looking into this.

fredthomsen:
To avoid this unnecessary copying, it seems the the `DataFrame` would need to have a member variable which was a `memoryview` object. However, it seems that that runs counter to statement you made above in terms of delaying the conversion to the `memoryview` until the frame is serialized:

> This may require some additional support in hyperframe, as ideally we'd only concretize the memoryview when we serialize the frame.

fredthomsen:
@Lukasa also you mention that you would prefer `memoryview` use to be external rather than internal; however, why not both?  I understand your preference for internal over external.  If a `memoryview` instance is passed in, then great, the user is handling it as they should; however, would you not also want the ability to do some buffering internally converting a `bytes` object to a `memoryview` then creating as many `DataFrame` s as necessary?

Lukasa:
No, sorry: "concretize" means "convert to bytes". We're in agreement. 

As to the other stuff: it may be worth doing. But memoryview objects aren't free to create, so we may not be able to come up with a good heuristic for when to use them and when not to. Hard to be sure. So my instinct is to solve one problem at a time.

fredthomsen:
Ok, I misread and am on the same page now; keep things as a `memoryview` as long as possible.

In terms of the cost of a `memoryview`, I don't follow.  I understand they are not free; however, they certainly are cheaper than the repeated copying of `bytes`.

Lukasa:
[...]

PR #541, opened 2017-04-03
Initial memory view work/idea

fredthomsen:
Work for #250.

I am probably being rather dense, but I was having a tough time with this one so I thought I would just put up where I was going with it.  I expect a good smack on the head more than anything else :face_with_head_bandage:

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -6,6 +6,7 @@ h2/connection
 An implementation of a HTTP/2 connection.
 """
 import base64
+import math
 
 from enum import Enum, IntEnum
 
@@ -877,7 +878,7 @@ class H2Connection(object):
         :param stream_id: The ID of the stream on which to send the data.
         :type stream_id: ``int``
         :param data: The data to send on the stream.
-        :type data: ``bytes``
+        :type data: ``memoryview``
         :param end_stream: (optional) Whether this is the last data to be sent
             on the stream. Defaults to ``False``.
         :type end_stream: ``bool``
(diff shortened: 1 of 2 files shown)


h2/connection.py
```diff
@@ -894,27 +895,25 @@ def send_data(self, stream_id, data, end_stream=False, pad_length=None):
...
         if pad_length is not None:
             if not isinstance(pad_length, int):
                 raise TypeError("pad_length must be an int")
             if pad_length < 0 or pad_length > 255:
                 raise ValueError("pad_length must be within range: [0, 255]")
             # Account for padding bytes plus the 1-byte padding length field.
-            frame_size += pad_length + 1
+            padding_per_frame = pad_length + 1
+        num_frames = math.ceil(data_size / float(self.max_outbound_frame_size))
+        data_size += padding_per_frame * num_frames
```
Lukasa: So this set of changes is unnecessary: we don't split data up into multiple frames. That's considered a user-level concern.

h2/stream.py
```diff
@@ -983,20 +984,26 @@ def send_data(self, data, end_stream=False, pad_length=None):
...
+            df.data = data[frame_num * self.max_outbound_frame_size:frame_num+1 * self.max_outbound_frame_size]
+            if pad_length is not None:
+                df.flags.add('PADDED')
+                df.pad_length = pad_length
+
+            # Subtract flow_controlled_length to account for possible padding
+            self.outbound_flow_control_window -= df.flow_controlled_length
+            assert self.outbound_flow_control_window >= 0
+
+            dfs.append(df)
```
Lukasa: Similarly here: we don't need to do this.

fredthomsen:
@Lukasa thanks for the feedback, and I'll do down that route.

Issue #466, opened 2017-02-13
Add support for property-based tests for Settings equality.


Lukasa:
Follows on from #464 and #465.

We've added some specific tests for the `Settings` object's equality behaviour, and have reached some minimum code coverage, but this is really something that hypothesis should be able to help us with.

In particular, it should be reasonably possible to construct a Hypothesis strategy that will produce `Settings` objects that we can then evaluate. This is not an easy thing to do as it requires some familiarity with Hypothesis to pull off, but is probably a worthwhile challenge for anyone who wants to get their hands dirty with Hypothesis.

PR #543, opened 2017-04-10
Hypothesis h2 settings equality

fredthomsen:
WIP for equality and property test using hypothesis to address #466.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -111,4 +111,5 @@ In chronological order:
 - Fred Thomsen (@fredthomsen)
 
   - Added logging.
-
+  - Enhance equality testing of ``h2.settings.Settings`` objects with
+    ``hypothesis``.
--- a/test/test_settings.py
+++ b/test/test_settings.py
@@ -12,7 +12,9 @@ import h2.exceptions
 import h2.settings
 
 from hypothesis import given, assume
-from hypothesis.strategies import integers
+from hypothesis.strategies import (
+    integers, booleans, fixed_dictionaries, builds
+)
 
 
 class TestSettings(object):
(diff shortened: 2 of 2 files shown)


test/test_settings.py
```diff
@@ -333,7 +333,7 @@ def test_cannot_set_invalid_values_for_max_frame_size(self, val):
             assert e.value.error_code == h2.errors.ErrorCodes.PROTOCOL_ERROR
             assert s[h2.settings.SettingCodes.MAX_FRAME_SIZE] == 16384
 
-    @given(integers())
+    @given(strategies.integers())
```
Lukasa: What's the reasoning for this change?

test/test_settings.py
```diff
@@ -367,104 +367,90 @@ class TestSettingsEquality(object):
...
+                h2.settings.SettingCodes.HEADER_TABLE_SIZE,
+                h2.settings.SettingCodes.ENABLE_PUSH,
+                h2.settings.SettingCodes.INITIAL_WINDOW_SIZE,
+                h2.settings.SettingCodes.MAX_FRAME_SIZE,
+                h2.settings.SettingCodes.MAX_CONCURRENT_STREAMS,
+                h2.settings.SettingCodes.MAX_FRAME_SIZE,
+                h2.settings.SettingCodes.MAX_HEADER_LIST_SIZE,
+            ]),
+            values=strategies.integers(min_value=0))
+        )
```
Lukasa: So this strategy generally seems ok, except that it'll frequently generate invalid initial settings that the settings object will reject. We'll need to expand this out to ensure that we only generate valid initial settings (or that we discard any attempt to create an invalid one).

test/test_settings.py
```diff
@@ -367,104 +367,90 @@ class TestSettingsEquality(object):
...
         Two objects that are equal to each other do not compare not equal to
         each other using the != operator.
         """
-        a = self.an_instance()
-        b = self.an_instance()
+        b = copy.deepcopy(a)
         assert not (a != b)
 
-    def test_different_eq(self):
+    @given(a=SettingsStrategy, b=SettingsStrategy)
```
Lukasa: So it's entirely possible that this would generate two settings objects that *are* identical. We need to be a bit careful here and construct a test that allows us to ensure that the two objects are different. In this case, the easiest way might be to create two strategies that cannot generate overlapping Settings objects in some way, or at least create a Settings object that our strategy cannot generate. Of course, if @alexwlchan has more useful things to suggest here that'd be really helpful (he's our resident Hypothesis expert).

alexwlchan:
ACK: I’ve seen this and will review it, but I probably won’t have time for a proper review until later this week.

Lukasa:
:heart: Thanks @alexwlchan.

test/test_settings.py
fredthomsen: I did this since I was using more functions within the `strategies` submodule along with `integers`.  Probably  just a nervous tick I have 😁 .  I am happy to conform to whatever style you deem appropriate ie change this back and just import all the functions from the strategies submodule that I use.

test/test_settings.py
fredthomsen: Yup.  Hadn't gotten there yet, but this is next.

test/test_settings.py
fredthomsen: Very true.  I'll play around with this some more and see what advice @alexwlchan has.

test/test_settings.py
Lukasa: I don't think it matters. 😉 Just curious.

test/test_settings.py
```diff
@@ -367,104 +367,90 @@ class TestSettingsEquality(object):
...
+                h2.settings.SettingCodes.MAX_FRAME_SIZE,
+                h2.settings.SettingCodes.MAX_CONCURRENT_STREAMS,
+                h2.settings.SettingCodes.MAX_FRAME_SIZE,
+                h2.settings.SettingCodes.MAX_HEADER_LIST_SIZE,
+            ]),
+            values=strategies.integers(min_value=0))
+        ) 
+
+    @given(o=SettingsStrategy)
+    def test_identical_eq(self, o):
```
alexwlchan: Also, a minor thing: can we pick variable names that are a bit more descriptive please?  😄

fredthomsen:
Believe all your comments are addressed now.

test/test_settings.py
```diff
@@ -367,125 +369,65 @@ class TestSettingsEquality(object):
...
-        each other using the != operator.
-        """
-        a = self.an_instance()
-        b = self.an_instance()
-        assert not (a != b)
+    SettingsStrategy = builds(
+        h2.settings.Settings,
+        client=booleans(),
+        initial_values=fixed_dictionaries({
+            h2.settings.SettingCodes.HEADER_TABLE_SIZE: integers(0),
```
alexwlchan: Hypothesis also has a `just()` strategy, which returns a single value every time. Mildly more efficient, IIRC, although I don't think it makes a massive difference.

test/test_settings.py
```diff
@@ -367,125 +369,65 @@ class TestSettingsEquality(object):
...
+            h2.settings.SettingCodes.ENABLE_PUSH: integers(0, 1),
+            h2.settings.SettingCodes.INITIAL_WINDOW_SIZE:
+                integers(0, 2**31 - 1),
+            h2.settings.SettingCodes.MAX_FRAME_SIZE: integers(16384, 16777215),
```
alexwlchan: Maybe replace these two with the constants they're derived from? i.e.

```python
h2.settings.SettingCodes.MAX_FRAME_SIZE: integers(2**14, 2**24 - 1)
```

test/test_settings.py
```diff
@@ -367,125 +369,65 @@ class TestSettingsEquality(object):
...
         """
-        a = self.an_instance()
-        b = self.another_instance()
-        assert not (a == b)
+        assert (settings == settings)
+        assert not (settings != settings)
 
-    def test_different_ne(self):
+    @given(settings=SettingsStrategy, o_settings=SettingsStrategy)
+    def test_equality_multiple(self, settings, o_settings):
```
alexwlchan: Maybe `other_settings`? We’re not tight for characters.

test/test_settings.py
```diff
@@ -367,125 +369,65 @@ class TestSettingsEquality(object):
...
-        Two objects that are not equal to each other do not compare equal to
-        each other using the == operator.
+        An object compares equal to itself using the == operator and the !=
+        operator.
         """
-        a = self.an_instance()
-        b = self.another_instance()
-        assert not (a == b)
+        assert (settings == settings)
+        assert not (settings != settings)
```
alexwlchan: I’ve realised since writing my last comment that this isn’t always the best test, because you’re comparing the identity of two identical objects. Python will default to using identity as an equality test, so this passes even if you don't supply custom equality methods:

```pycon
>>> class Foo: pass
...
>>> x = Foo()
>>> y = Foo()
>>> x == x
True
>>> x == y
False
>>> x != x
False
```

My preferred approach is now to test based on copies of objects, i.e.,

```python
other_settings = copy.deepcopy(settings)
assert (settings == other_settings)
assert not (settings != other_settings)
```

although this does rely on you being able to copy your objects, and I can’t remember what the options are here for `IntEnum` and its subclasses.

test/test_settings.py
```diff
@@ -367,125 +369,65 @@ class TestSettingsEquality(object):
...
-        class Delegate(object):
             def __ne__(self, other):
                 return [self]
 
-        a = self.an_instance()
-        b = Delegate()
-        assert (a != b) == [b]
+        delg = Delegate()
+        assert (settings == delg) == [delg]
+        assert (settings != delg) == [delg]
```
alexwlchan: I realise we nicked this test from elsewhere, but would it be better to tweak the return values from `__eq__` and `__ne__` so we can be sure we're really calling the right method?

e.g.

```python
def __eq__(self, other):
    return [self]

def __ne__(self, other):
    return [self, self]
```

CONTRIBUTORS.rst
```diff
@@ -111,4 +111,5 @@ In chronological order:
 - Fred Thomsen (@fredthomsen)
 
   - Added logging.
-
+  - Enchance equality testing of ``h2.settings.Settings`` objects with
```
alexwlchan: /s/Enchance/Enhance

test/test_settings.py
fredthomsen: Not sure I understand where you want to apply the `just` strategy here.  If you are referring to this specific line, the value listed here is just the minimum value.

test/test_settings.py
fredthomsen: Ok, then it seems like this test makes little sense then as we have another where two strategies are generated, so I'll just nix this one.

test/test_settings.py
fredthomsen: How embarrassing 😁 .  Yes definitely.

fredthomsen: Why is this worthy of it's own test btw?  Can you enlighten me how this is not a given?

test/test_settings.py
alexwlchan: Oh yeah, oops, I was misreading `integers(0)` as `integers(min_size=0, max_size=0)`.

Ignore me.

fredthomsen:
@alexwlchan Yes, and to your point I restored that test.  As you said, it will rarely go down that path of being equal when there are multiple settings generated from the strategy.

test/test_settings.py
alexwlchan: Err…

This comes from the PyOpenSSL test suite, a module which used to have a lot more C in its core lib than Python, so these things weren’t generally a given. I’m not sure it’s as useful for hyper-h2, and wouldn’t object if you chose to remove it.

Lukasa:
Cool, shall we bring it up to date with the current master first?

fredthomsen:
Branch has been rebased onto latest master.

test/test_settings.py
```diff
@@ -367,125 +370,67 @@ class TestSettingsEquality(object):
...
+        client=booleans(),
+        initial_values=fixed_dictionaries({
+            h2.settings.SettingCodes.HEADER_TABLE_SIZE: integers(0),
+            h2.settings.SettingCodes.ENABLE_PUSH: integers(0, 1),
+            h2.settings.SettingCodes.INITIAL_WINDOW_SIZE:
+                integers(0, 2**31 - 1),
+            h2.settings.SettingCodes.MAX_FRAME_SIZE:
+                integers(2**14, 2**24 - 1),
+            h2.settings.SettingCodes.MAX_CONCURRENT_STREAMS: integers(0),
+            h2.settings.SettingCodes.MAX_HEADER_LIST_SIZE: integers(0),
```
Lukasa: Are we worried about creating values outside the valid range of some of these settings? These settings do have upper limits on their valid values.

fredthomsen: Yes.  I looked at the specs and saw that these are 32 bit fields and thus have adjusted them accordingly.

In this case we should update the `_validate_settings` function as well.

CONTRIBUTORS.rst
Lukasa: Still probably want to deal with this. :grin:

fredthomsen:
Fixed misspelling in contributors file.  I had made that change earlier, but I must have inadvertently blown it away.

alexwlchan:
🎉

PR #594, opened 2017-04-25
Memory view in data frame

### LINKED ISSUE #250: Support sending memoryview objects in DATA frames.

fredthomsen:
Looks like there was no work needed to resolve #250. This just contains the unit test and bumping the required version of hyperframe.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
     install_requires=[
-        'hyperframe>=5.0, <6',
+        'hyperframe>=5.1.0, <6',
         'hpack>=2.3,<4',
     ],
     extras_require={
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,6 +7,9 @@ Release History
 API Changes (Backward-Compatible)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -62,7 +62,7 @@
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
     install_requires=[
-        'hyperframe>=5.0, <6',
+        'hyperframe>=5.0.1, <6',
```
alexwlchan: I think you want 5.1.0, not 5.0.1?

fredthomsen: Yup.  Thanks.

PR #835, opened 2017-07-30
refactor ping-related events

### LINKED ISSUE #834: API changes to Ping events
Similar to https://github.com/python-hyper/wsproto/pull/36, I would appreciate bringing the h2 API & events into a similar state as wsproto.

At the moment h2 fires a `PingAcknowledged` event when it receives a PING frame with the ACK flag. But it does not fire an event for receiving an initial PING.

I would propose introducing a new event: `PingReceived`,
and renaming the existing `PingAcknowledged` to `PingAckReceived`.
Both events should carry the `frame.opaque_data` as `ping_data` attribute.

For backward-compatibility `PingAcknowledged` should be an empty subclass of `PingAckReceived`, and be marked as deprecated.

Kriechi:
fixes #834

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,6 +9,11 @@ API Changes (Backward-Compatible)
 
 - ``h2.connection.H2Connection.send_data`` now supports ``data`` parameter
   being a ``memoryview`` object.
+- Refactor ping-related events: a ``h2.events.PingReceived`` event is fired
+  when a PING frame is received and a ``h2.events.PingAckReceived`` event is
+  fired when a PING frame with an ACK flag is received.
+  ``h2.events.PingAcknowledged`` is deprecated in favour of the identical
+  ``h2.events.PingAckReceived``.
 
 Bugfixes
 ~~~~~~~~
--- a/h2/events.py
+++ b/h2/events.py
@@ -356,20 +356,49 @@ class RemoteSettingsChanged(Event):
         )
 
 
-class PingAcknowledged(Event):
+class PingReceived(Event):
     """
-    The PingAcknowledged event is fired whenever a user-emitted PING is
-    acknowledged. This contains the data in the ACK'ed PING, allowing the
-    user to correlate PINGs and calculate RTT.
+    The PingReceived event is fired whenever a PING is received. It contains
+    the 'opaque data' of the PING frame. A ping acknowledgment with the same
+    'opaque data' is automatically emitted after receiving a ping.
+
+    .. versionadded:: 3.1.0
     """
     def __init__(self):
         #: The data included on the ping.
         self.ping_data = None
 
     def __repr__(self):
-        return "<PingAcknowledged ping_data:%s>" % (
+        return "<PingReceived ping_data:%s>" % (
             _bytes_representation(self.ping_data),
         )
+
+
+class PingAcknowledged(Event):
+    """
+    Same as PingAckReceived.
+
+    .. deprecated:: 3.1.0
+    """
+    def __init__(self):
+        #: The data included on the ping.
+        self.ping_data = None
+
+    def __repr__(self):
+        return "<PingAckReceived ping_data:%s>" % (
+            _bytes_representation(self.ping_data),
+        )
+
+
+class PingAckReceived(PingAcknowledged):
+    """
+    The PingAckReceived event is fired whenever a PING acknowledgment is
+    received. It contains the 'opaque data' of the PING+ACK frame, allowing the
+    user to correlate PINGs and calculate RTT.
+
+    .. versionadded:: 3.1.0
+    """
+    pass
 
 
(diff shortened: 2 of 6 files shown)


h2/events.py
```diff
@@ -356,22 +356,50 @@ def __repr__(self):
...
+        return "<PingAckReceived ping_data:%s>" % (
             _bytes_representation(self.ping_data),
         )
 
 
+class PingAcknowledged(PingAckReceived):
+    """
+    PingAcknowledged: same as PingAckReceived
+
+    .. deprecated:: 3.1.0
```
alexwlchan: Maybe emit a DeprecationWarning from the `__init__` method?

test/test_basic_logic.py
```diff
@@ -1341,7 +1346,7 @@ def test_ping_frame_opaque_data_must_be_length_8_bytestring(self,
 
     def test_receiving_ping_acknowledgement(self, frame_factory):
         """
-        Receiving a PING acknolwedgement fires a PingAcknolwedged event.
+        Receiving a PING acknolwedgement fires a PingAckReceived event.
```
alexwlchan: nit: while we’re here, can we fix the spelling of acknolwedgement?

HISTORY.rst
```diff
@@ -9,6 +9,9 @@ API Changes (Backward-Compatible)
 
 - ``h2.connection.H2Connection.send_data`` now supports ``data`` parameter
   being a ``memoryview`` object.
+- Refactor ping-related events: a ``h2.events.PingReceived`` event is fired
+  when a PING frame is received and a ``h2.events.PingAckReceived`` event is
+  fired when a PING frame with an ACK flag is received.
```
alexwlchan: Maybe include a note about the deprecation/replacement of `PingAcknowledged`?

test/test_basic_logic.py
Kriechi: Ah - I already found one typo. Guess there are more easter eggs to be found 😄

HISTORY.rst
Kriechi: 👍 
Done.

test/test_basic_logic.py
Kriechi: 👍 
Done.

h2/events.py
Kriechi: I didn't because of f0dd593bf6eebdf5b81822edbd7a2e99af242ce0, and I can't remember hyper-h2 having used them for previous deprecated things.
But if you want one, I won't block it.

HISTORY.rst
```diff
@@ -9,6 +9,11 @@ API Changes (Backward-Compatible)
...
+  ``h2.events.PingAcknowledged`` is deprecated in favour of the identical
+  ``h2.events.PingAckReceived``.
```
Lukasa: I just really struggle to get on board with this name change, I don't think the value it adds outweighs the code churn required to get this to function sensibly. 🙁

h2/events.py
```diff
@@ -356,22 +356,50 @@ def __repr__(self):
         )
 
 
-class PingAcknowledged(Event):
+class PingReceived(Event):
     """
-    The PingAcknowledged event is fired whenever a user-emitted PING is
-    acknowledged. This contains the data in the ACK'ed PING, allowing the
+    The PingReceived event is fired whenever a PING is received. It contains
+    the 'opaque data' of the PING frame.
```
Lukasa: Can we add a note here that this event requires no action on the part of the user, as hyper-h2 automatically replies to PING frames?

h2/events.py
```diff
@@ -356,22 +356,50 @@ def __repr__(self):
...
         self.ping_data = None
 
     def __repr__(self):
-        return "<PingAcknowledged ping_data:%s>" % (
+        return "<PingAckReceived ping_data:%s>" % (
             _bytes_representation(self.ping_data),
         )
 
 
+class PingAcknowledged(PingAckReceived):
```
Lukasa: I think this doesn't work. If you are doing `if isinstance(frame, PingAcknowledged)` in your current code, this will now fail to work with `PingAckReceived`. I think your class hierarchy is backwards.

Lukasa: This is also a good indication that if we're proceeding with the rename, we need to have a test to confirm that both names work. 😉

Kriechi:
Should be all fixed now. Thanks for pointing out the inheritance issue!

alexwlchan:
> My only question is whether the code churn for the event names is worth it. @alexwlchan, my view is that it's not: what do you think?

I’d lean towards changing the names – `PingAcknowledged` does does a bit like hyper-h2 is doing something, when it’s not done anything.

Lukasa:
Hrm, fair enough. Ok then, let's call this good. :smile: Thanks @Kriechi!

PR #977, opened 2017-08-27
s/amt/amount/

Kriechi:
Those few extra characters are not that expensive.
This brings the data_to_send API closer to wsproto.


Kriechi:
Done.

PR #1031, opened 2017-10-14
[documentation] Note about NPN and ALPN support on Python < 2.7.9

vladmunteanu:
Given the lack of support for NPN and ALPN on Python versions prior to 2.7.9, I thought this should be reflected in the documentation.

I'd like to point out that [RFC 7540](https://tools.ietf.org/html/rfc7540#section-3.3) specifies only ALPN for negotiating the HTTP/2 version.
I'm more than happy to make the adjustments in the documentation (remove NPN), if you see fit.

### DIFF
--- a/docs/source/negotiating-http2.rst
+++ b/docs/source/negotiating-http2.rst
@@ -5,17 +5,17 @@ Negotiating HTTP/2
 
 .. _starting-alpn:
 
-HTTPS URLs (ALPN and NPN)
+HTTPS URLs (ALPN)
 -------------------------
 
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this! Let’s restrict ourselves to ALPN for now: NPN is basically dead, and while mentioning it now is good for completeness in two years it’ll make this document seem really weird.

vladmunteanu:
Should i remove NPN from the rest of the paragraphs as well? Like i said, RFC 7540 only mentions ALPN: https://tools.ietf.org/html/rfc7540#section-3.3

Lukasa:
Mind bringing this up to date with the current master branch?

vladmunteanu:
@Lukasa Done!

Lukasa:
Thanks! :sparkles:

PR #1107, opened 2018-01-19
Logic error fixed, closes #1106

### LINKED ISSUE #1106: Bugs in the Server Plaintext Upgrade example
The Server Plaintext Upgrade in https://python-hyper.org/projects/h2/en/stable/negotiating-http2.html uses twice "if match is not None" when it should be "if match is None" since we do want these two headers to be present (just tested).

bortzmeyer:


Lukasa:
Per our [one of the team](http://python-hyper.org/en/latest/one-of-the-team.html) policy, I've sent you an invite to the project @bortzmeyer!

PR #1146, opened 2018-03-29
Add trio example HTTP/2 echo server

standy66:
Added example of HTTP/2 echo server that uses [trio](https://trio.readthedocs.io/en/latest/) async library.


Kriechi:
Hi! Thanks for sending this new example. Would you mind rebasing to the current master and giving it another look at and manual test. There have been some changes - just want to make sure your example still works the same as you intended.
We recently deprecated a bunch of stuff and Python versions - so I hope this makes it easier!

tomchristie:
I'd suggest we can safely close this off as stale.

PR #1164, opened 2018-08-01
Proof of concept for refactor closed streams

hellysmile:


hellysmile:
It fails right now only tests for `test/test_closed_streams.py` which seems are used only there

PR #1177, opened 2019-01-18
Minor fixes and improvements

pgjones:
These hopefully put the master branch in a good state to add new features and fixes too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: python
 
 branches:
@@ -15,28 +14,39 @@ matrix:
   include:
     # Main tests on supported Python versions.
     - python: "2.7"
+      dist: trusty
       env: TOXENV=py27
     - python: "3.4"
+      dist: trusty
       env: TOXENV=py34
     - python: "3.5"
+      dist: trusty
       env: TOXENV=py35
     - python: "3.6"
+      dist: trusty
       env: TOXENV=py36
+    - python: "3.7"
+      dist: xenial
+      env: TOXENV=py37
     - python: "pypy-5.3.1"
+      dist: trusty
       env: TOXENV=pypy
 
     # Linting, docs, and other non-test stuff.
-    - python: "3.4"
+    - python: "3.7"
+      dist: xenial
       env: TOXENV=lint
-    - python: "3.5"
+    - python: "3.7"
+      dist: xenial
       env: TOXENV=docs
-    - python: "2.7"
+    - python: "3.7"
+      dist: xenial
       env: TOXENV=packaging
 
     # Test we haven't broken our major dependencies.
     - python: "2.7"
+      dist: trusty
       env: TOXENV=py27-twistedMaster
-
 install:
   - "pip install -U pip setuptools"
   - "pip install -U tox"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -13,32 +12,20 @@ cache:
...
+      dist: trusty
+    - python: 3.7
+      dist: xenial
+    - python: pypy-5.3.1
+      dist: trusty
 install:
-  - "pip install -U pip setuptools"
-  - "pip install -U tox"
+  - pip install -U pip setuptools
+  - pip install -U tox-travis
```
Kriechi: I'm aware of this integration - but for some reason I don't remember I always avoided it and used the more explicit (verbose) system that you now removed...

If we switch over here, we should also switch it for wsproto (and possible other projects).
hyperlink uses the explicit form. 
hyperframe uses tox-travis.
h11 doesn't use tox (we probably should...)
mitmproxy uses the explicit form.

.travis.yml
```diff
@@ -13,32 +12,20 @@ cache:
...
+    - python: 3.4
+      dist: trusty
+    - python: 3.5
+      dist: trusty
+    - python: 3.6
+      dist: trusty
+    - python: 3.7
+      dist: xenial
+    - python: pypy-5.3.1
+      dist: trusty
```
Kriechi: Are we not running the docs tox env anymore?

.travis.yml
```diff
@@ -1,4 +1,3 @@
-sudo: false
```
Kriechi: is this the default now - or are we running on the slower VMs now?
I think the main point was that sudo:false gives us a speed boost because TravisCI uses faster containers instead of a full VM.

.travis.yml
pgjones: tox-travis does it for you (it runs all the tox-envs it can given the environment) so see the 3.7 CI job.

.travis.yml
pgjones: I think so, it has been deprecated (see https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)

.travis.yml
pgjones: Ah, I didn't realise there were reasons to avoid `tox-travis`. I'm happy to switch them and go around updating the CI setup whilst I'm at it. hyperframe is because I changed it recently, I'm happy to switch to whichever but prefer `tox-travis` as less CI stuff to figure out.

.travis.yml
Kriechi: oh interesting! 👍

.travis.yml
Kriechi: Yes - I don't remember the reason - I just remember avoiding it... ¯\_(ツ)_/¯
Lets use tox-travis everywhere if we can - and see if it gives us troubles.

.travis.yml
Kriechi: I don't see the docs env being run anywhere... https://travis-ci.org/python-hyper/hyper-h2/jobs/481283520

Maybe that was the reason why I avoided tox-travis, because it doesn't fully run the environments if you don't specify them explicitly.

pgjones: Ah, you are right - I had it configured incorrectly (and never noticed this over the last couple of years on another project). I've changed it now, however I think the matrix and tox is better. Shall I get rid of tox-travis?

pgjones: Updated to use tox as before, `tox-travis` is worse - I'll fix Hyperframe.

PR #1191, opened 2019-06-14
Fixed H2StreamStateMachine to ignore certain frames in the closed state

### LINKED ISSUE #1183: StreamClosedError when receiving RST_STREAM on a stream closed by remote
I am trying to make h2 state machine behave itself correctly when communicating with gRPC Core peers. Right now h2 raises `StreamClosedError` when receiving *RST_STREAM* on a closed stream that was closed by remote, e.g.

```
# Local (h2) -> Remote (gRPC Core)
-> HEADERS frame
-> DATA frame with END_STREAM flag set  # stream is now half-closed (local)
<- HEADERS frame
<- DATA frame
<- HEADERS frame with END_STREAM flag set  # stream is now closed
<- RST_STREAM
```

Unfortunately, this is what gRPC Core does for unary RPC calls, because it cannot accept more data than a single message, delivered in one or possibly multiple *DATA* frames. I am not sure if it's prohibited the the RFC or not, but it is what it is. The workaround I am using for now is to [monkey patch h2's stream transition dictionary](https://github.com/standy66/purerpc/blob/anyio/src/purerpc/grpclib/connection.py#L21) so it ignores any future *RST_STREAM* and *WINDOW_UPDATE* events on a remotely closed stream.

vmagamedov:
This PR fixes state transitions in `H2StreamStateMachine` to ignore `WINDOW_UPDATE` and `RST_STREAM` frames when stream is in the closed state. Also added tests to reproduce these scenarios.

Quote from RFC 7540 (https://tools.ietf.org/html/rfc7540#section-5.1):
```
closed:
      The "closed" state is the terminal state.
      ...
      WINDOW_UPDATE or RST_STREAM frames can be received in this state
      for a short period after a DATA or HEADERS frame containing an
      END_STREAM flag is sent.  Until the remote peer receives and
      processes RST_STREAM or the frame bearing the END_STREAM flag, it
      might send frames of these types.  Endpoints MUST ignore
      WINDOW_UPDATE or RST_STREAM frames received in this state, though
      endpoints MAY choose to treat frames that arrive a significant
      time after sending END_STREAM as a connection error
      (Section 5.4.1) of type PROTOCOL_ERROR.
      ...
```
@pgjones @Lukasa please review changes

Fixes #1183

### DIFF
--- a/h2/stream.py
+++ b/h2/stream.py
@@ -378,41 +378,6 @@ class H2StreamStateMachine(object):
         """
         raise ProtocolError("Attempted to push on closed stream.")
 
-    def window_on_closed_stream(self, previous_state):
-        """
-        Called when a WINDOW_UPDATE frame is received on an already-closed
-        stream.
-
-        If we sent an END_STREAM frame, we just ignore the frame, as instructed
-        in RFC 7540 Section 5.1. Technically we should eventually consider
-        WINDOW_UPDATE in this state an error, but we don't have access to a
-        clock so we just always allow it. If we closed the stream for any other
-        reason, we behave as we do for receiving any other frame on a closed
-        stream.
-        """
-        assert self.stream_closed_by is not None
-
-        if self.stream_closed_by == StreamClosedBy.SEND_END_STREAM:
-            return []
-        return self.recv_on_closed_stream(previous_state)
-
-    def reset_on_closed_stream(self, previous_state):
-        """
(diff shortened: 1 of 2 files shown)


pgjones:
This looks good to me. I've merged manually as I accidentally created a merge commit on this PR.

vmagamedov:
Thanks!

PR #1192, opened 2019-07-07
Remove twisted CI testing

pgjones:
This has been broken for a while (due to a Twisted update), and I've
insufficient Twisted knowledge to fix it. I understand from @Lukasa
that this can safely be removed.

I'll leave this open for at least a week in case someone else knows better.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,6 @@ matrix:
       dist: xenial
       env: TOXENV=packaging
 
-    # Test we haven't broken our major dependencies.
-    - python: "2.7"
-      dist: trusty
-      env: TOXENV=py27-twistedMaster
 install:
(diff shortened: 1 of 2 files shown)


alexwlchan:
I have no Twisted knowledge either, so I can’t help on that front!

Is it worth leaving it as an item in the build matrix with “allow_failures” (see [Travis docs](https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail)), so we’ll remember it was once there and needs addressing, but without it turning all the builds red?

Or do we think it’s unlikely we’ll ever go back to fix it?

pgjones:
@alexwlchan thanks, it is a good suggestion but I'm going to go ahead and merge as I think this is better tested by twisted and I therefore don't plan to fix.

PR #1195, opened 2019-07-20
[requires.io] dependency update on master branch

glyph:


pgjones:
Not this can't pass whilst Python 2 is supported.

glyph:
@pgjones Sounds like there should be an `rq.filter` in there somewherE?

pgjones:
@glyph thanks, I've added in 25743dc855f6c15eb5a8ce3120c94dbec71a83a8.

PR #1211, opened 2020-01-21
auto-emit WINDOW_UPDATE frames on unprocessed DATA frames

### LINKED ISSUE #1210: Flow control: WINDOW_UPDATE swallowed on closed streams
Imagine the following scenario:

* client sends `HEADERS` for a request + `end_stream` to the server
* server sends `HEADERS` for response and starts sending a few `DATA` frames
* client continuously sends `WINDOW_UPDATE` for the consumed data
* server sends more `DATA` frames
* client sends `RST_STREAM` to cancel this stream
* client has a few unprocessed `DATA` frames in it's incoming buffer
* and the server might have already sent a few more before having canceled the stream

The problem here seems to be that the client thinks the stream is already closed. Receiving `DATA` frames on a closed streams causes a `h2.exceptions.StreamClosedError`
https://github.com/python-hyper/hyper-h2/blob/306609aa72c5ad4860a72fc4caa5db9ef5ab5b26/h2/stream.py#L317

This exception is then swallowed and replaced with more generic error handling
https://github.com/python-hyper/hyper-h2/blob/306609aa72c5ad4860a72fc4caa5db9ef5ab5b26/h2/connection.py#L1487

This means the client is completely unaware that it just tried to process a `DATA` frame, yet it did!
Since there was no proper `DataReceived` event for the user to handle, nobody will call `acknowledge_received_data` to update the flow control and send a `WINDOW_UPDATE`.

Now we have effectively starved the whole h2 connection because one "rogue" stream was taking all the available connection flow window and nobody is able to update it because the information about the `DATA` frame is swallowed by the exception handling.

*A short PoC script*
It wraps one of the `H2Connection` functions to add a debug print for the root cause and it prints the remaining flow window size. After a few seconds this will go to 0, and therefore halting all streams in the connection.

The expected output of the script is a continuous output of `DATA` frames of stream 1, every second or so, with 61 bytes, and then many smaller ones for the additional requests.

If you call the script without any CLI args, it only create the first continuous stream, just to demontrate that the `acknowledge_received_data` works in principle.

If you call the script with a single CLI arg `foo`, the script will generate a few more requests, and then send a `RST_STREAM` shortly after. This causes a lot of unprocessed `DATA` frames to be swallowed, and the input flow window will shortly after be starved to 0.

```
#!/usr/bin/env python3

import certifi
import h2.connection
import h2.events

import traceback
import sys
import errno
import socket
import ssl
import time

SERVER_NAME = 'http2.golang.org'

socket.setdefaulttimeout(15)

class FoobarH2Connection(h2.connection.H2Connection):
    def _receive_data_frame(self, frame):
        try:
            return super()._receive_data_frame(frame)
        except Exception as e:
            print("Exception detected - swallowing the DataFrame and flow_controlled_length:", e.__dict__, c._inbound_flow_control_window_manager.current_window_size)
            raise e

c = FoobarH2Connection()
c.initiate_connection()

ctx = ssl.create_default_context(cafile=certifi.where())
ctx.set_alpn_protocols(['h2', 'http/1.1'])

s = socket.create_connection((SERVER_NAME, 443))
s = ctx.wrap_socket(s, server_hostname=SERVER_NAME)

s.sendall(c.data_to_send())

headers = [
    (':method', 'GET'),
    (':path', '/clockstream'),
    (':authority', SERVER_NAME),
    (':scheme', 'https'),
]
c.send_headers(1, headers, end_stream=True)
s.sendall(c.data_to_send())

count = 2
while True:
    data = s.recv(65536 * 1024)
    if not data:
        break

    count += 1
    if count % 2 == 1 and len(sys.argv) > 1:
        headers = [
            (':method', 'GET'),
            (':path', '/file/gopher.png'),
            (':authority', SERVER_NAME),
            (':scheme', 'https'),
        ]
        c.send_headers(count, headers, end_stream=True)
        s.sendall(c.data_to_send())
        time.sleep(0.25)
        c.reset_stream(count)

    events = c.receive_data(data)
    for event in events:
        if isinstance(event, h2.events.DataReceived):
            print(event, c._inbound_flow_control_window_manager.current_window_size)
            c.acknowledge_received_data(event.flow_controlled_length, event.stream_id)

    s.sendall(c.data_to_send())
```

The output on my system is:
```
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 3, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 49151
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 3, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 47867
<DataReceived stream_id:1, flow_controlled_length:1180, data:23207e314b42206f66206a756e6b20746f20666f> 46687
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 3, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 46687
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 5, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 30303
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 5, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 29019
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 5, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 29019
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 7, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 12635
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 7, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 11351
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 7, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 11351
Exception detected - swallowing the DataFrame and flow_controlled_length: <class 'h2.exceptions.StreamClosedError'> {'stream_id': 9, 'error_code': <ErrorCodes.STREAM_CLOSED: 5>, '_events': []} 0
...
<after 15 seconds we fail as expected with a socket.timeout: The read operation timed out>
```

ref https://github.com/mitmproxy/mitmproxy/issues/3780

Kriechi:
fixes #1210

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1450,7 +1450,7 @@ class H2Connection(object):
         :returns: A list of events that the remote peer triggered by sending
             this data.
         """
-        self.config.logger.debug(
+        self.config.logger.trace(
             "Process received data on connection. Received data: %r", data
         )
 
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1634,6 +1634,35 @@ class H2Connection(object):
 
         return frames, events + stream_events
 
+    def _handle_data_on_closed_stream(self, events, exc, frame):
+        # This stream is already closed - and yet we received a DATA frame.
+        # The received DATA frame counts towards the connection flow window.
+        # We need to manually to acknowledge the DATA frame to update the flow
+        # window of the connection. Otherwise the whole connection stalls due
+        # the inbound flow window being 0.
+        frames = []
+        conn_manager = self._inbound_flow_control_window_manager
+        conn_increment = conn_manager.process_bytes(
+            frame.flow_controlled_length
+        )
+        if conn_increment:
+            f = WindowUpdateFrame(0)
+            f.window_increment = conn_increment
+            frames.append(f)
+            self.config.logger.debug(
+                "Received DATA frame on closed stream %d - "
+                "auto-emitted a WINDOW_UPDATE by %d",
+                frame.stream_id, conn_increment
+            )
+        f = RstStreamFrame(exc.stream_id)
+        f.error_code = exc.error_code
+        frames.append(f)
+        self.config.logger.debug(
+            "Stream %d already CLOSED or cleaned up - "
+            "auto-emitted a RST_FRAME" % frame.stream_id
+        )
+        return frames, events + exc._events
+
     def _receive_data_frame(self, frame):
         """
         Receive a data frame on the connection.
@@ -1646,12 +1675,19 @@ class H2Connection(object):
         self._inbound_flow_control_window_manager.window_consumed(
             flow_controlled_length
         )
-        stream = self._get_stream_by_id(frame.stream_id)
-        frames, stream_events = stream.receive_data(
-            frame.data,
-            'END_STREAM' in frame.flags,
-            flow_controlled_length
-        )
+
+        try:
+            stream = self._get_stream_by_id(frame.stream_id)
+            frames, stream_events = stream.receive_data(
+                frame.data,
+                'END_STREAM' in frame.flags,
+                flow_controlled_length
+            )
+        except StreamClosedError as e:
+            # This stream is either marked as CLOSED or already gone from our
+            # internal state.
+            return self._handle_data_on_closed_stream(events, e, frame)
+
         return frames, events + stream_events
 
     def _receive_settings_frame(self, frame):
--- a/setup.py
+++ b/setup.py
@@ -58,18 +58,19 @@ setup(
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy',
     ],
     install_requires=[
         'hyperframe>=5.2.0, <6',
-        'hpack>=2.3,<4',
+        'hpack>=3.0,<4',
     ],
     extras_require={
-        ':python_version == "2.7" or python_version == "3.3"': ['enum34>=1.1.6, <2'],
+        ':python_version == "2.7"': ['enum34>=1.1.6, <2'],
     }
 )
--- a/test/test_stream_reset.py
+++ b/test/test_stream_reset.py
@@ -70,8 +70,8 @@ class TestStreamReset(object):
                                                      other_id,
                                                      frame_factory):
         """
-        A stream that has been reset still affects the connection flow control
-        window.
+        A stream that has been reset does not affect the connection flow
+        control window.
         """
         c = h2.connection.H2Connection()
         c.initiate_connection()
@@ -89,19 +89,19 @@ class TestStreamReset(object):
         c.clear_outbound_data_buffer()
 
         f = frame_factory.build_data_frame(
-            data=b'some data!',
+            data=b'some data',
             stream_id=close_id
         )
-        events = c.receive_data(f.serialize())
+        c.receive_data(f.serialize())
 
-        rst_frame = frame_factory.build_rst_stream_frame(
-            close_id, h2.errors.ErrorCodes.STREAM_CLOSED
-        )
-        assert not events
-        assert c.data_to_send() == rst_frame.serialize()
+        expected = frame_factory.build_rst_stream_frame(
+            stream_id=close_id,
+            error_code=h2.errors.ErrorCodes.STREAM_CLOSED,
+        ).serialize()
+        assert c.data_to_send() == expected
 
         new_window = c.remote_flow_control_window(stream_id=other_id)
-        assert initial_window - len(b'some data!') == new_window
+        assert initial_window - len(b'some data') == new_window
 
     @pytest.mark.parametrize('clear_streams', [True, False])
     def test_reset_stream_automatically_resets_pushed_streams(self,
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ matrix:
     - python: "3.7"
       dist: xenial
       env: TOXENV=py37
+    - python: "3.8"
+      dist: xenial
+      env: TOXENV=py38
     - python: "pypy-5.3.1"
       dist: trusty
       env: TOXENV=pypy
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -4,6 +4,19 @@ Release History
 3.1.1dev0
 ---------
 
+Bugfixes
+~~~~~~~~
+
+- Receiving DATA frames on closed (or reset) streams now properly emit a
+  WINDOW_UPDATE to keep the connection flow window topped up.
+
+API Changes (Backward-Incompatible)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- ``h2.config.logger`` now uses a `trace(...)` function, in addition
+  to `debug(...)`. If you defined a custom logger object, you need to handle
+  these new function calls.
+
 
 3.1.1 (2019-08-02)
 ------------------
--- a/h2/config.py
+++ b/h2/config.py
@@ -29,7 +29,7 @@ class DummyLogger(object):
     """
     An Logger object that does not actual logging, hence a DummyLogger.
 
-    For the class the log operation is merely a no-op.  The intent is to avoid
+    For the class the log operation is merely a no-op. The intent is to avoid
     conditionals being sprinkled throughout the hyper-h2 code for calls to
     logging functions when no logger is passed into the corresponding object.
     """
@@ -37,6 +37,12 @@ class DummyLogger(object):
         pass
 
     def debug(self, *vargs, **kwargs):
+        """
+        No-op logging. Only level needed for now.
+        """
+        pass
+
+    def trace(self, *vargs, **kwargs):
         """
         No-op logging. Only level needed for now.
         """
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
+[tool:pytest]
+testpaths = test
+
 [wheel]
 universal = 1
--- a/test/test_closed_streams.py
+++ b/test/test_closed_streams.py
@@ -195,7 +195,6 @@ class TestStreamsClosedByEndStream(object):
     @pytest.mark.parametrize(
         "frame",
         [
-            lambda self, ff: ff.build_data_frame(b'hello'),
             lambda self, ff: ff.build_headers_frame(
                 self.example_request_headers, flags=['END_STREAM']),
             lambda self, ff: ff.build_headers_frame(
@@ -245,7 +244,6 @@ class TestStreamsClosedByEndStream(object):
     @pytest.mark.parametrize(
         "frame",
         [
-            lambda self, ff: ff.build_data_frame(b'hello'),
             lambda self, ff: ff.build_headers_frame(
                 self.example_response_headers, flags=['END_STREAM']),
             lambda self, ff: ff.build_headers_frame(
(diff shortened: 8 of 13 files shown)


h2/connection.py
```diff
@@ -1646,7 +1646,19 @@ def _receive_data_frame(self, frame):
         self._inbound_flow_control_window_manager.window_consumed(
             flow_controlled_length
         )
-        stream = self._get_stream_by_id(frame.stream_id)
+
+        try:
+            stream = self._get_stream_by_id(frame.stream_id)
```
vmagamedov: There are two ways to know that stream is closed. First one is to handle a `StreamClosedError`. Second one is to check `if stream.closed`. I think that this issue is valid for both cases.

h2/connection.py
```diff
@@ -1646,7 +1646,19 @@ def _receive_data_frame(self, frame):
...
+        except StreamClosedError:
+            # This stream is already closed - and yet we received a DATA frame.
+            # The received DATA frame counts towards the connection flow window.
+            # We need to manually to acknowledge the DATA frame to update the flow window of the connection.
+            # Otherwise the whole connection stalls due to the inbound flow window being 0.
+            if flow_controlled_length > 0:
+                self.increment_flow_control_window(flow_controlled_length)
```
vmagamedov: `increment_flow_control_window()` method has a side-effect within and I don't think that it is appropriate to use it here. `_receive_data_frame` method is intended to **return** frames and events, but now it can also send frames on it's own.

Kriechi: We are already using that - in the existing flow when a DATA frame on a closed stream is received, it will emit a RST_FRAME to inform the other side that it should no longer send frames for this stream: https://github.com/python-hyper/hyper-h2/blob/306609aa72c5ad4860a72fc4caa5db9ef5ab5b26/h2/connection.py#L1492 and notice the `_prepare_for_sending` call in both cases.

h2/connection.py
Kriechi: I'm not sure I understand your comment - do you propose a change?

Kriechi:
> So maybe if stream is closed we have to send WINDOW_UPDATE and RST_STREAM[STREAM_CLOSED]

True, the original implementation already emitted the RST_STREAM, so if I just re-raise the exception after emit the new WINDOW_UPDATE, the existing code paths will take care of the RST_FRAME.

h2/connection.py
vmagamedov: Yes. In a case when:
- stream is not yet removed from the `Connection.streams` dict
- `StreamClosedError` is not raised
- `WINDOW_UPDATE` isn't sent
- stream's state machine raises `StreamClosedError` later in `H2StreamStateMachine.recv_on_closed_stream()`
- so only `RST_STREAM` is sent in this case.

I propose to do one of two things:
- handle `StreamClosedError` also during `stream.receive_data(...)`
- or check that stream is not in "open" or "half-closed (local)" states after first try..except block

... and emit `WINDOW_UPDATE` in any of those cases

h2/connection.py
vmagamedov: All `_receive_*_frame()` methods don't call `_prepare_for_sending()` method to send frames (I hope 😅), they return frames as a list instead. So only `_receive_frame()` method calls `_prepare_for_sending()` method. I think it would be nice to keep this simplicity, it will help to reason about this code's behaviour later.

h2/connection.py
Kriechi: Can `stream.receive_data` even raise a `StreamClosedError`? Can you point me to a possible code path? All state transitions are either valid, or result in `H2StreamStateMachine.reset_stream_on_error` - which we need to handle as discussed.

h2/connection.py
Kriechi: Yes - this makes totally sense - let me see how I can change my PR.

Kriechi:
@vmagamedov I think I now understood most of your comments and made the changes.

A curious problem though: test coverage reports a few missing lines/statements - yet, if I comment-out these lines, some tests will rightfully fail because the expected emitted frame is missing. Is this is bug in the coverage report?

pgjones:
@Kriechi I hadn't released that coverage could (and does in this project) do [this](https://coverage.readthedocs.io/en/coverage-5.0.3/branch.html) i.e. flag lines that haven't been visited via a different branch. Could this be the issue?

Kriechi:
Thanks @pgjones -- close enough! My branch was actually visited all the time! And coverage complained that we do not have a single test that does NOT hit that branch.

I'll push a final fix shortly.

Kriechi:
@vmagamedov @Lukasa @pgjones any chance of getting a review (and potentially a release afterwards)?

h2/connection.py
```diff
@@ -1450,7 +1450,7 @@ def receive_data(self, data):
         :returns: A list of events that the remote peer triggered by sending
             this data.
         """
-        self.config.logger.debug(
+        self.config.logger.trace(
```
pgjones: @Kriechi I don't follow this change, trace doesn't seem to do anything different - so why change it?

setup.py
```diff
@@ -58,18 +58,19 @@
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
```
pgjones: @Kriechi probably should drop 3.4 as well, also can you add to the changelog?

test/test_stream_reset.py
```diff
@@ -89,19 +89,22 @@ def test_reset_stream_keeps_flow_control_correct(self,
         c.clear_outbound_data_buffer()
 
         f = frame_factory.build_data_frame(
-            data=b'some data!',
+            data=b'some data',
```
pgjones: @Kriechi I don't follow this change, did the `!` have some effect on the test?

h2/connection.py
Kriechi: When using a logger that prints "debug" message it will get very very noisy, because it also prints every raw data packing coming in from the socket. 

Here I move this noisy raw data print one level down into a "trace" call - because most people will want to look at the "debug" message from hyper-h2 (what it has parsed and what state transitions, etc.). If anyone is really interested in raw on-the-wire bytestreams, then "trace" is the option to go.

setup.py
Kriechi: 3.3 was already dropped in one of the last releases - but for some reason was still mentioned here. This PR does not change the supported Python versions - it only makes it explicit in setup.py.

Yes, dropping py3.4 would have been my next PR 😄

test/test_stream_reset.py
Kriechi: Nothing important - some tests used `some data`, or `some data!`, or `foobar`.
I simply wanted to unify the test messages.

h2/connection.py
```diff
@@ -1634,6 +1634,33 @@ def _receive_push_promise_frame(self, frame):
...
+    def _handle_data_on_closed_stream(self, events, exc, frame):
+        # This stream is already closed - and yet we received a DATA frame.
+        # The received DATA frame counts towards the connection flow window.
+        # We need to manually to acknowledge the DATA frame to update the flow
+        # window of the connection. Otherwise the whole connection stalls due
+        # the inbound flow window being 0.
+        if self._stream_is_closed_by_reset(exc.stream_id):
+            frames = []
+            if frame.flow_controlled_length > 0:
+                self._inbound_flow_control_window_manager.window_opened(
```
vmagamedov: There is an important distinction between `window_opened()` and `process_bytes()` methods of a `WindowManager` class. I think that the right thing would be to call `process_bytes()` – to acknowledge received data, not to increase/resize the window.

Take a look into `window_opened()` method and it's docstring, there is even a possibility to get `FlowControlError` when window becomes big enough.

h2/connection.py
```diff
@@ -1634,6 +1634,33 @@ def _receive_push_promise_frame(self, frame):
 
         return frames, events + stream_events
 
+    def _handle_data_on_closed_stream(self, events, exc, frame):
+        # This stream is already closed - and yet we received a DATA frame.
+        # The received DATA frame counts towards the connection flow window.
+        # We need to manually to acknowledge the DATA frame to update the flow
+        # window of the connection. Otherwise the whole connection stalls due
+        # the inbound flow window being 0.
+        if self._stream_is_closed_by_reset(exc.stream_id):
```
vmagamedov: I don't think that this condition is useful. When we receive `DATA` frame for a closed stream, this is just a stream error, connection is still valid, and I think that we have to acknowledge received data without any conditions.

There is a possibility that we don't know why stream was closed – `Connection._closed_streams` has a limit, so I propose to simplify the logic and remove this condition.

h2/connection.py
Kriechi: Yes - you are correct!

h2/connection.py
Kriechi: This would mean a changing behaviour that is currently ensured by this test:
https://github.com/python-hyper/hyper-h2/blob/aa66d45c177e585259834eae61c2fd90e0f760a5/test/test_closed_streams.py#L236

If I remove the condition, we will never raise anything else - and therefore not product a ProtocolError...

vmagamedov: > If I remove the condition, we will never raise anything else - and therefore not product a ProtocolError...

There are two types of errors in the spec:
- [5.4.1. Connection Error Handling](https://httpwg.org/specs/rfc7540.html#rfc.section.5.4.1)
- [5.4.2. Stream Error Handling](https://httpwg.org/specs/rfc7540.html#rfc.section.5.4.2)

Connection error means that connection is broken and should be closed. Stream error means that connection still can be used.

I believe that `Connection.receive_data()` should raise `ProtocolError` only for connection errors.

Here are another interesting references to the spec:

- https://httpwg.org/specs/rfc7540.html#rfc.section.6.1
   > DATA frames are subject to flow control and can only be sent when a stream is in the "open" or "half-closed (remote)" state. The entire DATA frame payload is included in flow control, including the Pad Length and Padding fields if present. If a DATA frame is received whose stream is not in "open" or "half-closed (local)" state, the recipient MUST respond with a **stream error (Section 5.4.2) of type STREAM_CLOSED**.

- https://httpwg.org/specs/rfc7540.html#rfc.section.6.9
   > A receiver that receives a flow-controlled frame MUST always account for its contribution against the connection flow-control window, unless the receiver treats this as a connection error (Section 5.4.1). **This is necessary even if the frame is in error.** The sender counts the frame toward the flow-control window, but if the receiver does not, the flow-control window at the sender and receiver can become different.

So maybe those tests don't take this into account.

h2/connection.py
```diff
@@ -1634,6 +1634,33 @@ def _receive_push_promise_frame(self, frame):
...
+            frames = []
+            if frame.flow_controlled_length > 0:
+                self._inbound_flow_control_window_manager.process_bytes(
```
vmagamedov: `process_bytes()` method works in an interesting (optimised) way, it returns window increment – this value should be used to send `WINDOW_UPDATE` frame. You can take a look into `H2Connection.acknowledge_received_data()` code for a reference.

vmagamedov: `process_bytes()` method works in an interesting (optimised) way, it returns window increment – this value should be used to send `WINDOW_UPDATE` frame. You can take a look into `H2Connection.acknowledge_received_data()` code for a reference.

Kriechi:
Thanks @vmagamedov - this all makes sense. Please see my last commits that should address all your comments.

vmagamedov:
@Kriechi Looks great! I don't have any other comments.

PR #1225, opened 2020-05-26
ignore WINDOW_UPDATE frames on removed or non-existent stream

mihau:
This pull request fixes an issue with receiving `WINDOW_UPDATE` frames on streams that have been closed and removed - which could happen while opening a new stream.

Please let me know if anything should be done differently. Here are some doubts I have about this approach:

1. This code will also ignore streams that have never been created. Maybe checking in `_closed_streams` as well would be better.
2. Maybe a separate test would make more sense.

### DIFF
--- a/h2/connection.py
+++ b/h2/connection.py
@@ -1739,7 +1739,7 @@ class H2Connection(object):
                     frame.window_increment
                 )
             except StreamClosedError:
-                return [], []
+                return [], events
         else:
             # Increment our local flow control window.
             self.outbound_flow_control_window = guard_increment_window(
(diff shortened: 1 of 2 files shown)


h2/connection.py
```diff
@@ -1733,10 +1733,13 @@ def _receive_window_update_frame(self, frame):
...
-            frames, stream_events = stream.receive_window_update(
-                frame.window_increment
-            )
+            try:
+                stream = self._get_stream_by_id(frame.stream_id)
+                frames, stream_events = stream.receive_window_update(
+                    frame.window_increment
+                )
+            except StreamClosedError:
+                return [], []
```
Kriechi, line 1742: Just to keep the state machine logic happy, we should `return [], events` here.

mihau: Ok, done.

szerepak:
@mihau @Kriechi Any chance of merging this one here?

PR #1243, opened 2021-01-04
add related projects

akvadrako:
Mention related projects, i.e. those using hyper.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -36,5 +36,6 @@ Contents
    low-level
    api
    testimonials
+   related
    release-process
    release-notes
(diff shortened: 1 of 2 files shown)


sethmlarson:
There's already a section called "Testimonials", can we extend the usefulness of that section by changing it to "testimonials and related projects"? Also the release date, etc isn't relevant to h2 usage.

akvadrako:
> There's already a section called "Testimonials", can we extend the usefulness of that section by changing it to "testimonials and related projects"?

Yes, I thought of that, it just makes it a rather long title.

> Also the release date, etc isn't relevant to h2 usage.

For me this is what I would care about when looking for related projects; they should be active and popular.

PR #1246, opened 2021-02-21
Rename 'hyper-h2' to simply 'h2'

### LINKED ISSUE #1245: Proposal: Rename this repository to "h2"
The name has [caused confusion](https://github.com/encode/httpx/discussions/1477) due to the `hyper` project being deprecated. And to be fair I've definitely forgotten that this URL is special and doesn't match it's PyPI package name. Should we rename the project to simply `h2`?

sethmlarson:
Closes #1245 

I went the simple approach of changing `[hH]yper-h2` -> `h2` for all code occurrences, the only case I found we'd want to keep `hyper-h2` is Read the Docs for now, but maybe those can be changed to use `python-hyper.org`?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,15 +1,15 @@
-===============================
-hyper-h2: HTTP/2 Protocol Stack
-===============================
+=========================
+h2: HTTP/2 Protocol Stack
+=========================
 
-.. image:: https://github.com/python-hyper/hyper-h2/workflows/CI/badge.svg
-    :target: https://github.com/python-hyper/hyper-h2/actions
+.. image:: https://github.com/python-hyper/h2/workflows/CI/badge.svg
+    :target: https://github.com/python-hyper/h2/actions
     :alt: Build Status
-.. image:: https://codecov.io/gh/python-hyper/hyper-h2/branch/master/graph/badge.svg
-    :target: https://codecov.io/gh/python-hyper/hyper-h2
+.. image:: https://codecov.io/gh/python-hyper/h2/branch/master/graph/badge.svg
+    :target: https://codecov.io/gh/python-hyper/h2
     :alt: Code Coverage
-.. image:: https://readthedocs.org/projects/hyper-h2/badge/?version=latest
-    :target: https://hyper-h2.readthedocs.io/en/latest/
+.. image:: https://readthedocs.org/projects/h2/badge/?version=latest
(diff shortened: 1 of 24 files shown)


README.rst
```diff
@@ -67,10 +67,11 @@ please `read the contribution guidelines`_.
...
 
-``hyper-h2`` is made available under the MIT License. For more details, see the
+``h2`` is made available under the MIT License. For more details, see the
 ``LICENSE`` file in the repository.
 
 Authors
 =======
 
-``hyper-h2`` is maintained by Cory Benfield, with contributions from others.
+``h2`` was authored by Cory Benfield and is maintained
```
sethmlarson, line 76: Changed this section to more accurately reflect the maintainership of the package

README.rst
```diff
@@ -45,17 +45,17 @@ To install it, just run:
 
 .. code-block:: console
 
-    $ pip install h2
+    $ python -m pip install h2
```
sethmlarson, line 48: Changed `$ pip` to `$ python -m pip` in line with PyPA best practices

examples/twisted/head_request.py
```diff
@@ -94,7 +94,7 @@ def sendRequest(self):
             (':authority', AUTHORITY),
             (':scheme', 'https'),
             (':path', PATH),
-            ('user-agent', 'hyper-h2/1.0.0'),
+            ('user-agent', 'h2/1.0.0'),
```
Kriechi, line 97: nitpick: this might be a bit confusing now - it's just an example, so not really critical.
Maybe something: `python-hyper/h2/1.0.0`? 
or omit the user-agent entirely?

Kriechi:
Overall 👍 

Please add a changelog entry to mention the github repo rename, and the fact that the python package name is **unchanged**.

One thing we should watch out for, and have to be careful about, is when talking about h2 the project in close proximity to ALPN and the h2 spec itself. This might require some more explicit language, e.g., "the h2 project".

examples/twisted/head_request.py
sethmlarson: Going to omit it entirely

sethmlarson:
@Kriechi Renamed the RTD, added a changelog entry (is this the right place? not a bug fix or API change) and remove `user-agent` in the examples directory.

PR #1248, opened 2021-03-08
Workaround the issues with hypothesis 6.6

hroncok:
Hypothesis 6.6 will produce:

    E  hypothesis.errors.FailedHealthCheck: test/test_flow_control_window.py::TestAutomaticFlowControl::test_mixing_update_forms uses the 'frame_factory' fixture, which is reset between function calls but not between test cases generated by `@given(...)`.  You can change it to a module- or session-scoped fixture if it is safe to reuse; if not we recommend using a context manager inside your test function.  See https://docs.pytest.org/en/latest/fixture.html#sharing-test-data for details on fixture scope.
    E  See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.function_scoped_fixture to the suppress_health_check settings for this test.

Since the tests already workaround the problem,
acknowledging https://github.com/HypothesisWorks/hypothesis-python/issues/377,
we can safely disable the check.

Hypothesis 5.49 introduced the function_scoped_fixture health check value,
hence it is now the lowest required version of hypothesis.


bnavigator:
Thank you! This is what we need in openSUSE Tumbleweed as well.

Kriechi (approved):

Issue #1257, opened 2021-09-02
"IndexError: index out of range" in _reject_surrounding_whitespace


ad-m:
Hello,

I am working on integrating fuzzer to project `h2`. My goal is to include it in the `oss-fuzz` project. I believe that this can be a valuable contribution to the library, the importance of which is becoming more and more higher, among others due to the use by `httpx` - the library which popularity also growing. 

I identified the first problems and I would like to know if such problems are interesting for your project and your are interested to handle them.

```python
import h2.connection
import h2.config
from h2.exceptions import ProtocolError

payload = b'\x00\x00\x08\x01\x05\xef\x00\x00@\x00\x00\x05\x00\x00\x00\x00\x04\x8c\x00'
config = h2.config.H2Configuration()
conn = h2.connection.H2Connection(config=config)
try:
    conn.receive_data(payload)
    conn.data_to_send()
except ProtocolError:
    pass
```

```
Traceback (most recent call last):
  File "/tmp/h2-fuzz/test2.py", line 9, in <module>
    conn.receive_data(payload)
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/connection.py", line 1463, in receive_data
    events.extend(self._receive_frame(frame))
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/connection.py", line 1486, in _receive_frame
    frames, events = self._frame_dispatch_table[frame.__class__](frame)
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/connection.py", line 1560, in _receive_headers_frame
    frames, stream_events = stream.receive_headers(
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/stream.py", line 1057, in receive_headers
    events[0].headers = self._process_received_headers(
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/stream.py", line 1300, in _process_received_headers
    return list(headers)
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/utilities.py", line 492, in inner
    for header in headers:
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/utilities.py", line 437, in _validate_host_authority_header
    for header in headers:
  File "/tmp/tmp.mokMU0wrkK/lib/python3.9/site-packages/h2/utilities.py", line 331, in _reject_pseudo_header_fields
[...]

PR #1258, opened 2021-09-03
Add atheris integration

ad-m:
As mentioned in #1257 I am adding initial fuzzing integration. My goal is to integrate with oss-fuzz and - personally - explore the world of fuzzing.

Documentation has been provided that shows the initial development flow with the help of fuzzing. 

The initial corpus was prepared based on existing unit tests. I am convinced that it is not exhaustive corpus, but I would like to get some comments on the selected fuzz-targets. Using the proposed corpus requires fixing bug # 1257 (the corpus was generated with a local patch for this bug).

The location of the fuzzing corpus storage needs to be considered. Different projects approach it differently. In the future, `os-fuzz` might host itself as well, but it also requires a bootstrap too.

```
$ du -sh fuzz/corpus/
11M	fuzz/corpus/
$ ls -1 fuzz/corpus/ | wc -l
2587
```


tomchristie:
I think we can safely close this off...

Screenshot 2023-05-17 at 14 08 24

Does appear to have caught a valid bug in #1257 which was closed via #1259, but there hasn't been any discussion about adding the tooling directly into the workflow.

PR #1261, opened 2021-09-09
add plain sockets example client

### LINKED ISSUE #1260: [Question] How can I send a POST request?
I'm trying to communicate with the Apple Push Notification service (APN) which uses HTTP/2 ([docs](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html#//apple_ref/doc/uid/TP40008194-CH11-SW1)).

I've (sucessfully) used [hyper](https://pypi.org/project/hyper/) before for that with code that roughly looks like this:

```python
from hyper import HTTPConnection
import json

bearer_token = "foo"
device_token = "bar"
payload = {}

request_headers = {
    "apns-expiration": "0",
    "apns-priority": "10",
    "authorization": f"bearer {bearer_token}",
}

conn = HTTPConnection("api.development.push.apple.com:443")
conn.request(
    "POST", f"/3/device/{device_token}",
    json.dumps(payload).encode(),
    headers=request_headers
)
resp = conn.get_response()
print(resp.status)
print(resp.reason)
print(resp.read())
```

How would an equivalent with h2 look like?

Kriechi:
closes #1260

/cc @MT-Cash

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -12,11 +12,11 @@ Example Servers
    :maxdepth: 2
 
    asyncio-example
-   twisted-example
+   curio-example
    eventlet-example
    gevent-example
-   curio-example
    tornado-example
+   twisted-example
    wsgi-example
 
 Example Clients
@@ -25,5 +25,6 @@ Example Clients
 .. toctree::
    :maxdepth: 2
 
+   plain-sockets-example
    twisted-head-example
    twisted-post-example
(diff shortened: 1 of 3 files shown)


docs/source/examples.rst
```diff
@@ -12,11 +12,11 @@ Example Servers
...
 
    asyncio-example
-   twisted-example
+   curio-example
    eventlet-example
    gevent-example
-   curio-example
    tornado-example
+   twisted-example
    wsgi-example
```
Kriechi, line 20: just sorted alphabetically now.

PR #1263, opened 2021-09-16
Simplifies conditional logic and Refactor Multiple Functions

yezz123:
- Merges together multiply nested `if` conditions into one.
  > Too much nesting can make code difficult to understand, and this is especially true in Python, where there are no brackets to help out with the delineation of different nesting levels.
- Simplifies conditional logic using De Morgan's Identity
  > When reading logical statements it is important for them to be as simple as possible. This proposal applies De Morgan's identity to simplify code by removing double negatives or pushing negations deeper into statements.
- Removes conditional tests where the conditional is always `True` or `False`
  > This refactoring will re-structure conditionals where determines that one of the tests is always True or False. This reveals the actual logic of the conditional to the reader, making the code easier to understand.

### DIFF
--- a/examples/asyncio/wsgi-server.py
+++ b/examples/asyncio/wsgi-server.py
@@ -603,8 +603,7 @@ class Stream:
 
         # We only need the status code
         status = self._response_status.split(" ", 1)[0]
-        headers = [(":status", status)]
-        headers.extend(self._response_headers)
+        headers = [(":status", status), *self._response_headers]
         event = self._protocol.send_response(self.stream_id, headers)
         event.wait()
         return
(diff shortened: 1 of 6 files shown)


Kriechi:
* `_extracted_from_receive_headers_15`
* `_extracted_from_acknowledge_received_data_10`
* etc.

Are these auto-generated function names from your IDE or how did you come up with these names and numerical prefixes? 
Since I don't understand the pattern, I don't like it -- convince me otherwise! 😄 

I'm in favour of simplifying the conditions, but I'm not so sure about the function extractions. While it follows the DRY idea, I think in most cases here, the explicit code is easier to read and understand, given that it always follow the same pattern. With the current function names the readability suffers significantly. I do not see an improvement at the moment.

Would you mind splitting the condition refactorings into a separate PR or commit to review them independently from the function extractions?

PR #1273, opened 2022-09-20
[Docs] Possible typo in tutorial

rmNULL:
This seems like a semantic error, because we are not writing a server that writes another server to handle HTTP GET.


Kriechi:
Thanks! 🎉

PR #1275, opened 2022-12-28
encode/httpx#1214 prefer sending outbound cookies separately to improve headers compression

cdeler:
According to the [rfc7540#section-8.1.2.5](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.5), 

> To allow for better compression efficiency, the Cookie header field
>   MAY be split into separate header fields

The `h2` framework is used by the `httpx` client, and some users require to have an ability to send cookies via separate headers to follow http2 practice. Or at least to make this option configurable (more:  [encode/httpx#1214](https://github.com/encode/httpx/issues/1214)).

As `h2` decides how to properly send data over the transport, `h2` seems to be a right place to add an ability to split the cookie headers. Also the opposite functionality is added to the inbound cookies processing ([ref](https://github.com/python-hyper/h2/blob/63b6b97e691334464a57f933bfd7ed444f364708/src/h2/utilities.py#L621)).

### Notes:
I have concerns about my PR, may be you can help me with suggestings:
* I used the "server" connection in the client connections tests. I don't know if it's a right way to do
* may be it makes sense to hide this feature under a config? If so, do you have suggestions where to add it?


src/h2/utilities.py
```diff
@@ -603,6 +603,30 @@ def _combine_cookie_fields(headers, hdr_validation_flags):
         yield NeverIndexedHeaderTuple(b'cookie', cookie_val)
 
 
+def _split_outbound_cookie_fields(headers, hdr_validation_flags):
+    """
+    RFC 7540 § 8.1.2.5 allows for better compression efficiency,
+    to split the Cookie header field into separate header fields
+
+    We want to do it for outbound requests, as we are doing for
+    inbound
```
Kriechi, line 612:
```suggestion
    inbound.
```

src/h2/utilities.py
```diff
@@ -603,6 +603,30 @@ def _combine_cookie_fields(headers, hdr_validation_flags):
...
+    for header in headers:
+        if header[0] in (b'cookie', 'cookie'):
+            needle = b'; ' if isinstance(header[0], bytes) else '; '
+
+            if needle in header[1]:
+                for cookie_val in header[1].split(needle):
+                    if isinstance(header, HeaderTuple):
+                        yield header.__class__(header[0], cookie_val)
+                    else:
+                        yield header[0], cookie_val
```
Kriechi, line 623: At this point, what are the trade-offs to simply always create `HeaderTuple`'s?

Kriechi: Since we use the same approach in https://github.com/python-hyper/h2/blob/c7f967d77a6f32be374e33adca195e65a00afb63/src/h2/utilities.py#L549-L552, I don't mind either way.

sethmlarson:
Question for some context: if a client sends Cookie headers that are already split does h2 combine them somewhere or will the split header fields be encoded as expected?

Kriechi:
@sethmlarson I asked myself the same thing, and found this:
https://github.com/python-hyper/h2/blob/c7f967d77a6f32be374e33adca195e65a00afb63/src/h2/utilities.py#L583-L603

src/h2/utilities.py
cdeler: I also was thinking about the `NeverIndexedHeaderTuple` like
https://github.com/python-hyper/h2/blob/63b6b97e691334464a57f933bfd7ed444f364708/src/h2/utilities.py#L601-L603

as we know for sure that it's a `cookie` header

@Kriechi , do you have any suggestions about that?

cdeler: I'm not sure about that, but I see that all outbound cookies are handled using this pattern: `isinstance(header, HeaderTuple): ...` 

https://github.com/python-hyper/h2/blob/63b6b97e691334464a57f933bfd7ed444f364708/src/h2/utilities.py#L534-L538
https://github.com/python-hyper/h2/blob/63b6b97e691334464a57f933bfd7ed444f364708/src/h2/utilities.py#L548-L552

Also I'm not sure about using `NeverIndexedHeaderTuple` (e.g. how it can negatively affect compression)

Kriechi:
as a first step to a review, I fixed our project CI (which was a challenge by itself).
https://github.com/python-hyper/h2/pull/1276

I don't see any obvious issues with the code itself - though I still need time to think and read through the spec and implications. Happy to hear other thoughts in the mean time!

@cdeler you said:
> some users require to have an ability 
and
> to add an ability to split

I don't see an "option" in your PR - its "always on"?
Or do you think this should be the new one-and-only way to handle cookies?

cdeler:
Hello @Kriechi 

> Or do you think this should be the new one-and-only way to handle cookies?

I don't think so (because RFC says "may") :-) , so that I updated the PR adding `split_outbound_cookies` flag to the `h2.config.H2Configuration` (`False` by default)

Kriechi:
LGTM!

Kriechi (approved):

kuntaoKZ:
Hello,
seems this change was not covered in 4.1.0 version?

PR #1276, opened 2023-01-02
fix CI

Kriechi:
* Python 3.6 doesn't install any more on `ubuntu-latest` which currently is 22.04
* API rate limiting hit us without the correct token
* update test dependencies and fix their now incompatible config


sethmlarson (approved):
LGTM, thanks for this!

PR #1277, opened 2023-01-30
Run CI for Python 3.11

mhils:


mhils:
Absolutely - thanks for catching this :)

PR #1281, opened 2024-03-28
Fix bug when v is None

puke3615:
When we use a library called 'wechaty', the variable `v` may be None. And at this time, it will throw an exception: `'NoneType' object has no attribute 'encode'`.


Kriechi:
thanks! This should already be addressed by recent code refactoring.

PR #1283, opened 2024-08-14
Update docs

bmwant:
### Changes

* Add `curl` usage examples with http2
* Add `openssl` certificate generation command
* Fix `readlines` methods
* Fix `async`/`await` syntax
* Minor typos

### DIFF
--- a/docs/source/basic-usage.rst
+++ b/docs/source/basic-usage.rst
@@ -156,7 +156,7 @@ This is not a socket tutorial, so we're not going to dive too deeply into how
 this works. If you want more detail about sockets, there are lots of good
 tutorials on the web that you should investigate.
 
-When you want to listen for incoming connections, the you need to *bind* an
+When you want to listen for incoming connections, you need to *bind* an
 address first. So let's do that. Try setting up your file to look like this:
 
 .. code-block:: python
@@ -213,11 +213,12 @@ connection object and start handing it data. For now, let's just see what
 happens as we feed it data.
 
 To make HTTP/2 connections, we need a tool that knows how to speak HTTP/2.
-Most versions of curl in the wild don't, so let's install a Python tool. In
-your Python environment, run ``pip install hyper``. This will install a Python
-command-line HTTP/2 tool called ``hyper``. To confirm that it works, try
-running this command and verifying that the output looks similar to the one
-shown below:
+You can simply use `curl`_ or install a Python tool used throughout this
+tutorial. In your Python environment, run ``pip install hyper`` (Make sure to
+use ``Python < 3.10`` or install it in separate environment). This will
+install a Python command-line HTTP/2 tool called ``hyper``. To confirm that
+it works, try running this command and verifying that the output looks similar
+to the one shown below:
 
 .. code-block:: console
 
@@ -226,6 +227,16 @@ shown below:
      'headers': {'Host': 'nghttp2.org'},
      'origin': '10.0.0.2',
      'url': 'https://nghttp2.org/httpbin/get'}
+
+Equivalent code with curl would look like this:
+
+.. code-block:: console
+
+    $ curl --http2 https://nghttp2.org/httpbin/get
+
+To use it with our server though, you will need to invoke it with a different
+``--http2-prior-knowledge`` flag as we are going to serve over the insecure
+connection.
 
 Assuming it works, you're now ready to start sending HTTP/2 data.
 
@@ -290,7 +301,9 @@ function. Your ``h2server.py`` should end up looking a like this:
         handle(sock.accept()[0])
 
 Running that in one shell, in your other shell you can run
-``hyper --h2 GET http://localhost:8080/``. That shell should hang, and you
+``hyper --h2 GET http://localhost:8080/``. For the ``curl`` use
+``curl -v --http2-prior-knowledge http://localhost:8080/`` command.
+That shell should hang, and you
 should then see the following output from your ``h2server.py`` shell:
 
 .. code-block:: console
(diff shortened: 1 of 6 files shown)


docs/source/basic-usage.rst
```diff
@@ -213,11 +213,12 @@ connection object and start handing it data. For now, let's just see what
...
-your Python environment, run ``pip install hyper``. This will install a Python
-command-line HTTP/2 tool called ``hyper``. To confirm that it works, try
-running this command and verifying that the output looks similar to the one
-shown below:
+You can simply use `curl`_ or install a Python tool used throughout this
+tutorial. In your Python environment, run ``pip install hyper`` (Make sure to
+use ``Python < 3.10`` or install it in separate environment). This will
+install a Python command-line HTTP/2 tool called ``hyper``. To confirm that
+it works, try running this command and verifying that the output looks similar
+to the one shown below:
```
Kriechi, line 221: Since the `hyper` project itself is archived / no longer supported / not maintained, I would suggest to remove the mention of it in this paragraph entirely. `curl` is sufficiently modern nowadays and can be the only mention here.

docs/source/basic-usage.rst
```diff
@@ -227,6 +228,16 @@ shown below:
      'origin': '10.0.0.2',
      'url': 'https://nghttp2.org/httpbin/get'}
 
+Equivalent code with curl would look like this:
+
+.. code-block:: console
+
+    $ curl --http2 https://nghttp2.org/httpbin/get
+
+To use it with our server though, you will need to invoke it with a different
```
Kriechi, line 237:
```suggestion
To use it with a `python-hyper/h2`-based server, you can invoke it with a different
```

examples/asyncio/wsgi-server.py
```diff
@@ -688,41 +673,41 @@ def _build_environ_dict(headers, stream):
...
+        'wsgi.run_once': False,
     }
-    if u'content-type' in header_dict:
-        environ[u'CONTENT_TYPE'] = header_dict[u'content-type']
-    if u'content-length' in header_dict:
-        environ[u'CONTENT_LENGTH'] = header_dict[u'content-length']
+    if 'content-type' in header_dict:
+        environ['CONTENT_TYPE'] = header_dict.pop('content-type')
+    if 'content-length' in header_dict:
+        environ['CONTENT_LENGTH'] = header_dict.pop('content-length')
```
Kriechi, line 708: Any specific reason for using `pop` instead of `get` here?
I feel like using `get` would be more intuitive, unless you intended the side effect?

examples/asyncio/wsgi-server.py
```diff
@@ -534,23 +533,9 @@ def readline(self, hint=None):
...
-        lines = data.split(b'\n')
-
-        # Split removes the newline character, but we want it, so put it back.
-        lines = [line + b'\n' for line in lines]
-
-        # Except if the last character was a newline character we now have an
-        # extra line that is just a newline: pull that out.
-        if lines[-1] == b'\n':
-            lines = lines[:-1]
+        lines = data.splitlines(True)
```
Kriechi, line 538:
```suggestion
        lines = data.splitlines(keepends=True)
```

Kriechi: Not sure what the implications are regarding Unix vs. Windows line ending when using the library splitlines here.

examples/asyncio/wsgi-server.py
bmwant: https://wsgi.readthedocs.io/en/latest/definitions.html#envvar-CONTENT_TYPE
yes, both `CONTENT_TYPE` and `CONTENT_LENGTH` headers should stay "as-is" without prefixing, otherwise we will pass them to the environ with `HTTP_`, that's why we remove them (see same for the request method above)

examples/asyncio/wsgi-server.py
bmwant: Don't think there is any, I've also compared behaviour with Werkzeug's implementation and it does the same for both types on line endings. https://github.com/pallets/werkzeug/blob/main/src/werkzeug/wsgi.py#L425.

Also doesn't violate any requirements from the PEP itself https://peps.python.org/pep-3333/#input-and-error-streams

bmwant:
@Kriechi addressed your concerns, still have one open regarding phrasing, but otherwise should be ready for another check

docs/source/basic-usage.rst
Kriechi: Do you agree with removing `hyper` from this paragraph?

docs/source/basic-usage.rst
```diff
@@ -290,7 +301,9 @@ function. Your ``h2server.py`` should end up looking a like this:
         handle(sock.accept()[0])
 
 Running that in one shell, in your other shell you can run
-``hyper --h2 GET http://localhost:8080/``. That shell should hang, and you
+``hyper --h2 GET http://localhost:8080/``. For the ``curl`` use
+``curl -v --http2-prior-knowledge http://localhost:8080/`` command.
```
Kriechi, line 305: Same as above, please remove / replace the usage of `hyper` with `curl` here.

examples/asyncio/wsgi-server.py
Kriechi: Thanks - now it makes sense!

docs/source/basic-usage.rst
bmwant: Did update it, for some reason forgot to push changes, should be good now

Kriechi:
Thanks for making the changes! Looks good to me overall.

Seems the docs building with sphinx broke - likely unrelated to this PR though... I'll look into it in the next days, unless you beat me too it 😄

bmwant:
@Kriechi fixed `tox -e docs` by bumping Sphinx version used 
Screenshot 2024-08-21 at 09 47 07

Kriechi (approved):

Kriechi:
Thanks! 🎉

PR #1286, opened 2024-11-13
fix: No more BytesWarnings

### LINKED ISSUE #1236: Comparison between bytes and string in defining a frozenset throws exception
https://github.com/python-hyper/hyper-h2/blob/3b0b241d79f5a9ff9382bbc038f84862e0d80abf/src/h2/utilities.py#L20-L26.

Hi, when a python process runs with a flag `-bb`, the above part of code will throw exception and make `h2` not work. May I ask why we define both bytes and string in the frozenset? Is it possible to use only bytes or string? Frozenset will compare keys for deduplication.

BYK:
Fixes #1236.

This patch makes all header operations operate on `bytes` and converts all headers and values to bytes before operation. With a follow up patch to `hpack` it should also increase efficiency as currently, `hpack` casts everything to a `str` first before converting back to bytes: https://github.com/python-hyper/hpack/blob/02afcab28ca56eb5259904fd414baa89e9f50266/src/hpack/hpack.py#L150-L151


BYK:
Note to reviewer: I'm aware there are formatting changes (they came from "format on save") and although I think things are more readable now, I'm happy to find and revert all unrelated formatting changes to make reviews easier.

Unfortunately, I realized the auto format changes way too late and wanted to get the patch up as promised first.

Kriechi:
Please keep the changes in each PR to a minimum - in this case please revert the style-related changes.
Discussing coding style is outside of the scope here.

BYK:
@Kriechi we should be good now, sorry for the trouble.

src/h2/utilities.py
```diff
@@ -13,51 +13,56 @@
 
 from .exceptions import ProtocolError, FlowControlError
 
+
 UPPER_RE = re.compile(b"[A-Z]")
+SIGIL = ord(b':')
+STATUS_HEADER = b':status'
```
Kriechi, line 19: Why did you introduce the status header as constant here, but none of the other `:`-prefixed headers that get used frequently?

src/h2/events.py
```diff
@@ -589,11 +589,9 @@ def __init__(self):
...
 
     def __repr__(self):
-        return (
-            "<AlternativeServiceAvailable origin:%s, field_value:%s>" % (
-                self.origin.decode('utf-8', 'ignore'),
-                self.field_value.decode('utf-8', 'ignore'),
-            )
+        return "<AlternativeServiceAvailable origin:%s, field_value:%s>" % (
+            self.origin.decode("utf-8", "replace"),
+            self.field_value.decode("utf-8", "replace"),
```
Kriechi, line 594: why the change in decoding from ignore to replace? If this is an unrelated change, please revert it and let's discuss in a new issue or PR.

test/test_utility_functions.py
```diff
@@ -152,24 +152,17 @@ def test_does_not_increment_without_stream_send(self):
...
 
-    @pytest.mark.parametrize(
-        'headers', [example_request_headers, example_headers_with_bytes]
-    )
-    def test_extract_header_method(self, headers):
-        assert extract_method_header(headers) == b'GET'
+    def test_extract_header_method(self):
+        assert extract_method_header(
+            self.example_headers_with_bytes
+        ) == b'GET'
```
Kriechi, line 165: applies to all changed tests: 

I think we are missing tests now for the case where headers get changed using the newly introduced `utf8_encode_headers` function. Please add relevant tests to ensure string headers always get correctly encoded into bytes.

test/test_invalid_headers.py
```diff
@@ -296,7 +296,9 @@ def test_headers_event_skipping_validation(self, frame_factory, headers):
         c.send_headers(1, headers)
 
         # Ensure headers are still normalized.
-        norm_headers = h2.utilities.normalize_outbound_headers(headers, None, False)
+        norm_headers = h2.utilities.normalize_outbound_headers(
+            h2.utilities.utf8_encode_headers(headers), None, False
+        )
```
Kriechi, line 301: I'm contemplating making the encoding of headers part of the normalization pipeline, instead of being the responsibility of the outside caller.

What do you think about integrating it here?

src/h2/utilities.py
BYK: Carry over from the function where this was used. Moved back inline for consistency. Pulling all these common headers to globals _may_ help with performance but thoughts for future.

src/h2/events.py
BYK: Unrelated, reverting.

src/h2/stream.py
```diff
@@ -1319,7 +1322,7 @@ def _initialize_content_length(self, headers):
                     self._expected_content_length = int(v, 10)
                 except ValueError:
                     raise ProtocolError(
-                        "Invalid content-length header: %s" % v
+                        f"Invalid content-length header: {repr(v)}"
```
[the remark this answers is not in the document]
BYK: This (and the family of these) are not unrelated. Since all header name and values are now bytes, they need explicit casting to `str`. Using `repr()` gives us a better representation so I opted for that.

The only thing we can argue is the usage of f-strings. For that, I'd say since we are touching this line, it is fair game to switch the old `%` syntax to `f-strings` for readability. Would revert that part if you insist tho.

test/test_invalid_headers.py
BYK: That's exactly what I wanted but unfortunately we have an option to bypass normalization, which breaks that neat abstraction :(

test/test_utility_functions.py
BYK: Since I made all utility functions operate on bytes, I don't think we are missing any coverage here: they don't work with unicode and we don't have unicode tests. If we had typing, we could have enforced the "bytes-only" headers thing but we don't have them (yet?). Coverage report also says 100%:

```
---------- coverage: platform linux, python 3.10.12-final-0 ----------
Name                                                         Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------------------------------------
.tox/py310/lib/python3.10/site-packages/h2/__init__.py           1      0      0      0   100%
.tox/py310/lib/python3.10/site-packages/h2/config.py            54      0      8      0   100%
.tox/py310/lib/python3.10/site-packages/h2/connection.py       622      0    152      0   100%
.tox/py310/lib/python3.10/site-packages/h2/errors.py            22      0      0      0   100%
.tox/py310/lib/python3.10/site-packages/h2/events.py           137      0      4      0   100%
.tox/py310/lib/python3.10/site-packages/h2/exceptions.py        44      0      0      0   100%
.tox/py310/lib/python3.10/site-packages/h2/frame_buffer.py      62      0     20      0   100%
.tox/py310/lib/python3.10/site-packages/h2/settings.py         131      0     38      0   100%
.tox/py310/lib/python3.10/site-packages/h2/stream.py           447      0     88      0   100%
.tox/py310/lib/python3.10/site-packages/h2/utilities.py        225      0    130      0   100%
.tox/py310/lib/python3.10/site-packages/h2/windows.py           35      0     12      0   100%
--------------------------------------------------------------------------------------------------------
TOTAL                                                         1780      0    452      0   100%
Coverage XML written to file coverage.xml

Required test coverage of 100.0% reached. Total coverage: 100.00%
```

That said I'll do another pass to make sure we have normal `str` inputs to headers from upstream functions. I definitely saw a few examples but I'll make sure we cover both cases in tests for higher up functions.

src/h2/stream.py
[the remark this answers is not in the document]
Kriechi: Thanks for the explanation - makes sense now to have this change!

test/test_utility_functions.py
Kriechi: I originally typed "coverage" into my comment but then thought about it some more - and I agree it is likely we have 100% coverage - but since library users might pass in `str` headers, do we miss some cases somehow? While the encoding code path seems to be taken at least once - I'm not sure if we check, assert, and verify on all important code paths.

I need to go back through everything and check our public API definition and if everything is still as expected in terms of accepted & expected types.

BYK: Okay, went through the entire test suite and added bytes and str headers to all tests utilizing any request headers.

BYK: >I need to go back through everything and check our public API definition and if everything is still as expected in terms of accepted & expected types.

I think I've saved you the hassle as I went through all tests and made sure we used both `str` and `bytes` inputs from all high-level usage 😄

src/h2/events.py
Kriechi: Please revert.

BYK: Whoops, sorry. Should be reverted now

test/test_invalid_headers.py
Kriechi: `utf8_encode_headers` is currently called in two places: `connections.py - push_stream` and `stream.py - send_headers`. Both share a code path with `H2Stream._build_headers_frames`.

I'll try to refactor it and share the update to discuss.

BYK: Sounds great!

PR #1289, opened 2024-11-17
add typing information

### LINKED ISSUE #368: Add type information to API
Due to small type inconsistencies one might fall into a trap where incompatible objects might still be auto-mangled into something that fits, but isn't right. One example of such a case just happened: ref #367. The docs clearly stated one thing, but passing in something completely different still worked, until the the last release. The public API was still the same, but the underlying implementation changed.

How do you feel about adding type checks, type hints, or similar annotations to help the user avoid such mistakes?

E.g.:
http://mypy-lang.org/
https://www.python.org/dev/peps/pep-0484/
https://docs.python.org/3/library/typing.html
and similar

I think, since python-hyper aims for Python 2 support as well, this might be a bit more difficult.

Kriechi:
closes #368 

Based on the recently released hyperframe and hpack library updates to support typing now in h2 as well.


Kriechi:
@BYK happy to hear your thoughts on the header typing dilemma!

Kriechi:
@BYK I think I solved the headers type confusion and now all tests pass again. mypy is almost happy, but I think I would like to hear your input on the few remaining findings as they mostly relate to #1286 :

```
src/h2/utilities.py:130: error: Argument 1 to "startswith" of "str" has incompatible type "bytes"; expected "str | tuple[str, ...]"  [arg-type]
src/h2/utilities.py:138: error: Argument 1 to "startswith" of "str" has incompatible type "bytes"; expected "str | tuple[str, ...]"  [arg-type]
src/h2/utilities.py:259: error: Argument 1 to "search" of "Pattern" has incompatible type "bytes | str"; expected "Buffer"  [arg-type]
src/h2/utilities.py:604: error: Argument 1 to "append" of "list" has incompatible type "bytes | str"; expected "bytes"  [arg-type]
src/h2/utilities.py:624: error: Unsupported operand types for in ("bytes | str" and "bytes | str")  [operator]
src/h2/utilities.py:625: error: Argument 1 to "split" of "bytes" has incompatible type "bytes | str"; expected "Buffer | None"  [arg-type]
src/h2/utilities.py:625: error: Argument 1 to "split" of "str" has incompatible type "bytes | str"; expected "str | None"  [arg-type]
```

Kriechi:
@BYK the new `Headers` is defined here https://github.com/python-hyper/hpack/pull/277/files#diff-e10aaf7dbceff5ef795c06588c5b846012815fd4c2fec5b1008255699e9c7b0bR48

Though, I just noticed this syntax is not yet supported on Python 3.9... so I need to find a different solution.

BYK:
@Kriechi made a suggestion over there. I think that also takes care of the py3.9 issue as it doesn't have nested `Union`s?

PR #1290, opened 2024-11-23
No byteswarning

Kriechi:
super-seeds and closes #1286, #1236

@BYK please review. I combined most of your previous commits from #1286, so the only code change should be in my most recent commit on this PR.

### DIFF
--- a/src/h2/utilities.py
+++ b/src/h2/utilities.py
@@ -13,51 +13,55 @@ from hpack import HeaderTuple, NeverIndexedHeaderTuple
 
 from .exceptions import ProtocolError, FlowControlError
 
+
 UPPER_RE = re.compile(b"[A-Z]")
+SIGIL = ord(b':')
+INFORMATIONAL_START = ord(b'1')
+
 
 # A set of headers that are hop-by-hop or connection-specific and thus
 # forbidden in HTTP/2. This list comes from RFC 7540 § 8.1.2.2.
 CONNECTION_HEADERS = frozenset([
-    b'connection', u'connection',
-    b'proxy-connection', u'proxy-connection',
-    b'keep-alive', u'keep-alive',
-    b'transfer-encoding', u'transfer-encoding',
-    b'upgrade', u'upgrade',
+    b'connection',
+    b'proxy-connection',
+    b'keep-alive',
+    b'transfer-encoding',
+    b'upgrade',
 ])
 
 
 _ALLOWED_PSEUDO_HEADER_FIELDS = frozenset([
-    b':method', u':method',
-    b':scheme', u':scheme',
-    b':authority', u':authority',
-    b':path', u':path',
-    b':status', u':status',
-    b':protocol', u':protocol',
+    b':method',
+    b':scheme',
+    b':authority',
+    b':path',
+    b':status',
+    b':protocol',
 ])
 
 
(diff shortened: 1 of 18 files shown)


src/h2/utilities.py
```diff
@@ -105,27 +105,26 @@ def extract_method_header(headers):
...
+    :param headers: The HTTP/2 headers.
     :returns: A boolean indicating if this is an informational response.
     """
     for n, v in headers:
-        # If we find a non-special header, we're done here: stop looping.
+        if not isinstance(n, bytes) or not isinstance(v, bytes):
+            raise ProtocolError(f"header not bytes: {n=:r}, {v=:r}")  # pragma: no cover
 
-        if n and n[0] != SIGIL:
+        if not n.startswith(b':'):
```
BYK, line 121: Why this change? (And the other `.startswith()` below from my `n and n[0] == SIGIL` version?) I'm quite sure they are significantly slower as they involve a function call and are not optimized for 1-byte look up.

src/h2/utilities.py
```diff
@@ -515,14 +514,14 @@ def utf8_encode_headers(headers):
...
-        for header in headers
-    ]
+    encoded_headers = []
+    for header in headers:
+        h = (_to_bytes(header[0]), _to_bytes(header[1]))
+        if isinstance(header, HeaderTuple):
+            encoded_headers.append(header.__class__(h[0], h[1]))
+        else:
+            encoded_headers.append(h)
+    return encoded_headers
```
BYK, line 524: This is again, quite less efficient compared to using a list-comprehension directly.

src/h2/utilities.py
```diff
@@ -95,44 +99,32 @@ def extract_method_header(headers):
...
 def is_informational_response(headers):
     """
-    Searches a header block for a :status header to confirm that a given
+    Searches headers list for a :status header to confirm that a given
     collection of headers are an informational response. Assumes the header
-    block is well formed: that is, that the HTTP/2 special headers are first
-    in the block, and so that it can stop looking when it finds the first
-    header field whose name does not begin with a colon.
+    are well formed and encoded as bytes: that is, that the HTTP/2 special
+    headers are first in the block, and so that it can stop looking when it
```
BYK, line 111:
```suggestion
    headers are first in the list, and so that it can stop looking when it
```

src/h2/utilities.py
```diff
@@ -345,30 +323,26 @@ def _reject_pseudo_header_fields(headers, hdr_validation_flags):
     method = None
 
     for header in headers:
-        if _custom_startswith(header[0], b':', u':'):
+        if header[0][0] == SIGIL:
```
BYK, line 326: If we want to go down the `.startswith()` path, we should align this line with that too. This also probably means the top-level `SIGIL` and `INFORMATIONAL_START` become obsolete.

BYK (approved):
Don't fully agree with the `.startswith()` changes and removal of list comprehension in favor of incremental list building. That said I'm not gonna block them as fixing the bytes warning is more important.

If you agree with my performance concerns, you can revert those changes back to my version before merging.

Finally, thank you so much for pushing this through!

Kriechi:
Yes, let's merge this so we can iterate an smaller changes going forward.
Thanks for your contribution!

PR #1296, opened 2025-02-04
Make h2.events more typing-friendly

mhils:
This PR adds proper constructors for most of the h2 events. This allows us to tighten the type annotations, simplifying type checks in downstream implementations. These events should not be instantiated by anyone except h2 itself, so it shouldn't be a breaking change.

Based on #1297.

### DIFF
--- a/src/h2/events.py
+++ b/src/h2/events.py
@@ -433,6 +433,7 @@ class PingAckReceived(Event):
         return f"<PingAckReceived ping_data:{_bytes_representation(self.ping_data)}>"
 
 
+@dataclass(**kw_only)
 class StreamEnded(Event):
     """
     The StreamEnded event is fired whenever a stream is ended by a remote
@@ -440,9 +441,8 @@ class StreamEnded(Event):
     locally, but no further data or headers should be expected on that stream.
     """
 
-    def __init__(self) -> None:
-        #: The Stream ID of the stream that was closed.
-        self.stream_id: int | None = None
+    stream_id: int
+    """The Stream ID of the stream that was closed."""
 
     def __repr__(self) -> str:
         return f"<StreamEnded stream_id:{self.stream_id}>"
--- a/src/h2/events.py
+++ b/src/h2/events.py
@@ -431,6 +431,7 @@ class StreamEnded(Event):
         return f"<StreamEnded stream_id:{self.stream_id}>"
 
 
+@dataclass(**kw_only)
 class StreamReset(Event):
     """
     The StreamReset event is fired in two situations. The first is when the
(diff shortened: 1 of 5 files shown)


Kriechi:
LGTM! Thanks!

> These events should not be instantiated by anyone except h2 itself, so it shouldn't be a breaking change.

I assume you intended this as "most users will not use this API and won't be affected by this change"?
And not in the way of "this API should be made private and documented as do-not-use"?

Please add a quick note into the changelog. Thanks!

mhils:
> I assume you intended this as "most users will not use this API and won't be affected by this change"?

If by "this API" you mean the `__init__` method, then yes.

I will add constructors for the other events and then also a changelog entry.

mhils:
I've switched to `@dataclass` to make things simpler and less repetitive. What do you think?

Python 3.9 requires some extra logic because it doesn't have `kw_only` yet. We can drop `kw_only` entirely if you'd like, even though I think it's generally a good idea.

Kriechi:
I do like dataclasses!

I do like the kw_only pattern!

I don't like the extra fuzz about py3.9. But since py3.9 becomes EOL later this year, I think its acceptable. Happy to hear other thoughts on it!

mhils:
This is ready for review now. With these changes we get from 31 mypy complaints down to 3 in mitmproxy (see linked PR for the fixes). All other downstream consumers should profit similarly.

src/h2/stream.py
```diff
@@ -315,21 +307,23 @@ def recv_push_promise(self, previous_state: StreamState) -> list[Event]:
...
         return [event]
 
-    def send_end_stream(self, previous_state: StreamState) -> None:
+    def send_end_stream(self, previous_state: StreamState) -> list[Event]:
         """
         Called when an attempt is made to send END_STREAM in the
         HALF_CLOSED_REMOTE state.
         """
         self.stream_closed_by = StreamClosedBy.SEND_END_STREAM
+        return []
```
Kriechi, line 316: What's the idea with this API change? To make it look uniform with other similar methods?
I'm not sure if this will actually make it easier to use, if we keep sprinkling this pattern around and downstream users have to keep checking the returned list for potential events even though it should be explicitly clear that no events can happen.

Please explicitly document in changelog & `versionadded` comment.

src/h2/events.py
```diff
@@ -301,13 +334,16 @@ class WindowUpdated(Event):
...
-        self.stream_id: int | None = None
+    stream_id: int
+    """
+    The Stream ID of the stream whose flow control window was changed.
+    May be ``0`` if the connection window was changed.
+    """
 
-        #: The window delta.
-        self.delta: int | None = None
+    delta: int | None = None
```
Kriechi, line 343: can `delta` still be None here?

src/h2/stream.py
Kriechi: Ah - this is probably PR-bleed-over from https://github.com/python-hyper/h2/pull/1297.
Generally it now makes more sense - thanks!

If this only affects internal methods from the state machine, I think we can skip the `versionadded` comments. WDYT?

mhils: > If this only affects internal methods from the state machine, I think we can skip the versionadded comments. WDYT?

Agreed.

Kriechi (approved):
Please add a changelog entry.

The docs should be auto-generated, or do you think we need to update something?
https://github.com/python-hyper/h2/blob/master/docs/source/api.rst#events

LGTM! Thanks!

PR #1300, opened 2025-02-26
Use bytearray() in FrameBuffer

deedy5:
Changed self.data to use bytearray instead of bytes to improve performance. 
```python
import timeit

data = b"x" * 1024

def test_bytes():
    x = b""
    for _ in range(16):
        x += data

def test_bytearray():
    x = bytearray()
    for _ in range(16):
        x += data

bytes_time = timeit.timeit(test_bytes)
bytearray_time = timeit.timeit(test_bytearray)

print(f"{bytes_time=:.3} seconds")
print(f"{bytearray_time=:.3} seconds")
```
bytes_time=3.0 seconds
bytearray_time=1.09 seconds

### DIFF
--- a/src/h2/frame_buffer.py
+++ b/src/h2/frame_buffer.py
@@ -30,7 +30,7 @@ class FrameBuffer:
     """
 
     def __init__(self, server: bool = False) -> None:
-        self.data = b""
+        self.data = bytearray()
         self.max_frame_size = 0
         self._preamble = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n" if server else b""
         self._preamble_len = len(self._preamble)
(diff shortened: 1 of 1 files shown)


Kriechi:
Thanks for looking into this!

I'm wondering if a `memoryview` would be even more performant in this situation?
Looking at the code, the `data` attribute should probably be `_data` and considered private.

deedy5:
https://github.com/python/cpython/blob/45bc120d4504bff215938bca3f1d08cee2ed7a91/Doc/faq/programming.rst?plain=1#L1148-L1153

deedy5:
> I'm wondering if a memoryview would be even more performant in this situation?

**memoryview** don't support `extend`: https://github.com/python/cpython/blob/main/Objects/memoryobject.c

**bytearray** supports: https://github.com/python/cpython/blob/3f3e1c4095ae229b0c6d0364f289a20732281a96/Objects/bytearrayobject.c#L2082-L2191

Kriechi (approved):

Kriechi:
Thanks! 🎉

PR #1302, opened 2025-04-25
Fix packaging checks and simplify CI

janbrasna:
Due to a [bug](https://redirect.github.com/pypa/twine/issues/1146) in `twine<6.0.1` not reading the correct metadata using wildcards thus failing the `tox -e packaging` steps, I've moved the pin from `5.1.x` to `6.1.x` to include some additional compatibility with packaging dependencies.

This also bumps outdated `actions/*` and generally simplifies and speeds up the workflow.

Inlining the `h2spec` step also allows for getting rid of its install script and exposing the access token.

NB: To currently pass `ruff`/`mypy` checks, I had to fix [`runtime-cast-value` (TC006)](https://docs.astral.sh/ruff/rules/runtime-cast-value/#runtime-cast-value-tc006) by quoting the type expressions in `h2/stream` occurrences of `cast()` …

_Logs: [`actions/runs/14669502174`](https://github.com/janbrasna/h2/actions/runs/14669502174/job/41172508940)_


.github/workflows/ci.yml
```diff
@@ -7,10 +7,9 @@ on:
     branches: ["master"]
 
 jobs:
-  tox:
+  test:
```
janbrasna, line 10: Hmmh this was perhaps not exactly fortunate, as the `id` seems to be defined in repository branch protection settings as required checks… in that case feel free to revert this back:

```suggestion
  tox:
```

.github/workflows/ci.yml
```diff
@@ -21,25 +20,16 @@ jobs:
...
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      run: |
-        .github/workflows/install_h2spec.sh
-    - name: Initialize tox envs
-      run: |
-        tox --parallel auto --notest
-    - name: Test with tox
-      run: |
-        tox --parallel 0
+      run: mkdir bin && curl -sSL "https://github.com/summerwind/h2spec/releases/latest/download/h2spec_linux_amd64.tar.gz" | tar -xz --directory ./bin && ./bin/h2spec --version
```
Kriechi, line 28: Does this always work? The original intention of calling the GitHub API using a token was to get around rate-limiting.

Kriechi:
Not sure about the ruff/mypy typing related changes in this PR. The last commit to the master branch tested cleanly. Could you please shed some light on why you proposed this part? If relevant, please split it out into a dedicated PR for discussion.

.github/workflows/ci.yml
janbrasna: Basically this is not using any API endpoint or raw file access, but the convenience public link for releases — that unless flagged by their security is generally unlimited. (They have a ~5k hits/hr limit however I think they don't limit access from their own infra to stuff like packages, containers or releases?) — I'll check the frequency but we've been using this in rust workflows (instead of waiting for cargo builds) for many years, can't remember it ever failing over a rate limit in GHA.

janbrasna:
@Kriechi Same here, but also packaging tests passed back then — and now they'd fail [[e.g.](https://github.com/janbrasna/h2/actions/runs/14668205170/job/41168195621#step:6:722)], probably based on some downstream version bumps in dependencies introducing bugs. I'd much rather avoid the cast quoting myself, but the checks won't pass otherwise [[e.g.](https://github.com/janbrasna/h2/actions/runs/14668791009/job/41170156275#step:6:106)]. Would you like me to add `TC006` to ignored rules instead for now?

.github/workflows/ci.yml
janbrasna: @Kriechi (Just to make it more traceable: the CI never had used the public release URL, it only ever used the API, unauthenticated: a.) [GitHub@`9df8f94`](https://github.com/python-hyper/h2/commit/9df8f94ce983d44ef57c8f332463f7b3cbe0127b#diff-9070df4576ba98ab3ee64d545982fd44ff474cf01bfde83d33be95e669b19b29), b.) [Travis@`0cb4090`](https://github.com/python-hyper/h2/commit/0cb4090433f46f3fe792ba7bd796e0456e5a0771#diff-761510f2cd2b3c469c16c3be0ff0aefe6e8679fd6c05319866049ad25d79659f)… so I don't believe the non-APIv3 way was ever used.)

janbrasna:
(NB: On last _master_ CI run Feb 27, "ruff>=0.8.0,<1" collected ruff==0.9.8, today it resolves to ruff==0.11.7 👀 … The `twine` version wouldn't change, however on Feb 27 freeze shows packaging==24.2, and today it runs with packaging==25.0 so that might be a metadata change that hit the preexisting twine bug, not exhibited in earlier runs.)

janbrasna:
So, another option here might be, instead of adding the rule to ignorelist, to constrain ruff a bit more, and find a version that's comfortable for now, e.g. `"ruff>=0.8.0,<0.10.0"` — however that means freezing the rules at that point in time few months back and not getting any new cool toys. Fine with either way, please advise on direction to tweak this further, happy to follow whatever sounds more comfortable for everyone…

Kriechi:
I prefer going with a reasonably "latest" version of ruff et al. So let's keep the version pin as is.
Please split out the TC006-related changes into a separate PR.

Not sure I understand the twine changes here yet - are they related to the ruff dependency? If not, please also split them out - unless thats the only thing remaining in this PR here.

janbrasna:
I'm not exactly comfortable starting with presumably passing master and offering several individual PRs that would each definitely fail (as they'd need all the changes landed together to really pass), so I'll do something else — I have one additional unrelated fix that basically doesn't touch any code, so I'll submit that one first, which should log a more recent check state for master after landing, and once/if broken for real, I'll start submitting the individual CI/dep fixes, and eventually circling back here to what's left to address…

PR #1307, opened 2025-09-30
fix: Raise more informative error instead of generic KeyError on end_stream

joaquin-borggio-lc:
Ending an invalid stream leads to KeyError. This PR leverages a helper function that already exists to properly handle errors and raise a more informative exception.


Kriechi:
LGTM! Please add a changelog entry and tests to cover the new behaviour.

PR #1313, opened 2026-02-20
Fix substring check used instead of equality for cookie header name

bysiber:
## Summary

Fix `_secure_headers` incorrectly matching header names that are substrings of `"cookie"`.

## Problem

In `_secure_headers`, the cookie header check uses:

```python
header[0] in b"cookie"
```

The `in` operator on `bytes` performs **substring search**, not equality. This means any header name that is a byte-level substring of `b"cookie"` will match:

```python
>>> b"co" in b"cookie"
True
>>> b"ok" in b"cookie"
True
>>> b"e" in b"cookie"
True
```

Any such header with a value shorter than 20 bytes will be incorrectly promoted to a `NeverIndexedHeaderTuple`, which tells HPACK to never add it to the dynamic compression table. While the security impact is minimal (the wrong headers are over-protected, not under-protected), it changes the HPACK encoding behavior in unexpected ways and could affect compression efficiency.

The same pattern exists in `_reject_pseudo_header_fields` for the `:method` check, though it is harmless in practice because only validated pseudo-headers reach that point.

## Fix

Change both `in` checks to `==` for exact equality:

```python
# Before
header[0] in b"cookie"
header[0] in b":method"

# After
header[0] == b"cookie"
header[0] == b":method"
```

### DIFF
--- a/src/h2/utilities.py
+++ b/src/h2/utilities.py
@@ -88,7 +88,7 @@ def _secure_headers(headers: Iterable[Header],
     """
     for header in headers:
         assert isinstance(header[0], bytes)
-        if header[0] in _SECURE_HEADERS or (header[0] in b"cookie" and len(header[1]) < 20):
+        if header[0] in _SECURE_HEADERS or (header[0] == b"cookie" and len(header[1]) < 20):
             yield NeverIndexedHeaderTuple(header[0], header[1])
         else:
             yield header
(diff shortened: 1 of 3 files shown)


bysiber:
Added regression tests:

- `test_non_cookie_substring_headers_can_be_indexed`: verifies that headers with names like `cook`, `okie`, `cooki` (substrings of `cookie` but not equal) are not incorrectly treated as sensitive
- `test_extract_method_header_no_false_substring_match`: verifies that a header named `:me` (substring of `:method`) does not falsely match in `extract_method_header`

All 324 tests in the affected files pass.

PR #1314, opened 2026-05-06
Allow sending 0-byte DATA frames when flow-control window is negative

lagemeet:
fix empty 0-byte DATA frame broke with FlowControlError

error
`Cannot send 0 bytes, flow control window is -49151`

RFC allows empty END_STREAM even if window is negative

> Frames with zero length with the END_STREAM flag set (that
>    is, an empty DATA frame) MAY be sent if there is no available space
>    in either flow-control window. 

allow 0-byte frames even when window < 0
only enforce flow-control check if frame_size > 0
skip negative window assert for empty payload in connection.py + stream.py

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,7 +23,7 @@ dev
 
 **Bugfixes**
 
--
+- Fix to allow sending 0 bytes on a stream even if the flow control window is negative.
 
 4.3.0 (2025-08-23)
 ------------------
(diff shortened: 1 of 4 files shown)


Kriechi:
Thanks! Please add a test and changelog entry.

lagemeet:
added tests and changelog changes to the pull request

Kriechi:
Thanks! 🎉

lagemeet:
Thank you too! <3 I found that bug while using the xmpp client called gajim with an haproxy server. it was not possible to send files if the server had an initial window size lower than 65535. 
image

PR #1315, opened 2026-05-10
Broaden receive data buffer types

### LINKED ISSUE #1305: Update types to adhere to `--strict-bytes` / PEP688
Mypy 1.16 added [`--strict-bytes`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict-bytes) to `--strict`, and plans to enable it by default in mypy 2.0. As far as I can tell the type hints in this package has not been updated to reflect this, giving false alarms for downstream users.

Pyright also does not promote `bytearray`/`memoryview` to `bytes` with [`disableBytesTypePromotions`](https://microsoft.github.io/pyright/#/configuration?id=type-evaluation-settings) defaulting to `True`.

The one I noticed when updating https://github.com/pgjones/hypercorn/pull/303 was [`H2Connection.receive_data`](https://github.com/python-hyper/h2/blob/0583911b29d05764bbe3f7691d59f4d5e83e249b/src/h2/connection.py#L1495) only accepting `bytes` for the `data` parameter, but it only passes `data` onto [`FrameBuffer.add_data`](https://github.com/python-hyper/h2/blob/0583911b29d05764bbe3f7691d59f4d5e83e249b/src/h2/frame_buffer.py#L39) which should be perfectly capable of also accepting e.g. `bytearray`.

It's likely you can replace a lot of type hints with `collections.abc.Buffer`, or at least a protocol which inherits from `Buffer` + `Sized` (since `FrameBuffer.add_data` makes use of `len`).

Harshal96:
## Summary

- Broaden `H2Connection.receive_data` and `FrameBuffer.add_data` annotations from `bytes` to a local PEP 688-style `Buffer` protocol.
- Normalize frame buffer input through `memoryview` so byte-like inputs are handled consistently.
- Add a strict-bytes mypy fixture covering `bytearray` and `memoryview`, and run it in the lint tox environment.

Fixes #1305.

## Validation

Local CI-equivalent checks run before advancing this draft:

- `uv run --no-project --with tox --with tox-gh-actions tox -e py311`
- `uv run --no-project --with tox --with tox-gh-actions tox -e lint`
- `uv run --no-project --with tox --with tox-gh-actions tox -e py310,py312,py313,py314,packaging`
- `uv run --python 3.13 --no-project --with tox --with tox-gh-actions tox -r -e docs`

Upstream GitHub Actions note: the PR's `CI` workflow is currently `action_required` with no jobs, so it has not executed on GitHub yet. The remaining `h2spec` job is Linux-specific in the workflow because it downloads a Linux h2spec binary, so I could not reproduce that exact job on this macOS workspace.

### DIFF
--- a/tests/typing/strict_bytes.py
+++ b/tests/typing/strict_bytes.py
@@ -1,0 +1,17 @@
+from __future__ import annotations
+
+from h2.connection import H2Connection
+from h2.frame_buffer import FrameBuffer
+
+
+def receive_data_accepts_buffer_types(
+    connection: H2Connection,
+    frame_buffer: FrameBuffer,
+) -> None:
+    bytearray_data = bytearray(b"")
+    memoryview_data = memoryview(b"")
+
+    connection.receive_data(bytearray_data)
+    connection.receive_data(memoryview_data)
+    frame_buffer.add_data(bytearray_data)
+    frame_buffer.add_data(memoryview_data)
--- a/tests/typing/strict_bytes.py
+++ b/tests/typing/strict_bytes.py
@@ -1,17 +1,28 @@
 from __future__ import annotations
 
+from hyperframe.frame import Frame
+from typing_extensions import assert_type
+
 from h2.connection import H2Connection
+from h2.events import Event
 from h2.frame_buffer import FrameBuffer
 
 
-def receive_data_accepts_buffer_types(
-    connection: H2Connection,
-    frame_buffer: FrameBuffer,
-) -> None:
+def receive_data_accepts_buffer_types() -> None:
+    connection = H2Connection()
+    frame_buffer = FrameBuffer()
     bytearray_data = bytearray(b"")
     memoryview_data = memoryview(b"")
 
-    connection.receive_data(bytearray_data)
-    connection.receive_data(memoryview_data)
+    bytearray_events = connection.receive_data(bytearray_data)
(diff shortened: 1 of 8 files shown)


tests/typing/strict_bytes.py
```diff
@@ -0,0 +1,17 @@
...
+    connection: H2Connection,
+    frame_buffer: FrameBuffer,
+) -> None:
+    bytearray_data = bytearray(b"")
+    memoryview_data = memoryview(b"")
+
+    connection.receive_data(bytearray_data)
+    connection.receive_data(memoryview_data)
+    frame_buffer.add_data(bytearray_data)
+    frame_buffer.add_data(memoryview_data)
```
Kriechi, line 20: What is this test asserting? Please add a validation check after these functions are called with the new argument types.

Kriechi:
Please add a changelog entry and document any major or minor API changes.
Does the new Buffer class cause any backwards incompatible changes?

tests/typing/strict_bytes.py
Harshal96: Updated

Harshal96:
> Please add a changelog entry and document any major or minor API changes. Does the new Buffer class cause any backwards incompatible changes?

Hi, thanks for the review. No breaking changes, existing ``bytes`` callers are unaffected.

PR #1317, opened 2026-05-14
fix: reject conflicting content-length headers

### LINKED ISSUE #1316: Reject conflicting duplicate Content-Length headers
h2 currently accepts request header blocks that contain multiple content-length fields with conflicting values. In src/h2/stream.py (line 1367), _initialize_content_length() uses the first content-length it sees and returns immediately, so later conflicting values remain in the emitted headers but are not considered for body length validation.

This can create parser disagreement for applications, proxies, or protocol translators layered on top of h2: h2 may accept a message based on the first value while downstream code may use the last value, combine values, or reject the request. That shape is request-smuggling-prone.

fix:
Track all content-length values in _initialize_content_length(). If more than one value is present and all values are identical, accept that value. If any value differs, raise ProtocolError. Add inbound tests for duplicate matching and duplicate conflicting content-length fields.

happy to open a pr to fix it

Harshal96:
## Summary

- Check all inbound `content-length` header fields before setting the expected body length
- Reject duplicate `content-length` fields when their parsed decimal values differ
- Add inbound tests for matching and conflicting duplicate fields

Fixes #1316

## Tests

- Added unit tests

### DIFF
--- a/src/h2/stream.py
+++ b/src/h2/stream.py
@@ -1377,7 +1377,7 @@ class H2Stream:
                 if content_length is None:
                     content_length = parsed_content_length
                 elif parsed_content_length != content_length:
-                    msg = "Conflicting content-length headers"
+                    msg = f"Conflicting content-length headers: {content_length} and {parsed_content_length}"
                     raise ProtocolError(msg)
 
         if content_length is None:
--- a/src/h2/stream.py
+++ b/src/h2/stream.py
@@ -1364,15 +1364,24 @@ class H2Stream:
             self._expected_content_length = 0
             return
 
+        content_lengths = []
+
         for n, v in headers:
             if n == b"content-length":
                 try:
-                    self._expected_content_length = int(v, 10)
+                    content_lengths.append(int(v, 10))
                 except ValueError as err:
                     msg = f"Invalid content-length header: {v!r}"
                     raise ProtocolError(msg) from err
 
-                return
+        if not content_lengths:
+            return
+
+        if len(set(content_lengths)) != 1:
+            msg = "Conflicting content-length headers"
+            raise ProtocolError(msg)
+
+        self._expected_content_length = content_lengths[0]
 
     def _track_content_length(self, length: int, end_stream: bool) -> None:
         """
--- a/src/h2/stream.py
+++ b/src/h2/stream.py
@@ -1380,9 +1380,6 @@ class H2Stream:
                     msg = f"Conflicting content-length headers: {content_length} and {parsed_content_length}"
                     raise ProtocolError(msg)
 
-        if content_length is None:
-            return
-
         self._expected_content_length = content_length
 
     def _track_content_length(self, length: int, end_stream: bool) -> None:
--- a/src/h2/stream.py
+++ b/src/h2/stream.py
@@ -1364,24 +1364,26 @@ class H2Stream:
             self._expected_content_length = 0
             return
 
-        content_lengths = []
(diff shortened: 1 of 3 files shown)


src/h2/stream.py
```diff
@@ -1364,15 +1364,24 @@ def _initialize_content_length(self, headers: Iterable[Header]) -> None:
             self._expected_content_length = 0
             return
 
+        content_lengths = []
+
         for n, v in headers:
             if n == b"content-length":
                 try:
-                    self._expected_content_length = int(v, 10)
+                    content_lengths.append(int(v, 10))
```
Kriechi: Could this lead to a DoS if a malicous payload keeps sending new differing content-length headers and this list keeps growing?

Kriechi:
asking for a second opinion from @mhils:
how does mitmproxy deal with duplicate or conflicting content-length headers?
last-one-wins? first-one-wins? throw an error / close the connection?

src/h2/stream.py
Harshal96: No, I don't think so, it'll be discarded once _initialize_content_length() exits, but I think I have a better idea where only comparing the lengths can work as well, lemme push a fix

mhils:
> how does mitmproxy deal with duplicate or conflicting content-length headers?

We reject them, because it has tremendous request smuggling potential.

src/h2/stream.py
```diff
@@ -1364,15 +1364,26 @@ def _initialize_content_length(self, headers: Iterable[Header]) -> None:
...
+                    parsed_content_length = int(v, 10)
                 except ValueError as err:
                     msg = f"Invalid content-length header: {v!r}"
                     raise ProtocolError(msg) from err
 
-                return
+                if content_length is None:
+                    content_length = parsed_content_length
+                elif parsed_content_length != content_length:
+                    msg = "Conflicting content-length headers"
```
Kriechi: Please add both values into the error message here.

src/h2/stream.py
```diff
@@ -1364,15 +1364,26 @@ def _initialize_content_length(self, headers: Iterable[Header]) -> None:
...
+                    raise ProtocolError(msg)
+
+        if content_length is None:
+            return
+
+        self._expected_content_length = content_length
```
Kriechi, line 1383: Could be simplified here: if content_length is None, then just assign it anyway, as _expected_content_length is going to be also None.

Kriechi:
@Harshal96 please add a changelog entry to document the new behaviour - this is breaking change, as the previous code paths silently accepted the first found content-length header (ignoring all others), while the new behaviour will raise a ProtocolError if there is a second header with a different value.

I am still considering if such a compliance check should better be part of the existing checks in `utilities.py::validate_headers`. What do you think?

src/h2/stream.py
Harshal96: done

src/h2/stream.py
Harshal96: done

Harshal96:
@Kriechi the check is tied to stream body accounting, so maybe should not be a part of `validate_headers`, but this is one of my initial contributions in this repo, so happy to accomodate the feedback if you feel it's better suited in `validate_headers`

Harshal96:
@Kriechi is this PR good to review/merge?

Harshal96:
@Kriechi gentle reminder for a review

Kriechi:
Thanks @Harshal96! 🎉

Issue #316, opened 2016-09-19
Clients should reject any attempt to change the SETTINGS_ENABLE_PUSH setting to a non-zero value


alexwlchan:
[RFC 7540 section 8.2](https://tools.ietf.org/html/rfc7540#section-8.2):

> Clients MUST reject any attempt to change the SETTINGS_ENABLE_PUSH setting to a value other than 0 by treating the message as a connection error (Section 5.4.1) of type PROTOCOL_ERROR.

If I’ve read this correctly, “attempt to change” means a `SettingsFrame`, yes?
If so, we don’t do this yet – indeed, there doesn’t seem to be a way for the `h2.settings.Settings` object to know if it’s for a client or server.

I’m not sure exactly when the `Settings` object gets configured – the user can specify the `initial_values` parameter, so presumably a client could have `ENABLE_PUSH = 1` set by this way. The validation can’t occur on the `Settings` object, but we actually receive the frame, in `H2Connection._receive_settings_frame()`.

That seems sub-optimal – it would be good to keep settings-specific logic in `settings.py`.

Perhaps we add a `validate_received_setting()` to mirror `_validate_setting()`, which is defined in `settings.py` but called in `_receive_settings_frame()`.

Lukasa:
Attempt to change *does* mean a `SettingsFrame`, yes. However, it seems like you're getting a bit turned around about setting stuff.

Yes, a client can have `ENABLE_PUSH=1` set by the `initial_values` parameter, but that can only happen for the settings *it offers to its peer*. Remember that, like with flow control windows, there are always *two* collections of active settings: those sent by the client to the server, and those sent by the server to the client.

However, all of this is easier than you think because the `H2Settings` object already gets told whether it's for a client or not. All we need to do to add extra validation is to *persist* that value on the object, and then extend `_validate_setting` to take a `client` parameter that determines whether the entity setting the value is a client or not. This is probably pretty easy, tbh.

Mirochill:
This appears to be covered by #1318, which was merged on May 23, 2026. That PR added role-aware validation for received SETTINGS so a client rejects a server-sent \\SETTINGS_ENABLE_PUSH\\ value other than \\

PR #1318, opened 2026-05-18
Reject client enable-push settings

Mirochill:
## Summary
- validate peer SETTINGS before applying them
- reject non-zero `SETTINGS_ENABLE_PUSH` values received from clients
- add a regression test and changelog entry

## Why
RFC 7540 requires clients to be rejected if they attempt to change `SETTINGS_ENABLE_PUSH` to a non-zero value. Issue #316 notes that `h2` did not enforce that rule for received SETTINGS frames.

The implementation keeps this validation on the received-settings path rather than changing the broader `Settings` mutation semantics, so it closes the protocol gap without altering existing local-setting behavior.

## Validation
- Not run locally
- Verified statically that the new check runs before received settings are applied and is covered by a targeted regression test.
- Relying on the repository's remote CI for execution feedback.

### DIFF
--- a/src/h2/settings.py
+++ b/src/h2/settings.py
@@ -126,6 +126,8 @@ class Settings(MutableMapping[SettingCodes | int, int]):
     """
 
     def __init__(self, client: bool = True, initial_values: dict[SettingCodes, int] | None = None) -> None:
+        self._client = client
+
         # Backing object for the settings. This is a dictionary of
         # (setting: [list of values]), where the first value in the list is the
         # current value of the setting. Strictly this doesn't use lists but
@@ -286,6 +288,30 @@ class Settings(MutableMapping[SettingCodes | int, int]):
             self._settings[key] = items
 
         items.append(value)
+
+    def validate_received_setting(self, setting: SettingCodes | int, value: int) -> None:
+        """
+        Validate a setting received from the peer that owns this Settings
+        object.
+
+        Clients may advertise ``ENABLE_PUSH`` only as ``0`` in received
+        SETTINGS frames.
+        """
+        invalid = _validate_setting(setting, value)
+        if (
+            not invalid
+            and self._client
+            and setting == SettingCodes.ENABLE_PUSH
+            and value != 0
+        ):
+            invalid = ErrorCodes.PROTOCOL_ERROR
+
+        if invalid:
+            msg = f"Setting {setting} has invalid value {value}"
+            raise InvalidSettingsValueError(
+                msg,
+                error_code=invalid,
+            )
 
     def __delitem__(self, key: SettingCodes | int) -> None:
         del self._settings[key]
--- a/src/h2/settings.py
+++ b/src/h2/settings.py
@@ -334,19 +334,19 @@ def _validate_setting(
     setting: SettingCodes | int,
     value: int,
     *,
-    client: bool = False,
+    client: bool | None = None,
 ) -> ErrorCodes:
     """
     Confirms that a specific setting has a well-formed value. If the setting is
     invalid, returns an error code. Otherwise, returns 0 (NO_ERROR).
 
-    If ``client`` is true, the setting originated from a client endpoint.
+    If ``client`` is set, the setting originated from a peer with that role.
     """
     if setting == SettingCodes.ENABLE_PUSH:
         # RFC 9113 section 6.5.2: "A client MUST treat receipt of a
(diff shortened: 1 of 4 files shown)


src/h2/settings.py
```diff
@@ -287,6 +289,30 @@ def __setitem__(self, key: SettingCodes | int, value: int) -> None:
...
+        invalid = _validate_setting(setting, value)
+        if (
+            not invalid
+            and self._client
+            and setting == SettingCodes.ENABLE_PUSH
+            and value != 0
+        ):
+            invalid = ErrorCodes.PROTOCOL_ERROR
+
+        if invalid:
```
Kriechi: `_validate_setting` returns an ErrorCodes and should be explicitly compared against `ErrorCodes.NO_ERROR`.

src/h2/settings.py
```diff
@@ -287,6 +289,30 @@ def __setitem__(self, key: SettingCodes | int, value: int) -> None:
...
+        Clients may advertise ``ENABLE_PUSH`` only as ``0`` in received
+        SETTINGS frames.
+        """
+        invalid = _validate_setting(setting, value)
+        if (
+            not invalid
+            and self._client
+            and setting == SettingCodes.ENABLE_PUSH
+            and value != 0
+        ):
```
Kriechi: This check should probably live in the _validate_setting function.

tests/test_invalid_frame_sequences.py
```diff
@@ -266,6 +266,22 @@ def test_reject_invalid_settings_values(self, frame_factory, settings) -> None:
...
+    def test_reject_client_enable_push_updates(self, frame_factory) -> None:
+        """
+        Servers reject clients that advertise non-zero SETTINGS_ENABLE_PUSH
+        values in received SETTINGS frames.
+        """
+        c = h2.connection.H2Connection(config=self.server_config)
+        c.initiate_connection()
+        c.receive_data(frame_factory.preamble())
+
+        f = frame_factory.build_settings_frame(settings={0x2: 1})
```
Kriechi: use `SettingCodes.ENABLE_PUSH` instead of `0x2`.

src/h2/settings.py
```diff
@@ -311,13 +330,20 @@ def __ne__(self, other: object) -> bool:
...
+) -> ErrorCodes:
     """
     Confirms that a specific setting has a well-formed value. If the setting is
     invalid, returns an error code. Otherwise, returns 0 (NO_ERROR).
+
+    If ``client`` is true, the setting originated from a client endpoint.
     """
     if setting == SettingCodes.ENABLE_PUSH:
-        if value not in (0, 1):
+        if value not in (0, 1) or (client and value != 0):
```
Kriechi: add a quote of https://datatracker.ietf.org/doc/html/rfc9113#name-defined-settings
section [6.5.2](https://datatracker.ietf.org/doc/html/rfc9113#section-6.5.2)

> A client MUST treat receipt of a SETTINGS frame with SETTINGS_ENABLE_PUSH set to 1 as a [connection error](https://datatracker.ietf.org/doc/html/rfc9113#ConnectionErrorHandler) ([Section 5.4.1](https://datatracker.ietf.org/doc/html/rfc9113#ConnectionErrorHandler)) of type [PROTOCOL_ERROR](https://datatracker.ietf.org/doc/html/rfc9113#PROTOCOL_ERROR).

Mirochill:
Thanks — addressed the review feedback in the latest commits.

I replaced the raw setting code in the regression with `h2.settings.SettingCodes.ENABLE_PUSH`, added the RFC 9113 §6.5.2 quote near the validation, and corrected the regression to model a client receiving the server setting so it matches the quoted rule.

No local tests were run; I only did static diff review / `git diff --check` here.

Mirochill:
Pushed one follow-up commit (`b9520e8`) after reading the remote `tox (3.14)` failure. The previous revision applied the role-specific `ENABLE_PUSH` rule to local `Settings(...)` construction and setters as well; this now keeps ordinary setting validation to the generic `0` / `1` rule and applies the server-peer restriction only to received settings.

Not run locally; this was based on the remote failure and static review.

Mirochill:
Pushed one small wording follow-up (`2ea6551`) after re-reading the final diff against RFC 9113 section 6.5.2. The code was already applying the role-specific rule to a server peer, but the changelog and helper docstring still said clients; those now say servers to match the tested case and the quoted RFC text.

Not run locally; this was a static wording-only review.

PR #1319, opened 2026-05-19
Reject conflicting duplicate Content-Length headers

### LINKED ISSUE #1316: Reject conflicting duplicate Content-Length headers

Mirochill:
## Summary
- validate all inbound `content-length` header fields before initializing the expected body length
- accept duplicate `content-length` headers only when all values agree
- add coverage for matching and conflicting duplicate `content-length` request headers

Closes #1316.

## Validation
- `git diff --check`
- Not run locally

### DIFF
--- a/src/h2/stream.py
+++ b/src/h2/stream.py
@@ -1364,15 +1364,23 @@ class H2Stream:
             self._expected_content_length = 0
             return
 
+        content_length = None
         for n, v in headers:
             if n == b"content-length":
                 try:
-                    self._expected_content_length = int(v, 10)
+                    header_content_length = int(v, 10)
                 except ValueError as err:
                     msg = f"Invalid content-length header: {v!r}"
                     raise ProtocolError(msg) from err
 
-                return
+                if content_length is not None and header_content_length != content_length:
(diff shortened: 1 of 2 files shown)


Mirochill:
I missed that #1317 was already actively covering this issue. Closing this draft to avoid duplicating review effort; #1317 has the maintainer discussion and remote checks attached. Sorry for the noise.

PR #1321, opened 2026-07-26
Remove consumed frames in place

### LINKED ISSUE #474: Rewrite frame_buffer to use bytearray()
Good artists copy, great artists steal. One of @njsmith's great decisions with [h11](https://github.com/njsmith/h11) was to write his [buffer](https://github.com/njsmith/h11/blob/master/h11/_receivebuffer.py) in terms of bytearray. This has some major efficiency gains on Python 3.4.

An easy win for us would be to rewrite the buffering internals of our frame buffer using h11's, while removing the things that h11's buffer implements that we don't need (in particular, any line-oriented constructs). Before we do it, we should write a benchmark of the buffering performance under reasonable HTTP/2 conditions to see what kind of improvements we see.

While I'm here: @fijal, the comments in h11's framebuffer say:

```python
# Note that starting in Python 3.4, deleting the initial n bytes from a
# bytearray is amortized O(n), thanks to some excellent work by Antoine
# Martin:
#
#     https://bugs.python.org/issue19087

...

# make sure PyPy also has the optimization
```

Does it?

utkarshalpha:
Closes #474.

`FrameBuffer` now stores incoming bytes in a `bytearray`, but consuming a frame still assigns `self._data[9 + length:]` back to the attribute. That creates and copies a new buffer for every frame, so parsing many buffered frames remains quadratic.

Delete the consumed prefix in place instead. CPython's optimized left deletion can then advance the bytearray start offset without copying the full remaining suffix. A regression test verifies both that the same bytearray object is retained and that the next frame remains buffered.

Benchmark

I buffered repeated valid, empty SETTINGS frames and then iterated the `FrameBuffer` on CPython 3.11 / Windows:

| Frames | Before | After | Speedup |
| ---: | ---: | ---: | ---: |
| 50,000 | 0.935s | 0.157s | 6.0x |
| 100,000 | 4.018s | 0.322s | 12.5x |
| 200,000 | 40.451s | 0.641s | 63.1x |

The post-change throughput stays near 312,000 frames/s across the three input sizes.

Validation

- `python -m pytest` — 1,654 passed
- `python -m mypy --strict-bytes src tests/typing/strict_bytes.py`
- Ruff on the changed source and test (excluding three unrelated existing findings in `test_basic_logic.py`)
- `git diff --check`

### DIFF
--- a/src/h2/frame_buffer.py
+++ b/src/h2/frame_buffer.py
@@ -155,6 +155,13 @@ class FrameBuffer:
 
         # At this point, as we know we'll use or discard the entire frame, we
         # can update the data.
+        # Deleting the consumed prefix mutates the bytearray in place instead
+        # of copying the remaining bytes into a new object, as slicing would.
+        # ``del s[i:j]`` is documented for mutable sequences in
+        # https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types
+        # and CPython's bytearray tracks an internal offset (``ob_start`` in
+        # Objects/bytearrayobject.c) that makes repeated deletes from the
+        # front amortized O(1) per byte rather than O(len) per frame.
         del self._data[:9+length]
 
         # Pass the frame through the header buffer.
--- a/src/h2/frame_buffer.py
+++ b/src/h2/frame_buffer.py
@@ -155,7 +155,7 @@ class FrameBuffer:
 
         # At this point, as we know we'll use or discard the entire frame, we
         # can update the data.
-        self._data = self._data[9+length:]
+        del self._data[:9+length]
 
         # Pass the frame through the header buffer.
         new_frame = self._update_header_buffer(f)
--- a/tests/test_basic_logic.py
+++ b/tests/test_basic_logic.py
@@ -23,6 +23,20 @@ import h2.stream
 from . import helpers
 
 
+class TestFrameBuffer:
+    def test_consumed_frames_are_removed_in_place(self) -> None:
+        frame = hyperframe.frame.SettingsFrame(0).serialize()
+        buffer = h2.frame_buffer.FrameBuffer()
+        buffer.max_frame_size = 65535
+        buffer.add_data(frame * 2)
+        data = buffer._data
+
+        next(buffer)
+
+        assert buffer._data is data
+        assert buffer._data == frame
+
+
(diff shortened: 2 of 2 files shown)


src/h2/frame_buffer.py
```diff
@@ -155,7 +155,7 @@ def __next__(self) -> Frame:
 
         # At this point, as we know we'll use or discard the entire frame, we
         # can update the data.
-        self._data = self._data[9+length:]
+        del self._data[:9+length]
```
Kriechi, line 165: Please add a comment here about the behaviour with in-place update without copy, ideally referencing CPython docs.

tests/test_basic_logic.py
```diff
@@ -23,6 +23,20 @@
...
+    def test_consumed_frames_are_removed_in_place(self) -> None:
+        frame = hyperframe.frame.SettingsFrame(0).serialize()
+        buffer = h2.frame_buffer.FrameBuffer()
+        buffer.max_frame_size = 65535
+        buffer.add_data(frame * 2)
+        data = buffer._data
+
+        next(buffer)
+
+        assert buffer._data is data
```
Kriechi, line 40: please add a comment to clearly call out that this checks if the object is still the same (internally checked with the `id(...)` function.

Kriechi:
Thanks - this looks like an amazing and unexpectedly simple change 🎉 
Please add a short changelog entry.

src/h2/frame_buffer.py
utkarshalpha: Added in 3a59b4a — the comment explains that the slice delete mutates the bytearray in place instead of copying the remainder, with a reference to the mutable-sequence docs (https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types) and to the `ob_start` offset in CPython's `Objects/bytearrayobject.c` that makes front deletes amortized O(1).

tests/test_basic_logic.py
utkarshalpha: Done in 3a59b4a — the comment now calls out that `is` asserts object identity (CPython compares `id(...)` of the operands), i.e. the buffer is still the very same bytearray object rather than a sliced copy.

Kriechi:
Thanks! 🎉

